Issue #39, opened 2019-04-30
Support for Terraform Beta installs


devorbitus:
I understand you may want to keep the version list as clean as possible but it would be nice if you supported an override or something to be able to install the beta version of Terraform like https://releases.hashicorp.com/terraform/0.12.0-beta2/terraform_0.12.0-beta2_darwin_amd64.zip

warrensbox:
I will look at the code tonight

warrensbox:
This a useful feature, especially if a user wants to test out a terraform beta release version.
A couple of developers suggested that we pass a parameter to list all the version (including betas). The default, however, will only show the official release.


For example:-
The default `tfswitch` will display:
```
  ▸ 0.11.7
    0.9.8
    0.11.13
    0.11.12
↓   0.11.11
```
The optional arg  `tfswitch --list-all` or `tfswitch -l` will display all versions including rc and betas:
```
  ▸  0.12.0-rc1
     0.12.0-beta2
     0.12.0-beta1
     0.12.0
     0.11.13
↓    0.11.11
```

I will merge your changes to master and incorporate my changes to allow optional parameter `--list-all` list all tf version. 
I will release this new feature on Monday (May 20).

warrensbox:
@devorbitus 
Would you be able test the changes before I release it?

Steps:
1. git checkout develop (your changes are merged to develop)
2. `go build -v -o tfswitch`
3. `./tfswitch 0.12.0-rc1` then `terraform -v` to verify 
4. `./tfswitch -l` or `./tfswitch --list-all` - verify if it list all the versions

devorbitus:
@warrensbox Tested it as well as looked over the code and it looks good to me.

devorbitus:
You may want to make some reference to the new feature on the [documentation page](https://warrensbox.github.io/terraform-switcher)

warrensbox:
I have released this feature with #45 

It should install any version except for:
* 0.12.0-alpha4
* 0.12.0-alpha3

That's because the file was named incorrectly on the hashicorp release page.

For example, the file for `0.12.0-alpha2` looks like below:
<img width="469" alt="Screen Shot 2019-05-20 at 9 46 34 PM" src="https://user-images.githubusercontent.com/38867521/58374808-483f0180-7f0a-11e9-9c00-cf3de3cde143.png">


However the the file for `0.12.0-alpha4` looks like below:
[...]

PR #40, opened 2019-05-13
support for all terraform versions

devorbitus:
Allows support for every release of Terraform including betas and release candidates.

fix for #39


warrensbox:
@devorbitus Please see my comments on #39.

PR #46, opened 2019-06-06
document fix for 'command not found: add-zsh-hook'

erikdw:
I hit the `command not found: add-zsh-hook` error when using `tfswitch` in zsh (v5.6.2).  Searched the web for a bit and realized I needed to load the `add-zsh-hook` function.


erikdw:
thanks @warrensbox !

Issue #69, opened 2020-03-27
Add versions.tf support


kpucynski:
For now TF version to switch can be automatically taken from `.tfswitchrc` or `.terraform-version` files.
In the meantime `versions.tf` file became de facto standard for  TF version control.
It would be very useful to use already existing `required_version` from terraform code instead of duplicating this info in separate file(s).

warrensbox:
I will look at this today

warrensbox:
@kinowarrior Does your versions.tf look something like this:   
```
terraform {
  required_version = ">= 0.12.9"

  required_providers {
    aws        = ">= 2.52.0"
    local      = ">= 1.2"
    null       = ">= 2.1"
    template   = ">= 2.1"
    random     = ">= 2.1"
    kubernetes = ">= 1.11.1"
  }
}
```

kpucynski:
I've created PR with code that checks *.tf for required version and uses it if found.

warrensbox:
Not released yet. I am still testing the changes.

warrensbox:
### Use version.tf file  
If a .tf file with the terraform constrain is included in the current directory, it should automatically download or switch to that terraform version. For example, the following should automatically switch terraform to version `0.12.24`:     
```
terraform {
  required_version = ">= 0.12.9"

  required_providers {
    aws        = ">= 2.52.0"
    kubernetes = ">= 1.11.1"
  }
}
```
<img src="https://s3.us-east-2.amazonaws.com/kepler-images/warrensbox/tfswitch/versiontf.gif" alt="drawing" style="width: 170px;"/>

warrensbox:
@kpucynski your changes are released. You can upgrade with:   
Homebrew - `brew upgrade warrensbox/tap/tfswitch `.  
Other linux systems - `curl -L https://raw.githubusercontent.com/warrensbox/terraform-switcher/release/install.sh | bash
`   

Thank you for contributing!

kpucynski:
@warrensbox thanks! :)

PR #72, opened 2020-04-03
Support reading required version from TF code.

kpucynski:
Solves #69

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,11 +22,20 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"sort"
 	"strings"
+
+	"github.com/Masterminds/semver"
+
+	// original hashicorp upstream have broken dependencies, so using fork as workaround
+	// TODO: move back to upstream
+	"github.com/kiranjthomas/terraform-config-inspect/tfconfig"
+	//	"github.com/hashicorp/terraform-config-inspect/tfconfig"
 
 	"github.com/manifoldco/promptui"
 	"github.com/pborman/getopt"
 	"github.com/spf13/viper"
+
 	lib "github.com/warrensbox/terraform-switcher/lib"
 )
 
@@ -121,6 +130,49 @@ func main() {
 					os.Exit(1)
 				}
 			}
+		} else if module, _ := tfconfig.LoadModule(dir); len(module.RequiredCore) >= 1 && len(args) == 0 { //if there is required_version defined in TF code, and no commmand line arguments
+			tfversion := ""
+
+			// we skip duplicated definitions and use only first one
+			tfconstraint := module.RequiredCore[0]
+			tflist, _ := lib.GetTFList(hashiURL, true)
+			fmt.Printf("Reading required version from terraform code, constraint: %s\n", tfconstraint)
+			//fmt.Printf("Available versions: %#v\n", tflist)
+
+			c, err := semver.NewConstraint(tfconstraint)
+			if err != nil {
+				fmt.Println("Error parsing constraint:", err)
+				return
+			}
+			vs := make([]*semver.Version, len(tflist))
+			for i, r := range tflist {
+				v, err := semver.NewVersion(r)
+				if err != nil {
+					fmt.Printf("Error parsing version: %s", err)
+				}
+
+				vs[i] = v
+			}
+
+			sort.Sort(semver.Collection(vs))
+
+			for i := range vs {
+				// Validate a version against a constraint.
+				// Take highest version that matches
+				if c.Check(vs[len(vs)-1-i]) {
+					tfversion = string(vs[len(vs)-1-i].String())
+					fmt.Printf("Matched version: %s\n", tfversion)
+
+					if lib.ValidVersionFormat(tfversion) { //check if version is correct
+						lib.Install(string(tfversion), *custBinPath)
+					} else {
+						fmt.Println("Invalid terraform version format. Format should be #.#.# or #.#.#-@# where # is numbers and @ is word characters. For example, 0.11.7 and 0.11.9-beta1 are valid versions")
+						os.Exit(1)
+					}
+				}
+			}
+
+			fmt.Println("No version found to match constraint.")
 
 		} else if _, err := os.Stat(rcfile); err == nil && len(args) == 0 { //if there is a .tfswitchrc file, and no commmand line arguments
 			fmt.Printf("Reading required terraform version %s \n", rcFilename)
--- a/main.go
+++ b/main.go
@@ -130,41 +130,41 @@ func main() {
 					os.Exit(1)
 				}
 			}
-		} else if module, _ := tfconfig.LoadModule(dir); len(module.RequiredCore) >= 1 && len(args) == 0 { //if there is required_version defined in TF code, and no commmand line arguments
+		} else if module, _ := tfconfig.LoadModule(dir); len(module.RequiredCore) >= 1 && len(args) == 0 { //if there is a .tfswitchrc file, and no commmand line arguments
 			tfversion := ""
 
 			// we skip duplicated definitions and use only first one
 			tfconstraint := module.RequiredCore[0]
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -121,6 +130,49 @@ func main() {
...
+			for i, r := range tflist {
+				v, err := semver.NewVersion(r)
+				if err != nil {
+					fmt.Printf("Error parsing version: %s", err)
+				}
+
+				vs[i] = v
+			}
+
+			sort.Sort(semver.Collection(vs))
```
jellonek, line 157: If you would use https://golang.org/pkg/sort/#Reverse you could later iterate "in normal way" over elements, instead of using this index based notation.

main.go
```diff
@@ -121,6 +130,49 @@ func main() {
...
+
+			for i := range vs {
+				// Validate a version against a constraint.
+				// Take highest version that matches
+				if c.Check(vs[len(vs)-1-i]) {
+					tfversion = string(vs[len(vs)-1-i].String())
+					fmt.Printf("Matched version: %s\n", tfversion)
+
+					if lib.ValidVersionFormat(tfversion) { //check if version is correct
+						lib.Install(string(tfversion), *custBinPath)
```
jellonek, line 167: Shouldn't there be used `break`, or you want intentionally to install multiple versions?

main.go
```diff
@@ -121,6 +130,49 @@ func main() {
...
+					} else {
+						fmt.Println("Invalid terraform version format. Format should be #.#.# or #.#.#-@# where # is numbers and @ is word characters. For example, 0.11.7 and 0.11.9-beta1 are valid versions")
+						os.Exit(1)
+					}
+				}
+			}
+
+			fmt.Println("No version found to match constraint.")
```
jellonek, line 175: Isn't that an error? If so - missing `os.Exit(1)`.

main.go
```diff
@@ -121,6 +130,50 @@ func main() {
 					os.Exit(1)
 				}
 			}
+		} else if module, _ := tfconfig.LoadModule(dir); len(module.RequiredCore) >= 1 && len(args) == 0 { //if there is a .tfswitchrc file, and no commmand line arguments
+			tfversion := ""
+
+			// we skip duplicated definitions and use only first one
+			tfconstraint := module.RequiredCore[0]
+			tflist, _ := lib.GetTFList(hashiURL, true)
```
mlynekt, line 138: Bare boolean values in function call arguments are unclear at best. Please declare an variable with meaningful name and use it here.

main.go
```diff
@@ -121,6 +130,49 @@ func main() {
...
+			if err != nil {
+				fmt.Println("Error parsing constraint:", err)
+				return
+			}
+			vs := make([]*semver.Version, len(tflist))
+			for i, r := range tflist {
+				v, err := semver.NewVersion(r)
+				if err != nil {
+					fmt.Printf("Error parsing version: %s", err)
+				}
```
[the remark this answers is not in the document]
mlynekt: Please correct me if I am wrong but, in my opinion, skipping version in wrong format with proper warning log would be more desirable by an user. If you choose so, please remember to create a 0 length `vs` slice and append to it.

main.go
```diff
@@ -121,6 +130,50 @@ func main() {
...
 			}
+		} else if module, _ := tfconfig.LoadModule(dir); len(module.RequiredCore) >= 1 && len(args) == 0 { //if there is a .tfswitchrc file, and no commmand line arguments
+			tfversion := ""
+
+			// we skip duplicated definitions and use only first one
+			tfconstraint := module.RequiredCore[0]
+			tflist, _ := lib.GetTFList(hashiURL, true)
+			fmt.Printf("Reading required version from terraform code, constraint: %s\n", tfconstraint)
+
+			c, err := semver.NewConstraint(tfconstraint)
```
mlynekt, line 141: `c` is used too far from its declaration thus I suggest a more descriptive name.

main.go
```diff
@@ -121,6 +130,50 @@ func main() {
 					os.Exit(1)
 				}
 			}
+		} else if module, _ := tfconfig.LoadModule(dir); len(module.RequiredCore) >= 1 && len(args) == 0 { //if there is a .tfswitchrc file, and no commmand line arguments
+			tfversion := ""
```
mlynekt, line 134: This line is not needed. Please delete it.

main.go
```diff
@@ -121,6 +130,50 @@ func main() {
...
+
+				vs[i] = v
+			}
+
+			sort.Sort(sort.Reverse(semver.Collection(vs)))
+
+			for _, element := range vs {
+				// Validate a version against a constraint.
+				if c.Check(element) {
+					tfversion = string(element.String())
```
mlynekt, line 162: `element.String()` should return a `string` so no need to cast.

main.go
```diff
@@ -121,6 +130,50 @@ func main() {
...
+					fmt.Printf("Matched version: %s\n", tfversion)
+
+					if lib.ValidVersionFormat(tfversion) {
+						lib.Install(string(tfversion), *custBinPath)
```
mlynekt, line 166: ditto

main.go
```diff
@@ -121,6 +130,50 @@ func main() {
...
+						break
+					} else {
+						fmt.Println("Invalid terraform version format. Format should be #.#.# or #.#.#-@# where # is numbers and @ is word characters. For example, 0.11.7 and 0.11.9-beta1 are valid versions")
+						os.Exit(1)
+					}
+				}
+			}
+
+			fmt.Println("No version found to match constraint.")
+			os.Exit(1)
```
mlynekt, line 176: But now, correct flow will also print this msg and exit with code 1. One way of doing this correctly will be to declare `isVersionFound := false` before the loop, set it `true` before the `break` and check it after the loop.

main.go
```diff
@@ -121,6 +130,50 @@ func main() {
...
+			if err != nil {
+				fmt.Println("Error parsing constraint:", err)
+				os.Exit(1)
+			}
+			vs := make([]*semver.Version, len(tflist))
+			for i, r := range tflist {
+				v, err := semver.NewVersion(r)
+				if err != nil {
+					fmt.Printf("Error parsing version: %s", err)
+					os.Exit(1)
```
mlynekt, line 151: Please correct me if I am wrong but, in my opinion, skipping version in wrong format with proper warning log would be more desirable by an user. If you choose so, please remember to create a 0 length `vs` slice and append to it.

main.go
warrensbox: Can you elaborate?

main.go
warrensbox: Correct

main.go
warrensbox: The break is not necessary because the install function(**lib.Install(string(tfversion), *custBinPath)**) will exit if it is successful or fail if it is unsuccessful:
 For loop:
```
for _, element := range vs {
				// Validate a version against a constraint.
				if c.Check(element) {
					tfversion = string(element.String())
					fmt.Printf("Matched version: %s\n", tfversion)

					if lib.ValidVersionFormat(tfversion) {
						lib.Install(string(tfversion), *custBinPath) //will exit if it is successful or fail if it is unsuccessful
						break // Not needed
					} else {
						fmt.Println("Invalid terraform version format. Format should be #.#.# or #.#.#-@# where # is numbers and @ is word characters. For example, 0.11.7 and 0.11.9-beta1 are valid versions")
						os.Exit(1)
					}
				}
			}
```

See the install function:
```
func Install(tfversion string, binPath string) {
  .  
  os.Exit(1) // if something is unsuccessful 
  .  
  . 
  os.Exit(0)// if all is good
} 
```

main.go
warrensbox: What do you mean by this?  
Do you mean we do not exit 1 when there's a wrong format?

main.go
warrensbox: updated

PR #85, opened 2020-08-11
Added file extension var for basic Windows support

Dogers:
This makes tfswitch function properly when run under an administrative account


warrensbox:
The latest release should support this. Let me know if you find any bugs. Thanks!

PR #100, opened 2020-09-11
Execute tfswitch after cd in fish shell

timkrueger:
Add a snippet to execute tfswitch after a `cd` if a `tf` file with an `required_version` is present. This works only in [fish shell](http://fishshell.com).

I'm not really sure if the project wants to start a snippet collection. Feel free to advise, correct or reject.


warrensbox (approved):

warrensbox:
Thanks!

PR #153, opened 2021-03-31
Support for terragrunt files

gmcoringa:
Add support for reading terragrunt.hcl file.

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -47,6 +48,7 @@ const (
 	tfvFilename   = ".terraform-version"
 	rcFilename    = ".tfswitchrc"
 	tomlFilename  = ".tfswitch.toml"
+	tgHclFilename = "terragrunt.hcl"
 )
 
 var version = "0.11.0\n"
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -415,3 +431,23 @@ func installTFProvidedModule(dir string, custBinPath *string) {
...
 }
+
+// Install using version constraint from terragrunt file
+func installTGHclFile(tgFile *string, custBinPath *string) {
+	fmt.Printf("Terragrunt file found: %s\n", *tgFile)
+	content := retrieveFileContents(*tgFile)
+	var constraint = ""
+
+	for _, line := range strings.Split(content, "\n") {
+		regex, _ := regexp.Compile(`^terraform_version_constraint\s+=\s+"(?P<version>.*)".*`)
```
versus, line 442: I think you should move regexp.Compile()  before loop

gmcoringa: fixed

warrensbox:
This is merged to master but not released yet. I am still testing.
One thing, I would like is to use the HCL parser instead of regex. 
See: https://github.com/yutannihilation/go-parse-hcl-example/blob/master/main.go
The hcl parser would take parse `terraform_version_constraint` from the terragrunt file.

warrensbox:
@gmcoringa  your changes are released with version 0.12
Instead of using regex, I refactored your code to use the hcl parser:
https://github.com/warrensbox/terraform-switcher/blob/2865ec9eb81fa8dc8cac1804d9c946999e1a785e/main.go#L441
Release #163 
Please upgrade and verify:
https://tfswitch.warrensbox.com/Upgrade-or-Uninstall/

PR #166, opened 2021-05-06
Add alias documentation advice

samuelbaena:
Add alias documentation to avoid setting the binary path everytime


yermulnik (commented):
I'd suggest to recommend a [more "official" approach](https://github.com/warrensbox/terraform-switcher#use-tfswitchtoml-file--for-non-admin---users-with-limited-privilege-on-their-computers) instead of CLI aliasing.

samuelbaena:
@yermulnik You are absolutely right, that is cleaner. Thanks

PR #168, opened 2021-05-14
Fix newline formatting issue

mdeeks:
Lack of line break causes lines to print together

```
❯ tfswitch
Reading required version from terraform fileReading required version from constraint: ~> 0.14.0
Matched version: 0.14.11
Switched terraform to version "0.14.11"
```


warrensbox:
Thanks!

PR #174, opened 2021-05-30
♿  changes to make tfswitch work on windows, fix some test 🐛 bugs, ♻️ refactor code, add 🚀 ci on `windows/osx`

Shuliyey:
# Overview

* ♿ changes to make `tfswitch` work on windows;
  * before: fails at `rename` unzipped `terraform` binary to `terraform_<version>`, as windows terraform binary is `terraform.exe` (ends with `exe`)
  * now: installs `terraform` successfully to `$HOME/bin`, when `$HOME/.tfswitch.toml` is defined (with `bin="<%USERPROFILE%>\\bin\\terraform.exe"`) and `tfswitch` is run under `Administrator` Mode
* ✅ fix process holding `RECENT` file issue in `file_test.go` (`TestWriteLines`/`TestReadLines`) cases, this causes os panic;
* ♻️ refactor code use `filepath.Join` instead of concatenating strings manually, avoid duplicated colde on file path through defining variables to store file path;
* 🚀 add ci (`travis.yaml`) on `windows/osx`

https://user-images.githubusercontent.com/4773348/120096838-fbcdd280-c170-11eb-9b94-634412e2fb2f.mp4

## Todo

* on `windows` the `defaultBin    = "/usr/local/bin/terraform"` doesn't exists, and windows have no concept like `/usr/local/bin`, so it's a bit underministic on the `defaultBin` for windows (in this change, it gets converted to `/usr/local/bin/terraform.exe`), hence for windows a `$HOME/.tfswitch.toml` need to exists with minimum content `bin` need to be defined with a valid path, below is an example
```toml
bin = "C:\\Users\\老叶\\bin\\terraform.exe"
```

* `mklink` requires `Administrator` permission to work on windows (with `/D` option), tfswitch must be run under `Administrator` mode on windows to work, or `symLink` permission issue will be thrown, below is an example

```powershell
2021/05/30 15:59:13
                Unable to create new symlink.
                Maybe symlink already exist. Try removing existing symlink manually.
                Try running "unlink C:\Users\老叶\test-tfswitcher-src" to remove existing symlink.
                If error persist, you may not have the permission to create a symlink at C:\Users\老叶\test-tfswitcher-src.
                Error: symlink C:\Users\老叶\test-tfswitcher-dest C:\Users\老叶\test-tfswitcher-src: A required privilege is not held by the client.
```

* `promptui` doesn't work nicely on `windows` on option scrolling, to install earlier version of terraform on windows `tfswitch <version>` is the workaround

### DIFF
--- a/lib/files_test.go
+++ b/lib/files_test.go
@@ -21,8 +21,7 @@ import (
 // TestRenameFile : Create a file, check filename exist,
 // rename file, check new filename exit
 func TestRenameFile(t *testing.T) {
-
-	installFile := "terraform"
+	installFile := lib.ConvertExecutableExt("terraform")
 	installVersion := "terraform_"
 	installPath := "/.terraform.versions_test/"
 	version := "0.0.7"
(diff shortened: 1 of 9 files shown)


warrensbox:
looking

lib/files_test.go
```diff
@@ -233,11 +233,11 @@ func TestWriteLines(t *testing.T) {
 			}
 		}
 
+		file.Close()
```
warrensbox, line 236: Everything looks good. My only question here is why not use `defer` to close the file?

lib/files_test.go
```diff
@@ -250,21 +250,21 @@ func TestReadLines(t *testing.T) {
 	if errCurr != nil {
 		log.Fatal(errCurr)
 	}
-	installLocation := usr.HomeDir + installPath
+	installLocation := filepath.Join(usr.HomeDir, installPath)
```
warrensbox, line 253: Using `filepath.Join()`  is probably a better solution. Thanks!

warrensbox (commented):
This mostly looks good

warrensbox:
I will merge this tonight. I will release it at a later time. I'll update you here when I release it.

lib/files_test.go
Shuliyey: cheers warren, i came across a race condition between `file.Close()` and https://github.com/warrensbox/terraform-switcher/blob/309e0ba04a08c66b3a17a06cf1b0b86a51f9c1e8/lib/files_test.go#L240,

i believe the os is trying to remove the `installLocation`, however it couldn't do so until the `file.Close()` is complete

lib/files_test.go
Shuliyey: just to confirm this line is ok 😄 , it's using `filepath.Join`

PR #202, opened 2021-10-28
Install darwin arm64 terraform version when desired version is equal …

hlepesant:
…or greater than 1.0.2 and you run MacOS on M1 CPU

Hi this MR is to fix the : 
TODO: Workaround for macos arm64 since terraform doesn't have a binary for it yet
 line 106 of lib/install.go

### DIFF
--- a/lib/install.go
+++ b/lib/install.go
@@ -17,7 +17,7 @@ const (
 	installPath    = ".terraform.versions"
 	recentFile     = "RECENT"
 	defaultBin     = "/usr/local/bin/terraform" //default bin installation dir
-	tfDarwinArm64  = "1.0.2"
+	tfDarwinArm64StartVersion  = "1.0.2"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


analytically:
@warrensbox any change of merging and new version?

lib/install.go
```diff
@@ -102,8 +104,10 @@ func Install(tfversion string, binPath string, mirrorURL string) {
 	goarch := runtime.GOARCH
 	goos := runtime.GOOS
 
-	// TODO: Workaround for macos arm64 since terraform doesn't have a binary for it yet
-	if goos == "darwin" && goarch == "arm64" {
+	// Terraform darwin arm64 comes with 1.0.2 and next version
+	tfver, _ := version.NewVersion(tfversion)
+	tf102, _ := version.NewVersion(tfDarwinArm64)
```
vfoucault, line 109:
```suggestion
	tf102, _ := version.NewVersion(tfDarwinArm64StartVersion)
```

vfoucault (approved):
LGTM

PR #208, opened 2022-01-23
Fix/semver constraints

warrensbox:
Semver for ~> is not working as expected.
Updated to use offical hashicorp libabries


warrensbox (commented):
More testing required

Issue #210, opened 2022-02-09
tfwitch blindly removes default terraform symlink, when bin is passed and version does not exist locally


miguelhar:
Consider the following scenario,  we want to have 2 versions of terraform with 2 different binary values.

- `terraform`
- `terraform0.12`

```bash
$ls -ltra /usr/local/bin/terraform*
/usr/local/bin/terraform -> ~/.terraform.versions/terraform_1.0.6
/usr/local/bin/terraform0.12 -> ~/.terraform.versions/terraform_0.12.31
```

changing versions for terraform0.12 behaves as expected since version 0.12.31  is present(in `~/.terraform.versions/`):

```bash
$tfswitch --bin=/usr/local/bin/terraform0.12 0.12.31
Switched terraform to version "0.12.31"
```

```bash
$ls -ltra /usr/local/bin/terraform*
/usr/local/bin/terraform -> ~/.terraform.versions/terraform_1.0.6
 /usr/local/bin/terraform0.12 -> ~/.terraform.versions/terraform_0.12.31
```

Now we will try with version 0.12.30 which had not been downloaded...

```bash
$tfswitch --bin=/usr/local/bin/terraform0.12 0.12.30
Downloading to: ~/.terraform.versions
29172646 bytes downloaded
Switched terraform to version "0.12.30"
```

As you can see the symlink  `/usr/local/bin/terraform -> ~/.terraform.versions/terraform_1.0.6` is removed:

```bash
$ls -ltra /usr/local/bin/terraform*
 /usr/local/bin/terraform0.12 -> ~/.terraform.versions/terraform_0.12.30
```

miguelhar:
I believe the culprit is the hardcode of the terraform path, instead of looking at the `--bin` value

https://github.com/warrensbox/terraform-switcher/blob/ea69e4d842b42e108ce6d2fbb164104c821e5207/lib/install.go#L35

warrensbox:
@miguelhar what version of tfswitch are you using?
`tfswitch -v`

miguelhar:
removing [this](https://github.com/warrensbox/terraform-switcher/blob/master/lib/install.go#L93) function fixes the issue(tested locally)
as the comment states, it seems this function is not necessary.
```go

// initialize : removes existing symlink to terraform binary// I Don't think this is needed
func initialize() {

	/* Step 1 */
	/* initilize default binary path for terraform */
	/* assumes that terraform is installed here */
[...]

PR #211, opened 2022-02-15
Fixes Issue #210, initialize function is not needed

miguelhar:


miguelhar:
@warrensbox can you please take a look

Jukie:
Hi @miguelhar, is this still an issue on the latest release? I agree that it doesn't seem necessary but I'm still seeing the symlink issue after adding cherry-picking your change. Could you try rebasing over https://github.com/warrensbox/terraform-switcher/tree/feature/release-0.14-API to test?

PR #221, opened 2022-04-08
Handle multiple terraform architecture (arm64/amd64) for the same version

alex-ikse:
We use multiple version of terraform in arch amd64 and arm64.

With recent terraform stacks, no problem, we can run all terraform/provider as arm64 binaries, but for older stacks, that use unmaintained provider (available only as amd64), we need to run terraform as amd64 (via rosetta 2) ...

In this last case, tfswitch don't download terraform again because the version already exist in ~/.terraform.versions ... but with bad arch.

The proposed patch add the architecture to the terraform binary name to be able to have an arm64 and amd64 version of the same terraform version and let the user choose a specific architecture for terraform (with the new -a / --arch parameter).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 EXE  := tfswitch
 PKG  := github.com/warrensbox/terraform-switcher
-VER := $(shell git ls-remote --tags git://github.com/warrensbox/terraform-switcher | awk '{print $$2}'| awk -F"/" '{print $$3}' | sort -n -t. -k1,1 -k2,2 -k3,3 | tail -n 2 | head -n1)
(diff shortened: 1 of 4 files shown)


Jukie:
Hi @alex-ikse, could elaborate on the scenarios where you need to override `runtime.GOARCH`? Wouldn't you then need a machine running the given architecture to execute tfswitch? I don't think I'm following.

davidvasandani:
tfenv handles with `TFENV_ARCH=amd64` which downloads the amd64 version (instead of arm64 specified by `runtime.GOARCH`). The amd64 version of Terraform is run via Rosetta for compatibility with providers that haven't been updated to arm64 yet.

kneufeld-pbp:
I am looking for the options to switch between native `darwin_arm64` and `darwin_amd64` as well.  I have a number of providers that don't have native `darwin_arm64` support yet.

yermulnik:
Closing as stale.
Please update this thread if this is still a case and we'll look into it again. Thanks.

PR #227, opened 2022-05-11
Use new api endpoint

### LINKED ISSUE #231: Use new releases API
#224 was an initial fix after the announcement of [the releases API](https://www.hashicorp.com/blog/announcing-the-hashicorp-releases-api) but the current implementation uses the old (but still supported) releases endpoint.

Jukie:
Closes #231

These changes refactor for compliance with the updated endpoint.

One thing to note: After updating to this version, any _existing_ cached versions file will be dubbed "dirty" so first execution will fail. I'd be open to adding a retry or additional handling logic but bailing out of the install is likely the safest route and retains the current behavior.


warrensbox:
@Jukie I am open to this change.

Jukie:
@warrensbox `after` is still broken so this isn't useable as-is but I'll keep cleaning up the code until the issue is fixed. 
That can be tracked [here](https://github.com/hashicorp/terraform/issues/31033) and someone from Hashi gave end of the week as an eta.

yermulnik:
@Jukie Should this PR be converted to `Draft` PR to sort of mark it as WIP and that it should be not merged as of yet and unless there's a fix in https://github.com/hashicorp/terraform/issues/31033 ?
Ref: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request

Jukie:
Good idea @yermulnik, thanks!

yermulnik:
> Good idea @yermulnik, thanks!

Cheers @Jukie 
And thanks for the both contributions! You saved a life for some of us =)

Jukie:
https://github.com/hashicorp/terraform/issues/31033 was fixed so I'll start working on this again.

Jukie:
So this mostly works now but I'd like some suggestions for fixing the semver parsing tests.
Obviously needs some additional cleanup as well.

lib/list_versions.go
```diff
@@ -1,119 +1,102 @@
...
-	resp, errURL := http.Get(mirrorURL)
+	req.URL.RawQuery = q.Encode()
+	client := &http.Client{}
+	res, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode != http.StatusOK {
+		// should clean this up
+		return nil, errors.New("issue during request")
```
yermulnik, line 92: Would it make sense to expose error details to the user? Something like this:
```suggestion
		err := fmt.Errorf("issue during request (%d: %q)", res.StatusCode, res.Status)
		return nil, errors.New(err)
```

lib/list_versions.go
```diff
@@ -126,17 +109,71 @@ func GetTFURLBody(mirrorURL string) ([]string, error) {
...
+	}
+	if err := json.Unmarshal(body.Bytes(), &releases); err != nil {
+		log.Fatalf("%s: %s", err, body.String())
+	}
+	return releases, nil
+}
+
+func GetTFReleases(mirrorURL string, preRelease bool) ([]Release, error) {
+	// temp testing
+	queryParams := map[string]string{"limit": "20"}
```
yermulnik, line 124: Would it make sense to put `20` into a variable just in case they change their mind and allow a different limit so that we change the var value closer to the header of this file rather than looking up this value somewhere in the middle of the file? 🤔

lib/list_versions.go
```diff
@@ -126,17 +109,71 @@ func GetTFURLBody(mirrorURL string) ([]string, error) {
...
+	rel, _ := getReleases(mirrorURL, queryParams)
+
+	var releases []Release
+	rel, _ = getReleases(mirrorURL, queryParams)
+	releases = append(releases, rel...)
+	for len(rel) == 20 {
```
yermulnik, line 130: Argh, `20` once again — putting it as a variable indeed makes sense imho.

yermulnik: Also the question since I'm not any good with Go: why `for` and not `if`?

lib/list_versions.go
```diff
@@ -126,17 +109,71 @@ func GetTFURLBody(mirrorURL string) ([]string, error) {
...
+func GetTFRelease(mirrorURL, requestedVersion string) (Release, error) {
+	resp, errURL := httpGet(mirrorURL+"/"+requestedVersion, nil)
+	if errURL != nil {
+		log.Printf("[Error] : Getting url: %v", errURL)
+		os.Exit(1)
+		return Release{}, errURL
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
```
yermulnik, line 158: Is this possible to compare with `http`'s `StatusOK` constant instead of relying on assumption that `200` is equal to `OK` within the meaning of `http` package?

lib/list_versions.go
```diff
@@ -126,17 +109,71 @@ func GetTFURLBody(mirrorURL string) ([]string, error) {
...
+	})
+	return releases, nil
+}
+
+func GetTFRelease(mirrorURL, requestedVersion string) (Release, error) {
+	resp, errURL := httpGet(mirrorURL+"/"+requestedVersion, nil)
+	if errURL != nil {
+		log.Printf("[Error] : Getting url: %v", errURL)
+		os.Exit(1)
+		return Release{}, errURL
```
yermulnik, line 154: As I wrote above I'm not quite familiar with Golang hence the question: how does this work — shouldn't `os.Exit()` abort execution right away?
What I mean is that will it indeed get to the `return` statement at all?

lib/list_versions.go
```diff
@@ -126,17 +109,71 @@ func GetTFURLBody(mirrorURL string) ([]string, error) {
...
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		log.Printf("[Error] : Retrieving contents from url: %s", mirrorURL)
 		os.Exit(1)
```
yermulnik, line 160: Why not `log.Fatal()` instead of `log.Printf` + `os.Exit()`?

lib/list_versions.go
```diff
@@ -126,17 +109,71 @@ func GetTFURLBody(mirrorURL string) ([]string, error) {
...
+		return releases[i].Version > releases[j].Version
+	})
+	return releases, nil
+}
+
+func GetTFRelease(mirrorURL, requestedVersion string) (Release, error) {
+	resp, errURL := httpGet(mirrorURL+"/"+requestedVersion, nil)
+	if errURL != nil {
+		log.Printf("[Error] : Getting url: %v", errURL)
+		os.Exit(1)
```
yermulnik, line 153: Wouldn't `log.Fatal()` be a better fit instead of `log.Printf()` + `os.Exit()`?

lib/list_versions_test.go
```diff
@@ -2,23 +2,23 @@ package lib_test
...
 )
 
 const (
-	hashiURL = "https://releases.hashicorp.com/terraform/"
+	hashiURL = "https://api.releases.hashicorp.com/v1/releases/terraform"
 )
 
 // TestGetTFList : Get list from hashicorp
+
+/*
```
yermulnik: Out of interest: why need to keep this commented out function?

lib/list_versions.go
Jukie: Agreed, fixed.

lib/list_versions.go
Jukie: Agreed, granted this wasn't my code :) but I've updated it everywhere now.

lib/list_versions.go
Jukie: Because here I'm achieving a while loop vs a single conditional compare. I'm intentionally continuing the loop until the list is smaller than `limit`, which is the point we know we're done.

lib/list_versions.go
Jukie: Yeah it does exactly that. I didn't add that code but will clean this up too.

lib/list_versions_test.go
Jukie: We don't :)
That was only there while it was still WIP

Jukie: Should be fixed now

lib/list_versions.go
Jukie: Added, thanks!

lib/list_versions.go
Jukie: Added

Jukie:
@warrensbox could you please take a look as well?

lib/list_versions.go
yermulnik: Got you. Thanks for the eplanation.

lib/list_versions.go
yermulnik: Thanks @Jukie 🙇🏻 Sorry I didn't get to figure out which code was added by you and which you didn't add 🤷🏻

lib/list_versions.go
```diff
@@ -1,142 +1,165 @@
...
 }
 
-//GetTFLatest :  Get the latest terraform version given the hashicorp url
-func GetTFLatest(mirrorURL string) (string, error) {
+func getReleases(mirrorURL string, queryParams map[string]string) ([]Release, error) {
+	var releases []Release
+	resp, errURL := httpGet(mirrorURL, queryParams)
+	if errURL != nil {
+		log.Printf("[Error] : Getting url: %v", errURL)
+		os.Exit(1)
```
yermulnik, line 88: `log.Fatalf()`?
Or am I missing something essential?

lib/list_versions.go
```diff
@@ -1,142 +1,165 @@
...
-	r, _ := regexp.Compile(semver)
-	for i := range result {
-		if r.MatchString(result[i]) {
-			str := r.FindString(result[i])
-			trimstr := strings.Trim(str, "/\"") //remove '/' or '"' from /X.X.X/" or /X.X.X"
-			return trimstr, nil
-		}
+	if resp.StatusCode != http.StatusOK {
+		log.Printf("[Error] : Retrieving contents from url: %s", mirrorURL)
+		os.Exit(1)
```
yermulnik, line 95: `log.Fatalf()`?
Or am I missing something essential?

yermulnik (commented):
Looks good (apart from two minor comments below), though I'm not quite into Golang so would better @warrensbox has this reviewed also.

lib/list_versions.go
Jukie: Looks like I missed a few places. Will take care of these as well.

lib/list_versions.go
Jukie: Looks like I missed a few places. Will take care of these as well.

lib/list_versions_test.go
```diff
@@ -9,31 +9,30 @@ import (
...
-	hashiURL = "https://releases.hashicorp.com/terraform/"
+	hashiURL = "https://api.releases.hashicorp.com/v1/releases/terraform"
 )
 
 // TestGetTFList : Get list from hashicorp
-func TestGetTFList(t *testing.T) {
 
-	listAll := true
-	list, _ := lib.GetTFList(hashiURL, listAll)
+func TestGetTFReleases(t *testing.T) {
```
Jukie, line 17: My [last commit](https://github.com/warrensbox/terraform-switcher/pull/227/commits/1adaf20372c64cf17c177fd038ddbeca949eff68) broke this test case so it needs reworked. Will work on this later tonight.

yermulnik:
@Jukie It starts to seem you're about to re-write the entire code 😂

lib/install.go
```diff
@@ -95,8 +95,14 @@ func Install(tfRelease *Release, binPath string, mirrorURL string) {
...
-	tfver, _ := version.NewVersion(tfRelease.Version)
-	tf102, _ := version.NewVersion(tfDarwinArm64StartVersion)
+	tfver, err := version.NewVersion(tfRelease.Version)
+	if err != nil {
+		log.Fatalf("Error generating terraform version: %s", err)
+	}
+	tf102, err := version.NewVersion(tfDarwinArm64StartVersion)
+	if err != nil {
+		log.Fatalf("Error generating terraform version: %s", err)
+	}
```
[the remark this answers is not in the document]
Jukie: Thanks, was already planning on coming back to this today.

main.go
```diff
@@ -398,30 +390,33 @@ func usageMessage() {
...
-		fmt.Println("[ERROR] : List is empty")
-		os.Exit(1)
+		log.Fatalln("[ERROR] : List is empty")
 	}
 	/* prompt user to select version of terraform */
 	prompt := promptui.Select{
 		Label: "Select Terraform version",
-		Items: tflist,
+		Items: versions,
 	}
```
Jukie, line 409: Is there a way in Go to prompt the version from within `tflist` rather than creating a separate slice here?

main.go
```diff
@@ -398,30 +390,33 @@ func usageMessage() {
...
 	if len(tflist) == 0 {
-		fmt.Println("[ERROR] : List is empty")
-		os.Exit(1)
+		log.Fatalln("[ERROR] : List is empty")
 	}
 	/* prompt user to select version of terraform */
 	prompt := promptui.Select{
 		Label: "Select Terraform version",
-		Items: tflist,
+		Items: versions,
```
Jukie, line 408: Is there a way in Go to prompt the version from within `tflist` rather than creating a separate slice?
https://github.com/warrensbox/terraform-switcher/blob/22720e86abd77e94b43e7f6cde8e1fa0c4cea1f8/main.go#L395-L409

Jukie:
Should be ready to go @warrensbox @yermulnik

lib/list_versions.go
```diff
@@ -1,142 +1,166 @@
...
 
-	return result, nil
+//removePreReleases : Removes any prerelease versions from a given slice of Release.
+func removePreReleases(releases []*Release) []*Release {
+	for i, r := range releases {
+		if r.IsPrerelease {
+			releases = append(releases[:i], releases[i+1:]...)
+		}
+	}
+	return releases
```
natebrennand, line 163: I don't think this is working as you expect it to, this pattern can work if you do it in reverse, but not iterating forward:
https://go.dev/play/p/P9OiqbT-HRp



A simpler pattern is to just build up a new slice as you go
```suggestion
//removePreReleases : Removes any prerelease versions from a given slice of Release.
func removePreReleases(releases []*Release) []*Release {
	realReleases := []*Release{}
	for i, r := range releases {
		if !r.IsPrerelease {
			realReleases = append(realReleases, releases[i])
		}
	}
	return realReleases
```

lib/install.go
```diff
@@ -76,10 +73,10 @@ func GetInstallLocation() string {
...
 //Install : Install the provided version in the argument
-func Install(tfversion string, binPath string, mirrorURL string) {
+func Install(tfRelease *Release, binPath string) {
 
-	// if !ValidVersionFormat(tfversion) {
-	// 	fmt.Printf("The provided terraform version format does not exist - %s. Try `tfswitch -l` to see all available versions.\n", tfversion)
+	// if !ValidVersionFormat(tfRelease) {
+	// 	fmt.Printf("The provided terraform version format does not exist - %s. Try `tfswitch -l` to see all available versions.\n", tfRelease)
 	// 	os.Exit(1)
 	// }
```
natebrennand, line 81: should this just be deleted?

lib/list_versions.go
```diff
@@ -1,142 +1,166 @@
...
+	}
+	releases, err := GetTFReleases(mirrorURL, preRelease)
+	if err != nil {
+		return nil, err
+	}
+	semv, err := SemVerParser(&version, releases)
+	if err != nil {
+		return nil, err
 	}
+	return semv, nil
```
natebrennand, line 57:
```suggestion
	return semv, err
```

lib/list_versions.go
```diff
@@ -1,142 +1,166 @@
...
+	queryParams := map[string]string{"limit": strconv.Itoa(limit)}
+	releaseSet, err := getReleases(mirrorURL, queryParams)
+	if err != nil {
+		return nil, err
+	}
+	var releases []*Release
+	releases = append(releases, releaseSet...)
+	for len(releaseSet) == limit {
+		queryParams["after"] = releaseSet[len(releaseSet)-1].TimestampCreated.String()
+		releaseSet, err = getReleases(mirrorURL, queryParams)
```
natebrennand, line 120: This is only going to query up to 2 pages of results, right?
Is there an implicit guarantee that this will be sufficient? Or should this be refactored to loop until the results are exhausted?

lib/list_versions.go
```diff
@@ -1,142 +1,166 @@
...
+		return nil, err
 	}
-
-	body, errBody := ioutil.ReadAll(resp.Body)
-	if errBody != nil {
-		log.Printf("[Error] : reading body: %v", errBody)
-		os.Exit(1)
-		return nil, errBody
+	var release *Release
+	if err := json.Unmarshal(body.Bytes(), &release); err != nil {
```
natebrennand, line 149: Do we really need to dereference `release` here? It's already a pointer.
Seems like it's unnecessary, but functional:
https://go.dev/play/p/wVqmov9406V

lib/list_versions.go
```diff
@@ -1,142 +1,166 @@
...
+		return nil, fmt.Errorf("%s: %s", err, body.String())
```
natebrennand, line 150: If `String() string` is defined on a value being sent to a `%s` directive, it will be called automatically by the `fmt` package.

```suggestion
		return nil, fmt.Errorf("%s: %s", err, body)
```


```go
	// Types that implement Stringer are printed the same as strings. Because
	// Stringers return a string, we can print them using a string-specific
	// verb such as %q.
	now := time.Unix(123456789, 0).UTC() // time.Time implements fmt.Stringer.
	fmt.Printf("%v %q\n", now, now)
	// Result: 1973-11-29 21:33:09 +0000 UTC "1973-11-29 21:33:09 +0000 UTC"
```
https://pkg.go.dev/fmt#example-package-Formats

main.go
natebrennand: I haven't used this library before, but it kinda looks like you can do some templating, would probably be overriding it to be `"{{ .Version }}"`.
https://pkg.go.dev/github.com/manifoldco/promptui#Select

But you seem to be using doing some deduping of data coming form `tfList`, so it's not clear that you'd want to use `tfList` without that filtering?

main.go
```diff
@@ -289,32 +294,27 @@ func installVersion(arg string, custBinPath *string, mirrorURL *string) {
 		}
 
 		//if the requested version had not been downloaded before
-		listAll := true                                     //set list all true - all versions including beta and rc will be displayed
-		tflist, _ := lib.GetTFList(*mirrorURL, listAll)     //get list of versions
-		exist := lib.VersionExist(requestedVersion, tflist) //check if version exist before downloading it
+		tfRelease, _ := lib.GetTFRelease(*mirrorURL, requestedVersion) //get list of versions
```
natebrennand, line 297: gotta check this error

main.go
```diff
@@ -339,19 +339,13 @@ func fileExists(filename string) bool {
 func checkTFModuleFileExist(dir string) bool {
 
 	module, _ := tfconfig.LoadModule(dir)
```
natebrennand, line 341: probably want to check this error

main.go
```diff
@@ -339,19 +339,13 @@ func fileExists(filename string) bool {
...
-	if len(module.RequiredCore) >= 1 {
-		return true
-	}
-	return false
+	return len(module.RequiredCore) >= 1
 }
 
 // checkTFEnvExist - checks if the TF_VERSION environment variable is set
 func checkTFEnvExist() bool {
 	tfversion := os.Getenv("TF_VERSION")
```
natebrennand, line 347: just keep reducing it:

```go
return os.Getenv("TF_VERSION") != ""
```

main.go
```diff
@@ -249,23 +255,22 @@ func installLatestImplicitVersion(requestedVersion string, custBinPath, mirrorUR
...
 	}
 	fmt.Printf("Error parsing constraint: %s\n", err)
 	lib.PrintInvalidMinorTFVersion()
 }
 
 // show latest - argument (version) must be provided
-func showLatestImplicitVersion(requestedVersion string, custBinPath, mirrorURL *string, preRelease bool) {
+func showLatestImplicitVersion(requestedVersion string, mirrorURL *string, preRelease bool) {
 	if lib.ValidMinorVersionFormat(requestedVersion) {
 		tfversion, _ := lib.GetTFLatestImplicit(*mirrorURL, preRelease, requestedVersion)
```
natebrennand, line 269: Probably want to check this error?
The following line is liable to get a nil pointer error if it fails.

main.go
```diff
@@ -398,30 +390,33 @@ func usageMessage() {
 /* listAll = true - all versions including beta and rc will be displayed */
 /* listAll = false - only official stable release are displayed */
 func installOption(listAll bool, custBinPath, mirrorURL *string) {
-	tflist, _ := lib.GetTFList(*mirrorURL, listAll) //get list of versions
-	recentVersions, _ := lib.GetRecentVersions()    //get recent versions from RECENT file
-	tflist = append(recentVersions, tflist...)      //append recent versions to the top of the list
-	tflist = lib.RemoveDuplicateVersions(tflist)    //remove duplicate version
+	tflist, _ := lib.GetTFReleases(*mirrorURL, listAll) //get list of versions
+	recentVersions, _ := lib.GetRecentVersions()        //get recent versions from RECENT file
```
natebrennand, line 394: should check these errors

main.go
```diff
@@ -398,30 +390,33 @@ func usageMessage() {
...
 
 	_, tfversion, errPrompt := prompt.Run()
 	tfversion = strings.Trim(tfversion, " *recent") //trim versions with the string " *recent" appended
 
 	if errPrompt != nil {
-		log.Printf("Prompt failed %v\n", errPrompt)
-		os.Exit(1)
+		log.Fatalf("Prompt failed %v\n", errPrompt)
 	}
+	rel, _ := lib.GetTFRelease(*mirrorURL, tfversion)
```
natebrennand, line 417: last place to check the error

natebrennand (changes requested):
This is mostly nits and Go commentary, but we gotta fix the bug identified in `removePreReleases`

lib/list_versions.go
```diff
@@ -1,142 +1,165 @@
...
-		fmt.Printf("Cannot get list from mirror: %s\n", mirrorURL)
+//GetTFLatestImplicit :  Get the latest implicit terraform version given the hashicorp url
+func GetTFLatestImplicit(mirrorURL string, preRelease bool, version string) (Release, error) {
+	if preRelease {
+		version = fmt.Sprintf(`%s{1}\.\d+\-[a-zA-z]+\d*`, version)
+	} else if !preRelease {
+		version = fmt.Sprintf("~> %v", version)
+	}
+	releases, err := GetTFReleases(mirrorURL, preRelease)
+	//version := fmt.Sprintf(`%s{1}\.\d+\-[a-zA-z]+\d*`, version)
```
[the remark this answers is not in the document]
Jukie: Think you're review may have been outdated, this is removed now.

main.go
Jukie: Added, thanks!

lib/list_versions.go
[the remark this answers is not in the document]
natebrennand: oops, tried to delete all the outdated comments but evidently missed a few

main.go
Jukie: Should be fixed now

main.go
Jukie: Should be fixed now

main.go
Jukie: Thanks, added

main.go
Jukie: Thanks, added

lib/install.go
```diff
@@ -2,14 +2,12 @@ package lib
 
 import (
 	"fmt"
+	"github.com/hashicorp/go-version"
```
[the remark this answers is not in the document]
Jukie: Thanks for the tip, updated.

lib/list_versions.go
Jukie: This should already loop over all pages, are you seeing different results?
It should continue looping while the size of results is equal to the specified `limit`, after which we know we've reached the last page.

natebrennand: nope, just haven't had coffee and hopelessly misread the `for` as `if` 🤦‍♂️

lib/list_versions.go
Jukie: I was doing this so I could later return it as a nil'able pointer value. I'd be open to changing it, just don't quite understand what you'd be suggesting.

lib/list_versions.go
```diff
@@ -1,142 +1,165 @@
...
+		log.Fatalf("%s: %s", err, body.String())
+	}
+	return releases, nil
 }
 
-//GetTFLatestImplicit :  Get the latest implicit terraform version given the hashicorp url
-func GetTFLatestImplicit(mirrorURL string, preRelease bool, version string) (string, error) {
+func GetTFReleases(mirrorURL string, preRelease bool) ([]Release, error) {
+	limit := 20
+	queryParams := map[string]string{"limit": strconv.Itoa(limit)}
```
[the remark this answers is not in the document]
Jukie: I'll look into this

lib/list_versions.go
Jukie: No worries mate!

main.go
Jukie: Fixed

lib/list_versions.go
natebrennand: Since we don't need to dereference this, we shouldn't (to keep it simpler)
```suggestion
	if err := json.Unmarshal(body.Bytes(), release); err != nil {
```

The playground link was just to display that these are equivalent.

lib/list_versions.go
Jukie: Ah nice, thanks!

main.go
Jukie: I'd rework the removedup function if I could get this part working.

Jukie: Fixed! Thanks for the suggestions

lib/list_versions.go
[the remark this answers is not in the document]
Jukie: @natebrennand Should be fixed with the last commit if you can take another look: https://github.com/warrensbox/terraform-switcher/pull/227/commits/23a6aa8fb44e2f9d3f5f33b996140a89f6fb3ca5

Jukie:
Found a bug after moving to semverparser for prereleases. Will come back to this later.

lib/list_versions.go
```diff
@@ -102,7 +102,7 @@ func GetTFReleases(mirrorURL string, preRelease bool) ([]*Release, error) {
 	limit := 20
 	u, err := url.Parse(mirrorURL)
 	if err != nil {
-		return nil, fmt.Errorf("Error parsing url: %s", err)
+		return nil, fmt.Errorf("[Error] : parsing url: %s", err)
```
yermulnik, line 105: This is not essential, though, if I got things right, when using `%q` (instead of `%s`) we can get properly quoted strings, which may help to identify start and end of the original error message per se to visually abstract it from fancy message we put right before it. Leaving this for your consideration since this doesn't bring much of benefit, but rather somewhat more of a readability for end users.

natebrennand (commented):
Read through the updates to my comments.
From a code review perspective, it looks pretty good.

I think @yermulnik is right that a lot of the error reporting could be improved.
An easier way to do this might be through explicitly wrapped errors:
https://pkg.go.dev/fmt#Errorf

> If the format specifier includes a %w verb with an error operand, the returned error will implement an Unwrap method returning the operand. It is invalid to include more than one %w verb or to supply it with an operand that does not implement the error interface. The %w verb is otherwise a synonym for %v.

A non-standard library provides some alternative API's for this:
https://pkg.go.dev/github.com/pkg/errors

----

> Am really considering reworking that by parsing the original json and input params into semver directly but it's probably not worth the time dump.

Can always come back and get that cleanup in, probably better to get this (already substantial) change merged.

main.go
```diff
@@ -398,37 +403,49 @@ func usageMessage() {
...
+	if err != nil {
+		log.Fatalf("Error while reading local versions file: %v", err)
+	}
+	tflist = append(recentVersions, tflist...)
+	tflist = lib.RemoveDuplicateVersions(tflist)
+
+	templates := promptui.SelectTemplates{
+		Active:   `> {{ .Version | green | bold }}`,
+		Inactive: `{{ .Version | cyan }}`,
+		Selected: `{{ "✔" | green | bold }} {{ .Version | cyan }}`,
```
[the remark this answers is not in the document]
yermulnik: Just to keep compatibility with dump terminals can we make `✔` configurable via command line option like `--ascii` (as in when option is provided use ASCII char instead of `✔`? Or to not complicate code just replace it with some 7bit ASCII char? (I'm not sure which exact char would be a good fir though)

lib/list_versions.go
```diff
@@ -1,172 +1,188 @@
...
-		semv, err := SemVerParser(&version, tflist)
+//GetTFReleases :  Get all terraform releases given the hashicorp url
+func GetTFReleases(mirrorURL string, preRelease bool) ([]*Release, error) {
+	limit := 20
+	u, err := url.Parse(mirrorURL)
+	if err != nil {
+		return nil, fmt.Errorf("[Error] : parsing url: %s", err)
+	}
+	values := u.Query()
+	values.Add("limit", strconv.Itoa(limit))
```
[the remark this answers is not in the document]
Jukie: This should already be done but I may need to push my local branch.

main.go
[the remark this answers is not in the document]
Jukie: I doubt many users will have a strong desire for customizing this but good point on making it a safer character.

Jukie:
Moved back to the prior implementation for prerelease filtering so the PR should be ready to go as is.
Will follow up with the suggestions for more centralized error handling but I don't think that's a blocker.

I've also forked Hashicorp's go-version lib to add support for interacting with individual semver elements like min, major, patch. That should allow for a common semver function here that will make comparing prereleases easier.
I thought about adding a lib that already implements these features but given terraform uses the hashicorp one I'd rather keep it consistent.

lib/list_versions.go
Jukie: I'm away for the weekend and didn't want to be on my computer (on mobile now, even pushed the last commit from my phone haha) too much so I'd rather address some of the cleanup and improvements in a follow up PR if you don't see it as a blocker.
That or if it's not reviewed by Tuesday I'll add these in.

Jukie: Update with https://github.com/warrensbox/terraform-switcher/commit/7346f8d07bbf111c65171769e9eec3b13839b8cb

main.go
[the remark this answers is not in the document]
Jukie: @yermulnik updated with https://github.com/warrensbox/terraform-switcher/commit/7346f8d07bbf111c65171769e9eec3b13839b8cb

main.go
```diff
@@ -416,9 +416,9 @@ func installOption(listAll bool, custBinPath, mirrorURL *string) {
 	tflist = lib.RemoveDuplicateVersions(tflist)
 
 	templates := promptui.SelectTemplates{
-		Active:   `> {{ .Version | green | bold }}`,
+		Active:   `»{{ .Version | green | bold }}`,
 		Inactive: `{{ .Version | cyan }}`,
-		Selected: `{{ "✔" | green | bold }} {{ .Version | cyan }}`,
+		Selected: `{{ "»" | green | bold }} {{ .Version | green }}`,
```
yermulnik, line 421: Well, the `»` isn't exactly ASCII too. It is UTF-8. So basically `»` isn't any safer than `✔` 😿 
ASCII is 7-bit and 128 chars long table and hence has quite limited number of chars we can use for the selector. And this is why I suggested to have a switch to turn off using chars from the extended table (UTF-8) for dumb terminals along with not limiting users with UTF-compatible terminal.

What we can use as "safe" chars for selector is e.g. `> ` and for the selected item is e.g. `+` or `~`.

Let's revert this back to `>` and `✔` and deal with this later as this is not essential and is basically out of scope of this PR. Thanks.

main.go
```diff
@@ -416,8 +416,8 @@ func installOption(listAll bool, custBinPath, mirrorURL *string) {
 	tflist = lib.RemoveDuplicateVersions(tflist)
 
 	templates := promptui.SelectTemplates{
-		Active:   `»{{ .Version | green | bold }}`,
-		Inactive: `{{ .Version | cyan }}`,
+		Active:   `{{ "»" | green | bold }} {{ .Version | green | bold }}`,
```
yermulnik, line 419: I feel like `»` being outside parenthesis was on purpose: to visually separate currently selected version and other versions available though not yet selected.

main.go
Jukie: Can you link me to a proper ascii code list then? I pulled that from [here](https://theasciicode.com.ar/) so thought it would be safe. Worst case I'll go back to `>` as you suggested.

yermulnik: Oh, my bad. Apologies, @Jukie. I missed the extended ASCII (the one up to 256 chars). Sorry for the noise. Just disregard my comment about `»`. Thanks.

Jukie: No worries, appreciate the concern!

main.go
yermulnik: With "outside" I meant that this char doesn't get styled with color and bold font.
To be honest I don't really get which selection menu (?) this is about and hence can't see it myself =(

Jukie: Is that built from my branch? It should look something like this.
Screen Shot 2022-05-16 at 8 15 35 AM

Jukie: So before this commit `»` wasn't styled in the prompt so this brings it in line with the output of the versioning.

yermulnik: Oh, my. I should have better not get myself in front of my laptop today at all 🤦🏻 Please disregard 😺 Going to get myself some more coffee probably 🤣

lib/list_versions.go
warrensbox: Go ahead. I will review this over the weekend (May 21 and 22)

Jukie: Updated with some additional error formatting suggestions. Please take another look @yermulnik if this is sufficient.

lib/list_versions.go
```diff
@@ -1,172 +1,194 @@
...
-	// Getting versions from body; should return match /X.X.X/ where X is a number
-	semver := `\/(\d+\.\d+\.\d+)\/?"`
-	r, _ := regexp.Compile(semver)
-	for i := range result {
-		if r.MatchString(result[i]) {
-			str := r.FindString(result[i])
-			trimstr := strings.Trim(str, "/\"") //remove '/' or '"' from /X.X.X/" or /X.X.X"
-			return trimstr, nil
-		}
+	if resp.StatusCode != http.StatusOK {
```
yermulnik, line 93: Is this something expected? 🤔 
image

lib/list_versions.go
yermulnik: @Jukie This is awesome huge piece of work! Thank you.

lib/list_versions.go
Jukie: Nope that was intended to be deleted, cleaning it up now!

Jukie:
I could use a hand or input on fixing the semver tests: https://github.com/warrensbox/terraform-switcher/pull/227/commits/f669501b06ac640440558d5058cf03fce33d39ad

I can't figure out why the constraints aren't being parsed correctly but it's probably staring me in the face....
Manually testing the CLI is successful.

```
$ go test ./lib -'run=^TestSemverParserCase2$'
Version does not exist or invalid terraform version format.
 Format should be #.#.# or #.#.#-@# where # are numbers and @ are word characters.
 For example, 0.11.7 and 0.11.9-beta1 are valid versions
--- FAIL: TestSemverParserCase2 (0.00s)
    semver_test.go:53: "error parsing constraint: ~> 1.1.0"
FAIL
FAIL    github.com/warrensbox/terraform-switcher/lib    0.693s
FAIL
```

Jukie:
Rebased over 0.14 release branch

lib/semver.go
```diff
@@ -2,52 +2,37 @@ package lib
 
 import (
 	"fmt"
-	"sort"
 
 	semver "github.com/hashicorp/go-version"
 )
 
 // GetSemver : returns version that will be installed based on server constaint provided
```
yermulnik, line 9:
```suggestion
// GetSemver : returns version that will be installed based on server constraint provided
```

main.go
```diff
@@ -59,7 +59,7 @@ func main() {
 	showLatestStable := getopt.StringLong("show-latest-stable", 'S', defaultLatest, "Show latest implicit version. Ex: tfswitch --show-latest-stable 0.13 prints 0.13.7 (latest)")
 	latestFlag := getopt.BoolLong("latest", 'u', "Get latest stable version")
 	showLatestFlag := getopt.BoolLong("show-latest", 'U', "Show latest stable version")
-	mirrorURL := getopt.StringLong("mirror", 'm', defaultMirror, "Install from a remote other than the default. Default: https://releases.hashicorp.com/terraform")
+	mirrorURL := getopt.StringLong("mirror", 'm', defaultMirror, "Install from a remote API other than the default. Default: https://api.releases.hashicorp.com/v1/releases/terraform")
```
yermulnik, line 62: Can a reference to `defaultMirror` from `const()` be used here instead of plain `https://api.releases.hashicorp.com/v1/releases/terraform` (to decrease repetition)?

yermulnik (approved):
LGTM. Minor comments below are not essential though.

main.go
Jukie: Updated

lib/semver.go
yermulnik: Yep, I know. Appreciate all the someone else's bits you've got fixed in scope of this straightly say refactoring PR 🙇🏻‍♂️

yermulnik:
@Jukie FYI the conflicts since recent updates to main branch.

Jukie:
@yermulnik already on it! Thanks, should be back to green.

PR #229, opened 2022-05-11
Fixes Hashicorp link issue

warrensbox:


warrensbox:
Release 0.13.1221 should fix this.

PR #230, opened 2022-05-12
MacOs -> macOS to be consistent with the standard casing

erikdw:
Just happened to notice this today. 

The 's' being lowercase instead of uppercase is what caught my eye, but apparently the 'm' is supposed to be lowercase.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -73,7 +73,7 @@ brews:
 
     # Caveats for the user of your binary.
     # Default is empty.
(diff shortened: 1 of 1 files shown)


yermulnik (approved):
LGTM. Good catch. Thanks @erikdw

@warrensbox Are you fine with me merging it?

warrensbox:
> LGTM. Good catch. Thanks @erikdw
> 
> @warrensbox Are you fine with me merging it?

Yes, seems like a minor change

PR #240, opened 2022-05-24
Fix chDirPath option for absolute paths

### LINKED ISSUE #222: --chdir with absolute project paths not working
Due to implementation we are not able to pass absolute paths to tfswitch:

https://github.com/warrensbox/terraform-switcher/blob/092ccf5400887b7e02e296a8ff507fb2f6f7c4eb/main.go#L74-L76

Before discovering `chdir` option we had to change dir at a sub shell, but were happy to discover a better way to reference terraform project directories and circumvent the problems involved by using subshells in bash scripts.

Jukie:
Closes #222
There's no need to add a prefix in the code because the current path would already be respected by default. In turn, having this prefix breaks absolute paths during execution (i.e. tfswitch -c /some/other/path) as described in #222 and demonstrated below.

Current behavior:
```
### works correctly 
$ ./tfswitch -c test-data/test_tfswitchtoml
Reading configuration from current directory for .tfswitch.toml
Switched terraform to version "0.11.3" 

### chDirPath is incorrectly being read as: /Users/Jukie/terraform-switcher//Users/Jukie/terraform-switcher/test-data/test_tfswitchtoml
$ ./tfswitch -c $PWD/test-data/test_tfswitchtoml
? Select Terraform version: 
  ▸ 0.11.3 *recent
    1.2.1
    1.2.0
    1.1.9
↓   1.1.8
```

After fixes:
```
### still works
$ ./tfswitch -c test-data/test_tfswitchtoml
Reading configuration from current directory for .tfswitch.toml
Switched terraform to version "0.11.3" 

### chDirPath is correctly being read as: /Users/Jukie/terraform-switcher/test-data/test_tfswitchtoml
$ ./tfswitch -c $PWD/test-data/test_tfswitchtoml
Reading configuration from current directory for .tfswitch.toml
Switched terraform to version "0.11.3" 
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -72,7 +72,7 @@ func main() {
 	homedir := lib.GetHomeDirectory()
 
 	if *chDirPath != "" {
-		dir = dir + "/" + *chDirPath
+		dir = *chDirPath
 	}
 
 	TFVersionFile := filepath.Join(dir, tfvFilename)           //settings for .terraform-version file in current directory (tfenv compatible)
--- a/main.go
+++ b/main.go
@@ -51,6 +51,7 @@ const (
 var version = "0.12.0\n"
 
 func main() {
+	dir := lib.GetCurrentDirectory()
 	custBinPath := getopt.StringLong("bin", 'b', lib.ConvertExecutableExt(defaultBin), "Custom binary path. Ex: tfswitch -b "+lib.ConvertExecutableExt("/Users/username/bin/terraform"))
 	listAllFlag := getopt.BoolLong("list-all", 'l', "List all versions of terraform - including beta and rc")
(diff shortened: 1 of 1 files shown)


Jukie:
CC @yermulnik @warrensbox

yermulnik (approved):
LGTM
@warrensbox has there been any background for the previous implementation which prevented use of absolute paths with `chdir` cmdline opt that we're not aware of?

Jukie:
I tried testing with every combination of specifying directories I could think of to break things after my changes but I think this was just a bug.

yermulnik:
> I tried testing with every combination of specifying directories I could think of to break things after my changes but I think this was just a bug.

Yep, I can't think of any sensible use case also given that escaping from current working directory wasn't restricted. Let's wait for Warren's input overnight (by my TZ) and merge tomorrow if there's no objections from him (I'd like to ask to remind me about this PR tomorrow so that I merge it. thanks).

warrensbox (approved):

PR #249, opened 2022-06-15
:bug: Corrected bad installations if user doesn't input 2 dots in ver…

afreyermuth98:
…sion

The constraint with only one dot returns the first version in the list, the last published.

Steps to reproduce : 
* Check the different outputs of these 2 commands :
```
tfswitch -s 1.0
tfswitch -s 1.0.0
```
Indeed, `tfswitch -s 1.0` downloads terraform v 1.2.2 (at the day I'm writing this) but the tfswitch --help documentation about the flag -s or --latest-stable show that it downloads 0.13.7 when you input 0.13

### DIFF
--- a/lib/list_versions.go
+++ b/lib/list_versions.go
@@ -73,7 +73,6 @@ func GetTFLatest(mirrorURL string) (string, error) {
 
 //GetTFLatestImplicit :  Get the latest implicit terraform version given the hashicorp url
 func GetTFLatestImplicit(mirrorURL string, preRelease bool, version string) (string, error) {
-
 	if preRelease == true {
 		//TODO: use GetTFList() instead of GetTFURLBody
 		versions, error := GetTFURLBody(mirrorURL)
@@ -94,6 +93,9 @@ func GetTFLatestImplicit(mirrorURL string, preRelease bool, version string) (str
 			}
 		}
 	} else if preRelease == false {
+		if strings.Count(version, string(".")) == 1 {
+			version += ".0"
+		}
 		listAll := false
 		tflist, _ := GetTFList(mirrorURL, listAll) //get list of versions
 		version = fmt.Sprintf("~> %v", version)
--- a/lib/list_versions.go
+++ b/lib/list_versions.go
@@ -93,9 +93,6 @@ func GetTFLatestImplicit(mirrorURL string, preRelease bool, version string) (str
 			}
 		}
 	} else if preRelease == false {
-		if strings.Count(version, string(".")) == 1 {
-			version += ".0"
-		}
 		listAll := false
 		tflist, _ := GetTFList(mirrorURL, listAll) //get list of versions
 		version = fmt.Sprintf("~> %v", version)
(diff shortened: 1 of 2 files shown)


lib/list_versions.go
```diff
@@ -94,6 +93,9 @@ func GetTFLatestImplicit(mirrorURL string, preRelease bool, version string) (str
 			}
 		}
 	} else if preRelease == false {
+		if strings.Count(version, string(".")) == 1 {
```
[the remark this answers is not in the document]
warrensbox: Disregard this

Jukie:
I think the docs in the help output may be outdated. Isn't that behavior expected according to the terraform docs?
Also, was this behaving differently on older versions? I've tried going back a few versions but can't produce a different result.

[link](https://www.terraform.io/language/expressions/version-constraints#-3)
> ~>: Allows only the rightmost version component to increment. For example, to allow new patch releases within a specific minor release, use the full version number: ~> 1.0.4 will allow installation of 1.0.5 and 1.0.10 but not 1.1.0. This is usually called the pessimistic constraint operator.

We use that [here](https://github.com/warrensbox/terraform-switcher/blob/master/lib/list_versions.go#L99) so in turn `~> 0.13` maps to `>= 0.13.0, <1.0.0` which would correctly return 0.15.5.
Same with 1.0 vs 1.0.0 which would return `<2.0.0, >=1.0.0` and `<1.1.0, >=1.0.0` respectively.

warrensbox:
@afreyermuth98 
The existing implementation should behave as expected. 
I will need to update my documentation:

```
tfswitch -s 0.13
Matched version: 0.15.5

tfswitch -s 0.13.0
Matched version: 0.13.7

tfswitch -s 1.0
Matched version: 1.2.3

tfswitch -s 1.0.0
Matched version: 1.0.11
```

When the user passes: `1.0.0`, they are want the latest version of the `1.0.0` release
When the user passes: `1.0`, the are expecting the latest version for `1.0` - in this case, it would be `1.2.3`

afreyermuth98:
@warrensbox Indeed I think that an update of the documentation is more appropriate for this problem since the code uses the native constraint function of Go, which means that it must work like that.

main.go
```diff
@@ -56,7 +56,7 @@ func main() {
 	listAllFlag := getopt.BoolLong("list-all", 'l', "List all versions of terraform - including beta and rc")
 	latestPre := getopt.StringLong("latest-pre", 'p', defaultLatest, "Latest pre-release implicit version. Ex: tfswitch --latest-pre 0.13 downloads 0.13.0-rc1 (latest)")
 	showLatestPre := getopt.StringLong("show-latest-pre", 'P', defaultLatest, "Show latest pre-release implicit version. Ex: tfswitch --show-latest-pre 0.13 prints 0.13.0-rc1 (latest)")
-	latestStable := getopt.StringLong("latest-stable", 's', defaultLatest, "Latest implicit version. Ex: tfswitch --latest-stable 0.13 downloads 0.13.7 (latest)")
+	latestStable := getopt.StringLong("latest-stable", 's', defaultLatest, "Latest implicit version based on a constraint. Ex: tfswitch --latest-stable 0.13.0 downloads 0.13.7 and 0.13 downloads 0.15.5 (latest)")
```
Jukie, line 59: Minor nit, but could you make the text in parentheses more descriptive?
Possibly `(e.g. latest 0.x.x)`

Jukie: Disregard, I guess you didn't add that text.

Jukie (approved):

Jukie:
This looks good @warrensbox.
Was also thinking about different output to make it more descriptive without adding too much text but don't have any great suggestions at the moment.

warrensbox (approved):

main.go
warrensbox: The `based on a constraint` part is good
I might consider this too:
`Ex: tfswitch --latest-stable 1.0.0 downloads <1.1.0, >=1.0.0, 1.0 downloads <2.0.0, >=1.0.0"`

warrensbox (approved):

Issue #254, opened 2022-06-27
Update golang.org/x/text` and gopkg.in/yaml` version pinning to address CVEs


electricthunder:
There are a number of vulnerabilities in the following go libraries:

```bash
usr/local/bin/tfswitch (gobinary)

Total: 4 (UNKNOWN: 2, LOW: 0, MEDIUM: 1, HIGH: 1, CRITICAL: 0)

┌───────────────────┬────────────────┬──────────┬───────────────────┬───────────────┬──────────────────────────────────────────────────────────────┐
│      Library      │ Vulnerability  │ Severity │ Installed Version │ Fixed Version │                            Title                             │
├───────────────────┼────────────────┼──────────┼───────────────────┼───────────────┼──────────────────────────────────────────────────────────────┤
│ golang.org/x/text │ CVE-2021-38561 │ HIGH     │ v0.3.5            │ 0.3.7         │ golang: out-of-bounds read in golang.org/x/text/language     │
│                   │                │          │                   │               │ leads to DoS                                                 │
│                   │                │          │                   │               │ https://avd.aquasec.com/nvd/cve-2021-38561                   │
├───────────────────┼────────────────┼──────────┼───────────────────┼───────────────┼──────────────────────────────────────────────────────────────┤
│ gopkg.in/yaml.v2  │ CVE-2019-11254 │ MEDIUM   │ v2.2.2            │ 2.2.8         │ kubernetes: Denial of service in API server via crafted YAML │
│                   │                │          │                   │               │ payloads by...                                               │
│                   │                │          │                   │               │ https://avd.aquasec.com/nvd/cve-2019-11254                   │
├───────────────────┼────────────────┼──────────┼───────────────────┼───────────────┼──────────────────────────────────────────────────────────────┤
│ gopkg.in/yaml.v2  │ GMS-2019-2     │ UNKNOWN  │ v2.2.2            │ v2.2.3        │ XML Entity Expansion                                         │
├───────────────────┼────────────────┼──────────┼───────────────────┼───────────────┼──────────────────────────────────────────────────────────────┤
│ gopkg.in/yaml.v2  │ GO-2021-0061   │ UNKNOWN  │ v2.2.2            │ 2.2.3         │ Due to unbounded alias chasing, a maliciously crafted YAML   │
│                   │                │          │                   │               │ file                                                         │
│                   │                │          │                   │               │ can cause the...                                             │
[...]

PR #256, opened 2022-06-29
Create codeql-analysis.yml

jukie:
Adds code scanning for vulnerabilities like the ones discovered by a user in #254

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,72 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master", release ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "master" ]
+  schedule:
+    - cron: '23 1 * * 3'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'go' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
+        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v2
+      with:
+        languages: ${{ matrix.language }}
+        # If you wish to specify custom queries, you can do so here or in a config file.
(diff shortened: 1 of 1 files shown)


yermulnik:
1. Should we exclude non-Golang paths to [Avoid unnecessary scans of pull requests](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#avoiding-unnecessary-scans-of-pull-requests)? Like e.g. in this PR CodeQL sort of blocks this PR whereas this PR has nothing to do with the alerted file.
2. Also is there an option for CodeQL to ignore some code lines, blocks and snippets? I was able to only find how to ignore files, but couldn't find an option for inline comments to disable check for specific fragments of code 🤷🏻 Like for when we do want or need some flagged code block to remain as is.

Jukie:
Yeah I need to clean this up @yermulnik, converting to draft

MatrixCrawler:
I think this might not longer necessary as we have dependabot scanning our dependencies and suggests fixes.
Closing this PR

MatrixCrawler:
I rebased and updated the codeql defintion. Maybe we should go forward with this?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,93 @@
...
+        # To learn more about changing the languages that are analyzed or customizing the build mode for your analysis,
+        # see https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/customizing-your-advanced-setup-for-code-scanning.
+        # If you are analyzing a compiled language, you can modify the 'build-mode' for that language to customize how
+        # your codebase is analyzed, see https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/codeql-code-scanning-for-compiled-languages
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
```
yermulnik, line 65:
```suggestion
        uses: actions/setup-go@v5
```

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,93 @@
...
+          # For more details on CodeQL's query packs, refer to: https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs
+          # queries: security-extended,security-and-quality
+
+      # If the analyze step fails for one of the languages you are analyzing with
+      # "We were unable to automatically build your code", modify the matrix above
+      # to set the build mode to "manual" for that language. Then modify this step
+      # to build your code.
+      # ℹ️ Command-line programs to run using the OS shell.
+      # 📚 See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun
+      - run: go build -v ./...
```
yermulnik, line 88: Given we have `CGO_ENABLED=0` defined for `build` and `release` workflows, do we maybe need to add it to this workflow too?

yermulnik:
> Maybe we should go forward with this?

I'm fine with this going forward.
@jukie seems to has been inactive for some time 😢, so we may need to proceed on our own.

.github/workflows/codeql-analysis.yml
MatrixCrawler: Not sure if that affects the code scanning but i will add it anyway as it would reflect our build process.

yermulnik (approved):

yermulnik:
> Thanks @yermulnik! Looks like a lot of activity here, I'll start getting involved again

Yay, that's great 🥳

Issue #260, opened 2022-07-19
Won't install on Raspberry pi: "warrensbox/terraform-switcher crit platform linux/armv7 is not supported."


JohnSmith31:
Won't install on Raspberry pi: "warrensbox/terraform-switcher crit platform linux/armv7 is not supported."

JohnSmith31:
Hi. I installed using the command on the tfswitch web page:

curl -L https://raw.githubusercontent.com/warrensbox/terraform-switcher/release/install.sh | bash

The full output is:

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  9216  100  9216    0     0  19120      0 --:--:-- --:--:-- --:--:-- 19160
warrensbox/terraform-switcher crit platform linux/armv7 is not supported.  Make sure this script is up-to-date and file request at https://github.com/warrensbox/terraform-switcher/issues/new

JohnSmith31:
(BTW, that's running on a Raspberry Pi 4 on Raspberry Pi OS. uname -a gives:
Linux mel 5.10.17-v7l+ #1403 SMP Mon Feb 22 11:33:35 GMT 2021 armv7l GNU/Linux)

yermulnik:
@JohnSmith31 Would you mind trying `curl -L https://raw.githubusercontent.com/warrensbox/terraform-switcher/feature/add_more_platforms_to_install.sh/install.sh | bash` to see whether https://github.com/warrensbox/terraform-switcher/pull/261 fixes the issue for your platform? Thanks.

yermulnik:
@JohnSmith31 ?

JohnSmith31:
That certainly seemed to do the trick! Thanks :)

yermulnik:
Cool. Thanks for confirmation.
Closing this issue to continue in #261

PR #261, opened 2022-07-19
Add Linux ARMv{6,7} arch to `install.sh`

yermulnik:
- As seen from #260 Linux ARMv{6,7} are missing from `install.sh`
  despite binaries are built for these platforms in recent releases, so
  add these two to the pool.


Jukie (approved):

warrensbox:
Merged master but not released yet

PR #264, opened 2022-07-22
feat: use terragrunt parser to parse terragrunt.hcl

### LINKED ISSUE #259: Parse terragrunt.hcl and inherit terraform_version_constraint from parent
We're using terragrunt with a nested moule structure, meaning:
```
stages
└── test
    └── module1
    └── module2
└── prod
    └── module1
    └── module2
```

We have `terragrunt.hcl` in both `stages/{stage}`, to define global parameters, and in `stages/{stage}/{module}`, where we define the specific module variable.

We locked the `terraform_version_constraint` inside `stages/{stage}`. Running `tfswitch` from module does not work, we are prompted for a version, so we run it from the module folder.

From time to time we override the terraform_constraint in one specific module. In this scenario, we need to run `tfswitch` inside the module folder.

This is bad DX and causes issues with CI systems, since we do not have a predictable path to run `tfswitch` in.

It would be helpful to be able to parse the hcl file, so that we would inherit the parent declarations, and behave accordingly.

Each of our modules has an include:
```hcl
include {
  path = find_in_parent_folders()
}
```

so, by parsing HCL, the terraform_constraint_version may be inherited from the parent.

-

**Additional context**
-

marcoreni:
Fixes #259 .

This PR introduces the Terragrunt parser to handle `terragrunt.hcl` parsing, so that `include` blocks can be read and more complex logics of constraints can be handled (for example, having the constraint inside a parent file).

Pros:
- Uses Terragrunt logics, that is thoroughly tested on the terragrunt repo.
- The same mechanism can be added to `tgswitch` with minimal effort.

Cons:
- Since Terragrunt does not expose APIs, using it requires lots of libraries (+20MB in my sample compiled build) and requires `logrus` to be initialized to avoid crash.

LMKWYT


Tensho:
Hey 👋 I have the same use case with hierarchical terraform/terragrunt constraints definition.
```
terragrunt.hcl     # <= terraform_version_constraint = "1.1.6"
test
  component
    terragrunt.hcl # <= implicit include of the root terragrunt.hcl config, no explicit terraform_version_constraint definition
prod
  component
    terragrunt.hcl
```
As far as `tfswitch` already supports current `terragrunt.hcl` parsing I think it's reasonable to extend it further.

Another option could be checking `.tfswitchrc` file presence from the current folder to root one, a bottom-up search through a hierarchy.
```
.tfswitchrc     # <= 1.1.6
test
  component 
                # <= Switch to default 1.1.6
prod
  component
    .tfswitchrc # <= Switch to override 1.1.7
```

marcoreni:
@Tensho that is another possibility, yes. I went this way because by using the hcl I would need to declare the constraint once and I would be able to delete the `.tfswitchrc` files entirely

MatrixCrawler:
@marcoreni Could you rebase and update this PR to the latest terragrunt version? There seems to be a lot of changes since you opened that PR.
Also it would be good to have a test case for that.

yermulnik:
Closing as outdated and abandoned(?).
Let me know if it's still relevant and to be worked on.
Thank you.

Issue #165, opened 2021-05-06
tfswitch should accept a "--non-interactive" argument


toralf:
to avoid a hang in a pipeline.
The default might be to errorred out ?

jakubbujny:
+1 here - I can try to raise PR with such functionality if maintainers are open for it?

warrensbox:
@jakubbujny I'd be open for it

aucampia:
as a somewhat of a hackaround I'm piping yes into tfswitch:


```
root@4547b3eb3246:/srv/workspace# terraform version
Terraform v1.1.6
on linux_amd64

Your version of Terraform is out of date! The latest version
is 1.1.9. You can update by downloading from https://www.terraform.io/downloads.html
root@4547b3eb3246:/srv/workspace# tfswitch
Creating directory for terraform binary at: /root/.terraform.versions

2022/04/20 19:54:16 Prompt failed ^C
root@4547b3eb3246:/srv/workspace# ^C
root@4547b3eb3246:/srv/workspace# yes | tfswitch
✔ 1.1.9
Downloading to: /root/.terraform.versions
19262029 bytes downloaded
Switched terraform to version "1.1.9" 
root@4547b3eb3246:/srv/workspace# echo 1.1.2 > .terraform-version
root@4547b3eb3246:/srv/workspace# yes | tfswitch
Reading file .terraform-version 
Downloading to: /root/.terraform.versions
18687805 bytes downloaded
Switched terraform to version "1.1.2" 
```

Issue #150, opened 2021-03-11
Allow fallback version


johannes-mathes:
We define a version requirement in most terraform modules, because we want the full control about updates. We use tfswitch also in a pipeline, without any parameter as we want to use the feature that tfswitch parses the version constraint and uses this. However, if it is not defined, tfswitch asks interactively and fails. 
We would like to specify a fallback version for such a case -> no interactive questions, as this does not work in a scripted environment

warrensbox:
@johannes-mathes 
You have some parts where you have `terraform version` defined in the module, like this:
```
terraform {
  required_version = ">= 0.12.9"

  required_providers {
    aws        = ">= 2.52.0"
    kubernetes = ">= 1.11.1"
  }
}
```
And you have some places where terraform is not defined, am I correct?

If the terraform version is not defined, what would you suggest `tfswitch` to do? 
Someone else suggested to have a -q argument [here](https://github.com/warrensbox/terraform-switcher/pull/108).

johannes-mathes:
Yes, you are right. I would think a switch -q <FALLBACK_DEFAULT_VERSION>. 
Behaviour: 
If available, take terraform version specified by required_version.
Otherwise, take FALLBACK_DEFAULT_VERSION.

And by the way, your tool is very great. We use it allover our project.

calebplum:
+1 for this, our use case is also running tfswitch within a pipeline. We like to specify `required_version` in our templates but need the ability to fallback to the latest stable version if a template is missing `required_version`.

Being that the [order of precedence](https://github.com/warrensbox/terraform-switcher#order-of-precedence) has environment variables prioritised last it would make sense to be able to specify 'latest' or 'latest-stable' in the `TF_VERSION` environment variable and have it work in the same way as the `-u` and `-s` command-line flags. This would allow us to specify `required_version` in some projects and default to the latest stable when that's missing.

anherrera:
In addition to this, I'd like the option to return an error if no valid version was found. Currently, tfswitch doesn't return 1 when you specify an invalid version, only if you specify an invalid flag. I'd like a way of using tfswitch in an automated fashion but also enforce that developers are setting versions either in their tf files or in env vars, and then error out if not. 

```
╰─$ tfswitch butts              
Invalid terraform version format. Format should be #.#.# or #.#.#-@# where # are numbers and @ are word characters. For example, 0.11.7 and 0.11.9-beta1 are valid versions
Args must be a valid terraform version


[...]

PR #275, opened 2022-10-11
Default version flag added

sivaramsajeev:
This flag will allow us to specify a default version to install which will be useful for the CI systems instead of going into the prompt

### DIFF
--- a/main.go
+++ b/main.go
@@ -63,6 +63,7 @@ func main() {
 	mirrorURL := getopt.StringLong("mirror", 'm', defaultMirror, "Install from a remote API other than the default. Default: "+defaultMirror)
 	chDirPath := getopt.StringLong("chdir", 'c', dir, "Switch to a different working directory before executing the given command. Ex: tfswitch --chdir terraform_project will run tfswitch in the terraform_project directory")
 	versionFlag := getopt.BoolLong("version", 'v', "Displays the version of tfswitch")
+	defaultVersion := getopt.StringLong("default", 'd', defaultLatest, "Default to version to pick up. Ex: tfswitch --default 1.2.4")
 	helpFlag := getopt.BoolLong("help", 'h', "Displays help message")
 	_ = versionFlag
 
@@ -210,6 +211,10 @@ func main() {
 		tfversion := os.Getenv("TF_VERSION")
 		fmt.Printf("Terraform version environment variable: %s\n", tfversion)
 		installVersion(tfversion, custBinPath, mirrorURL)
+
+	/* if default version is provided - Pick this instead of going for prompt */
+	case *defaultVersion != "":
+		installVersion(*defaultVersion, custBinPath, mirrorURL)
 
 	// if no arg is provided
 	default:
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -63,6 +63,7 @@ func main() {
 	mirrorURL := getopt.StringLong("mirror", 'm', defaultMirror, "Install from a remote API other than the default. Default: "+defaultMirror)
 	chDirPath := getopt.StringLong("chdir", 'c', dir, "Switch to a different working directory before executing the given command. Ex: tfswitch --chdir terraform_project will run tfswitch in the terraform_project directory")
 	versionFlag := getopt.BoolLong("version", 'v', "Displays the version of tfswitch")
+	defaultVersion := getopt.StringLong("default", 'd', defaultLatest, "Default to version to pick up. Ex: tfswitch --default 1.2.4")
```
yermulnik, line 66:
```suggestion
	defaultVersion := getopt.StringLong("default", 'd', defaultLatest, "Default to this version in case no other versions could be detected. Ex: tfswitch --default 1.2.4")
```

yermulnik (commented):
Could you please outline what's the difference wuth `TF_VERSION` env var (https://github.com/warrensbox/terraform-switcher/blob/e77b48c2a9b44855846a19feb29b51123cb865d0/main.go#L209-L213) and why `TF_VERSION` cannot suffice in your use case?
Thanks.

main.go
sivaramsajeev: @yermulnik In cases where TF_VERSION is already set (`say 1.2.4`) in the environment, if we add `TF_VERSION=1.2.3 tfswitch` it'll install version `1.2.3` which is not intended. But `tfswitch -d 1.2.3` can really act as a default version.

sivaramsajeev:
@yermulnik In cases where TF_VERSION is already set (say 1.2.4) in the environment, if we add `TF_VERSION=1.2.3 tfswitch` it'll install version `1.2.3` which is not intended. But `tfswitch -d 1.2.3` can really act as a default version. 
Thanks

yermulnik:
@sivaramsajeev Aha, got you. The new option is intended to be a fallback.
Though could you please elaborate on how this is expected to work as I might be missing something essential since I'm not familiar with all code bits?
From what I see based on example with `TF_VERSION`: when this Env var is defined (let's suppose its value refers to a non-existing TF version so that fallback should be used) `tfswitch` triggers `installVersion()` function, which calls `os.Exit(1)` on errors (e.g. when requested TF version doesn't exist), which in its turn should interrupt `tfswitch` execution and hence would not return back into `switch` statement which handles command line options. What am I missing? 🤔

sivaramsajeev:
The `--default` option is just meant to be the version to be picked up if tfswicth is unable to get the intended version (Either the version is not set using one of the preceding methods OR tfswitch is unable to parse it ). This wont change the original behavior of exiting upon calling invalid version so it's not a fallback in that sense.  It's also calling the same installVersion function.

`TF_VERSION=x.x.x tfswicth` won't act like  `default` because it could potentially end up overriding the intended TF version env var. So if we give this one liner instead of something like `tfswitch -d x.x.x`, there is a chance that the intended TF version was also set by an environment variable explicitly or sourced during runtime & we end up overriding it with our one-liner.

yermulnik:
> Either the version is not set using one of the preceding methods OR tfswitch is unable to parse it

Aah, I see now. This is the same as requested in #150 and #165 and actually is, broadly speaking, a programmatic implementation of "hackaround" from https://github.com/warrensbox/terraform-switcher/issues/165#issuecomment-1104400753
Yep, this makes sense to me. You got an approval from me, though @warrensbox and @Jukie would better weigh in to review implementation details.

yermulnik (approved):

warrensbox:
I am still reviewing this

warrensbox (approved):

PR #287, opened 2023-02-06
Feature: Fallback or default version

warrensbox:
## Context
When using a CICD pipeline, you may want a default or fallback version to avoid the pipeline from hanging.

### Upgrade
N/A

### Added
N/A

### Removed
N/A

### Fixes:
#275 #150 #165 #108


warrensbox:
This has been approved by @yermulnik previously. I only fixed some minor README content.

PR #294, opened 2023-03-21
Fix init function

Jukie:
For environments without write access to /usr/local/bin, the existing function will fail because it attempts to remove the symlink if terraform already exists.
tfswitch should instead check for symlinks at the given binPath or default to user home directory.

### DIFF
--- a/lib/install.go
+++ b/lib/install.go
@@ -26,16 +26,16 @@ var (
 )
 
 // initialize : removes existing symlink to terraform binary// I Don't think this is needed
-func initialize() {
+func initialize(binPath string) {
 
 	/* Step 1 */
 	/* initilize default binary path for terraform */
 	/* assumes that terraform is installed here */
 	/* we will find the terraform path instalation later and replace this variable with the correct installed bin path */
-	installedBinPath := "/usr/local/bin/terraform"
+	installedBinPath := binPath + "terraform"
 
 	/* find terraform binary location if terraform is already installed*/
-	cmd := NewCommand("terraform")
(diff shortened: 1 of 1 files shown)


yermulnik (approved):

lib/install.go
```diff
@@ -25,17 +25,17 @@ var (
...
-func initialize() {
+// initialize : removes existing symlink to terraform binary based on provided binPath
+func initialize(binPath string) {
 
 	/* Step 1 */
 	/* initilize default binary path for terraform */
 	/* assumes that terraform is installed here */
 	/* we will find the terraform path instalation later and replace this variable with the correct installed bin path */
-	installedBinPath := "/usr/local/bin/terraform"
+	installedBinPath := binPath + "terraform"
```
yermulnik, line 35: Should we ensure the `binPath` has trailing slash?

Jukie: Good call

Jukie: We actually don't need this line at all because `/terraform` (or whatever a user desires as executable name) is already expected with the binPath flag.

yermulnik: Ah, so it's inherited from `custBinPath` which defaults to `defaultBin` which is `/usr/local/bin/terraform` via `main.go` (if I didn't miss something though 😛)

yermulnik:
This is not related to this PR, though I can see `defaultBin` definition in `lib/install.go` and can't make up my mind of why it is (re-)defined here (it is defined in `main.go`). Am I missing something?

yermulnik (approved):

Issue #154, opened 2021-03-31
Feature request: option to dry-run


yermulnik:
I'd like to ask to implement a feature which would allow me to run `tfswitch` in a dry-run mode where it only outputs whether it would switch terraform version in current directory or not with different exit codes for the two operations (e.g. `0` for when no switching is required and `1` for when `tfswitch` would switch versions).

My use case: we're managing different infrastructures from different repositories including legacy setups – most of these are using different versions of terraform and we're not planning to unify them. Most of the time I need to switch terraform version when I work on something in several different setups at the same time but it's not always that I need to actually switch terraform version (e.g. when I only need to cd into dir to check something) hence suggested `cdtfswitch()` Bash alias is not something I indeed need but would rather want to inject something like `tfswitch --dry-run >/dev/null || echo "tfswitch to use correct terraform version"` into my `PROMPT_COMMAND` env var so that I get notified when my Terraform version doesn't match Terraform version constraint every time I hit enter on the command line.

Thanks in advance.

yermulnik:
As an interim solution (sort of FWIW) here's the ugly bit of Go code I came up with:
```Go
package main

import (
	"fmt"
	"os"

	"github.com/hashicorp/go-version"
	"github.com/jessevdk/go-flags"
)

var opts struct {
	Verbose []bool `short:"v" long:"verbose" description:"Show verbose debug information"`
}

func usage() {
	fmt.Fprintf(os.Stderr, "Usage: %s [-v] <tf version> <tf version constraint>\n", os.Args[0])
	os.Exit(2)
}

func main() {
	args, err := flags.Parse(&opts)
	if err != nil {
		usage()
	}

	if len(args) != 2 {
		usage()
	}

	v1, err := version.NewVersion(args[0])
	if err != nil {
		fmt.Println("ERROR:", err)
		usage()
	}

	constraints, err := version.NewConstraint(args[1])
	if err != nil {
		fmt.Println("ERROR:", err)
		usage()
	}

	if constraints.Check(v1) {
		if len(opts.Verbose) > 0 {
			fmt.Printf("\"%s\" satisfies constraint \"%s\"\n", v1, constraints)
		}
		os.Exit(0)
	} else {
		if len(opts.Verbose) > 0 {
			fmt.Printf("\"%s\" DOES NOT satisfy constraint \"%s\"\n", v1, constraints)
		}
		os.Exit(1)
	}
}
```
```Bash
> terraform version | head -1
Terraform v0.13.6

[...]

PR #301, opened 2023-04-20
feat: allow to show required by module version explicitly

ikorchynskyi:
Proposed changes simplify the possibility to use `tfswitch` in automation: my own use-case is to fetch the required by project terraform version and to use specific version of docker image with terraform inside the CI/CD pipeline.

Related diagnostic messages has been moved to `stderr` to allow output to be fetched like
```
$ echo "Version required by module: $(go run main.go --chdir ~/foo/ --show-latest-required)"
Reading required version from terraform file
Reading required version from constraint: ~> 1.4.0, < 1.4.5
Matched version: 1.4.4
Version required by module: 1.4.4
```

### DIFF
--- a/lib/semver.go
+++ b/lib/semver.go
@@ -2,6 +2,7 @@ package lib
 
 import (
 	"fmt"
+	"os"
 	"sort"
 
 	semver "github.com/hashicorp/go-version"
@@ -12,7 +13,7 @@ func GetSemver(tfconstraint *string, mirrorURL *string) (string, error) {
 
 	listAll := true
 	tflist, _ := GetTFList(*mirrorURL, listAll) //get list of versions
-	fmt.Printf("Reading required version from constraint: %s\n", *tfconstraint)
+	fmt.Fprintf(os.Stderr, "Reading required version from constraint: %s\n", *tfconstraint)
 	tfversion, err := SemVerParser(tfconstraint, tflist)
 	return tfversion, err
 }
@@ -39,7 +40,7 @@ func SemVerParser(tfconstraint *string, tflist []string) (string, error) {
 	for _, element := range versions {
 		if constraints.Check(element) { // Validate a version against a constraint
 			tfversion = element.String()
-			fmt.Printf("Matched version: %s\n", tfversion)
+			fmt.Fprintf(os.Stderr, "Matched version: %s\n", tfversion)
 			if ValidVersionFormat(tfversion) { //check if version format is correct
 				return tfversion, nil
 			}
@@ -52,10 +53,10 @@ func SemVerParser(tfconstraint *string, tflist []string) (string, error) {
 
 // Print invalid TF version
 func PrintInvalidTFVersion() {
-	fmt.Println("Version does not exist or invalid terraform version format.\n Format should be #.#.# or #.#.#-@# where # are numbers and @ are word characters.\n For example, 0.11.7 and 0.11.9-beta1 are valid versions")
+	fmt.Fprintln(os.Stderr, "Version does not exist or invalid terraform version format.\n Format should be #.#.# or #.#.#-@# where # are numbers and @ are word characters.\n For example, 0.11.7 and 0.11.9-beta1 are valid versions")
 }
 
 // Print invalid TF version
 func PrintInvalidMinorTFVersion() {
-	fmt.Println("Invalid minor terraform version format. Format should be #.# where # are numbers. For example, 0.11 is valid version")
+	fmt.Fprintln(os.Stderr, "Invalid minor terraform version format. Format should be #.# where # are numbers. For example, 0.11 is valid version")
 }
--- a/main.go
+++ b/main.go
@@ -456,7 +456,7 @@ func installOption(listAll bool, custBinPath, mirrorURL *string) {
 func installTFProvidedModule(dir string, custBinPath, mirrorURL *string) {
 	fmt.Println("Reading required version from terraform file")
 	module, _ := tfconfig.LoadModule(dir)
-	if len(module.RequiredCore) >= 1 {
+	if len(module.RequiredCore) == 0 {
 		fmt.Println("The provided directory does not have required terraform version constraint.")
 		os.Exit(1)
 	}
--- a/main.go
+++ b/main.go
@@ -60,6 +60,8 @@ func main() {
 	showLatestStable := getopt.StringLong("show-latest-stable", 'S', defaultLatest, "Show latest implicit version. Ex: tfswitch --show-latest-stable 0.13 prints 0.13.7 (latest)")
 	latestFlag := getopt.BoolLong("latest", 'u', "Get latest stable version")
 	showLatestFlag := getopt.BoolLong("show-latest", 'U', "Show latest stable version")
+	latestRequiredFlag := getopt.BoolLong("latest-required", 'r', "Get latest required by module version")
+	showLatestRequiredFlag := getopt.BoolLong("show-latest-required", 'R', "Show latest required by module version")
 	mirrorURL := getopt.StringLong("mirror", 'm', defaultMirror, "Install from a remote API other than the default. Default: "+defaultMirror)
 	chDirPath := getopt.StringLong("chdir", 'c', dir, "Switch to a different working directory before executing the given command. Ex: tfswitch --chdir terraform_project will run tfswitch in the terraform_project directory")
 	versionFlag := getopt.BoolLong("version", 'v', "Displays the version of tfswitch")
@@ -118,6 +120,9 @@ func main() {
 		/* latest stable version */
 		case *latestFlag:
 			installLatestVersion(custBinPath, mirrorURL)
+		/* latest version required by module */
+		case *latestRequiredFlag:
+			installTFProvidedModule(*chDirPath, custBinPath, mirrorURL)
 		/* version provided on command line as arg */
 		case len(args) == 1:
 			installVersion(args[0], &binPath, mirrorURL)
@@ -181,6 +186,14 @@ func main() {
 	/* show latest stable version */
 	case *showLatestFlag:
 		showLatestVersion(custBinPath, mirrorURL)
+
+	/* latest required by module version */
+	case *latestRequiredFlag:
+		installTFProvidedModule(*chDirPath, custBinPath, mirrorURL)
+
+	/* show latest required by module version */
+	case *showLatestRequiredFlag:
+		showLatestRequiredVersion(*chDirPath, custBinPath, mirrorURL)
 
 	/* version provided on command line as arg */
 	case len(args) == 1:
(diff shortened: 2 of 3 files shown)


yermulnik:
Somewhat alike use case that I created issue two years ago for: https://github.com/warrensbox/terraform-switcher/issues/154

lib/semver.go
```diff
@@ -12,7 +13,7 @@ func GetSemver(tfconstraint *string, mirrorURL *string) (string, error) {
 
 	listAll := true
 	tflist, _ := GetTFList(*mirrorURL, listAll) //get list of versions
-	fmt.Printf("Reading required version from constraint: %s\n", *tfconstraint)
+	fmt.Fprintf(os.Stderr, "Reading required version from constraint: %s\n", *tfconstraint)
```
yermulnik, line 16: I think change from use of `fmt.Printf()` to `fmt.Fprintf(os.Stderr, …` should not be applied as an ad-hoc and partially, but the whole `tfswitch` codebase should be revisited to see where `fmt.Fprintf(os.Stderr, …` is going to be more appropriate.
Instead of switching to `fmt.Fprintf(os.Stderr, …` for this particular use case, I'd suggest to have a cmdline flag created to provide verbose or terse (and even JSON) output, so that the new feature from this PR can be used programmatically w/o it printing human-friendly messages.

main.go
```diff
@@ -60,6 +60,8 @@ func main() {
 	showLatestStable := getopt.StringLong("show-latest-stable", 'S', defaultLatest, "Show latest implicit version. Ex: tfswitch --show-latest-stable 0.13 prints 0.13.7 (latest)")
 	latestFlag := getopt.BoolLong("latest", 'u', "Get latest stable version")
 	showLatestFlag := getopt.BoolLong("show-latest", 'U', "Show latest stable version")
+	latestRequiredFlag := getopt.BoolLong("latest-required", 'r', "Get latest required by module version")
```
yermulnik, line 63: What's this argument use case? Ain't this the default behavior of `tfswitch` when no cmdline args are provided?

main.go
```diff
@@ -182,6 +187,14 @@ func main() {
 	case *showLatestFlag:
 		showLatestVersion(custBinPath, mirrorURL)
 
+	/* latest required by module version */
```
yermulnik, line 190: Maybe a bit more obviousness might make sense (despite I;m not sure this option is what we need in general):
```suggestion
	/* install latest required by module version */
```

main.go
```diff
@@ -428,15 +454,18 @@ func installOption(listAll bool, custBinPath, mirrorURL *string) {
 
 // install when tf file is provided
 func installTFProvidedModule(dir string, custBinPath, mirrorURL *string) {
-	fmt.Printf("Reading required version from terraform file\n")
+	fmt.Println("Reading required version from terraform file")
 	module, _ := tfconfig.LoadModule(dir)
-	tfconstraint := module.RequiredCore[0] //we skip duplicated definitions and use only first one
+	if len(module.RequiredCore) >= 1 {
```
yermulnik, line 459: Wrong conditional RHS copied from another function?
```suggestion
	if len(module.RequiredCore) == 0 {
```

main.go
```diff
@@ -428,15 +454,18 @@ func installOption(listAll bool, custBinPath, mirrorURL *string) {
...
+		fmt.Println("The provided directory does not have required terraform version constraint.")
+		os.Exit(1)
+	}
+	tfconstraint := strings.Join(module.RequiredCore, ", ")
```
yermulnik, line 463: This is the change into existing logic which may change UX.

Since I'm not aware of why @warrensbox initially decided to pick only first TF version constraint and ignore all consequent definitions, we might want to implement this the way Terraform does this. Is `strings.Join(module.RequiredCore, ", ")` what TF does? Could you please point me to TF code where this is done. Thanks.

yermulnik (commented):
1. This feature might probably need to be reflected in README
2. New cmdline arg should be covered with test(s) if possible please
3. This works oddly on my end:
```shell
> git branch
  master
* pr/301

> git diff
diff --git a/main.go b/main.go
index 614e572..286be1c 100644
--- a/main.go
+++ b/main.go
@@ -455,8 +455,9 @@ func installOption(listAll bool, custBinPath, mirrorURL *string) {
 // install when tf file is provided
 func installTFProvidedModule(dir string, custBinPath, mirrorURL *string) {
        fmt.Println("Reading required version from terraform file")
+       fmt.Printf("%s\n", dir)
        module, _ := tfconfig.LoadModule(dir)
-       if len(module.RequiredCore) >= 1 {
+       if len(module.RequiredCore) == 0 {
                fmt.Println("The provided directory does not have required terraform version constraint.")
                os.Exit(1)
        }

> go run main.go --chdir /some/path/ --show-latest-required
Reading configuration from home directory for .tfswitch.toml
Reading required version from terraform file /some/path/
Reading required version from constraint: ~> 1.1.0
Matched version: 1.1.9
Switched terraform to version "1.1.9"

> terraform version | head -1
Terraform v1.1.9

> go run main.go --chdir /some/path/ --show-latest-required
Reading configuration from home directory for .tfswitch.toml
Reading required version from terraform file /some/path/
Reading required version from constraint: ~> 1.1.0
Matched version: 1.1.9
Switched terraform to version "1.1.9"
```
I would not expect this new feature to switch TF version for me.

main.go
ikorchynskyi: It is added in the same style as other comments above, e.g.
```
	/* latest stable version */
```

main.go
ikorchynskyi: Yes, should be "== 0" for sure.

main.go
ikorchynskyi: I'd say it is a way to check all of them. An example of doing that one by one:
https://github.com/hashicorp/terraform-schema/blob/main/earlydecoder/decoder.go#L30-L42
```
	var coreRequirements version.Constraints
	for _, rc := range mod.RequiredCore {
		c, err := version.NewConstraint(rc)
		if err != nil {
			diags = append(diags, &hcl.Diagnostic{
				Severity: hcl.DiagError,
				Summary:  "Unable to parse terraform requirements",
				Detail:   fmt.Sprintf("Constraint %q is not a valid constraint: %s", rc, err),
			})
			continue
		}
		coreRequirements = append(coreRequirements, c...)
	}
```

lib/semver.go
ikorchynskyi: The reason behind that changes for me is that without output separation the tool is _unusable_ in automation (I am not taking into account the "parse output" approach for obvious reasons).

ikorchynskyi:
> 1. This feature might probably need to be reflected in README
> 2. New cmdline arg should be covered with test(s) if possible please
> 3. This works oddly on my end:

But seems any "--show-*" flag works in a same way when there is a .toml file provided
```
	case fileExists(TOMLConfigFile) || fileExists(HomeTOMLConfigFile):
```
in that case there is no --show flag check at all.

main.go
yermulnik: Aha, got it. So it's sort of an enforcement/explicitness of what does `case checkTFModuleFileExist(*chDirPath) && len(args) == 0:` conditional. Makes sense.
Should probably use `installTFProvidedModule(*chDirPath, &binPath, mirrorURL)` then (prefix `binPath` with `&`) at line 125 in `main.go` in this PR.

main.go
yermulnik: Got it. Thanks.

main.go
yermulnik: What I mean is changing "usual" way of "detecting" required TF version just like that may cause unpredictable and unexpected behavior on user end. So I'd suggest to split this off into a dedicated PR **and** discuss whether we want to go with `strings.Join()` like you implemented this in this PR, or e.g. like this is implemented in the example above that you provided so that we don't introduce custom way of determining required TF version.

@warrensbox @jukie FYI

lib/semver.go
yermulnik: My point is that such a change should be applied for all code instead of just a bit of it.
As for your reasoning (it all makes sense!) this is why I suggested to add a command line flag to "turn on" or "turn off" verbosity ("verbosity" as in printing to stdout).

@warrensbox @jukie What do you think?

yermulnik:
> > 3. This works oddly on my end:
> 
> But seems any "--show-*" flag works in a same way when there is a .toml file provided
> 
> ```
> 	case fileExists(TOMLConfigFile) || fileExists(HomeTOMLConfigFile):
> ```
> 
> in that case there is no --show flag check at all.

Oh, good catch. I missed this behavior =( Could you probably file an issue for this to get fixed (so that `--show*` flags don't actually switch TF version)? Thanks.

@warrensbox @jukie Any clue on whether this was implemented this way on purpose or this is indeed a bug?

main.go
ikorchynskyi: > prefix binPath with &

Well, `custBinPath` is already a pointer to string, and `binPath` is just the dereferenced value, so they are the same:
```
	case fileExists(TOMLConfigFile) || fileExists(HomeTOMLConfigFile):
		version := ""
		binPath := *custBinPath
```

yermulnik: My point is about consistency across the same codebits.

main.go
ikorchynskyi: I'd say that taking into account all available constraints is a proper way to do that, but I've reverted that change - it is a kind of an edge case and may be discussed separately

ikorchynskyi:
Created #302 to clarify the flags logic

lib/semver.go
yermulnik: > I may prepare the appropriate change if needed

That would be great if you feel like doing that. Contributions are very welcome.
Would be great if you got `zerolog`'ing implemented with pretty logging turned on by default (since the most audience for `tfswitch` are humans) with an option to turn it off (as in use default of JSON output format) for convenience in CI/CD.
Thank you.

main.go
yermulnik: Aha, I got your point. We're talking about different "consistencies" =) I'm basing mine off of the use of `installTFProvidedModule()` for `checkTFModuleFileExist()` case within the same `case fileExists(TOMLConfigFile) || fileExists(HomeTOMLConfigFile)` block.

Though... it looks like `binPath` vs `custBinPath` logic is messed up and is a bit convoluted within the `case fileExists(TOMLConfigFile) || fileExists(HomeTOMLConfigFile):` conditional. Especially given the comment that `custBinPath` should override TOML config value.
So seemingly the logic intended should've been something like `if custBinPath == defaultBin; then read TOML config; else binPath = custBinPath`, but it is vice versa and hence some nested conditional cases use `custBinPath`, while others use `bibPath` 😲 Or I'm missing something essential… 🤔

@ikorchynskyi @jukie @warrensbox What do you guys think of last paragraph above ⬆️ re the intended logic — should we rectify this or it's all fine and @ikorchynskyi should keep using `custBinPath` for `latestRequiredFlag` case (see blow)? https://github.com/warrensbox/terraform-switcher/blob/cb690d56cee6c724a4af38fc222b02c14afb9d46/main.go#L123-L125

lib/semver.go
ikorchynskyi: @yermulnik I will wait till the @warrensbox response, as I do not have a clear vision in terms of a direction the tool suppose to move :)
My interest here is to make the CI/CD ready solution which may help with the following flow:

1. We have a terraform project
2. We do _something_ to get the `terraform` version that would be able to make `init`, `plan` and `apply` with as little configuration and duplication as possible (having `required_version` and some specific version in `.terraform-version` is still treated a duplication)
3. We use that exact version to pull the specific tag of the official Docker image with `terraform` to execute the p. 2

As of now - I've come with a https://github.com/ikorchynskyi/terraform-version-inspect as a solution - the part of version constraint functionality from `tfswitch` and some stricter parsing using `json` release version "API" w/o relying on regex parsing

yermulnik: > so it is likely the plan for `v0.2` :)

Sounds great. This is how OSS comes into the world =)
Pre-built binaries and Homebrew tap would also be kewl to have 😺

ikorchynskyi: Yes, will add.
I will leave this PR for further discussion with maintainers, but as of now - I've updated `terraform-version-inspect` to use releases API, and it seems that it serves our needs quite well - already integrated in CI/CD pipelines and works as expected

MatthewJohn:
I know this is quite old, but we have a recent issue that @yermulnik thinks it will fix. 
Is it worth updating to align with main? Is there still interest in this?

yermulnik:
Closing in favour of #536

Issue #311, opened 2023-07-01
Issue when configuring Artifactory mirror: cannot get list from mirror


fcatacut:
My company has created a mirror of https://releases.hashicorp.com/terraform in Artifactory and intends to block direct downloads from the HashiCorp site.  I honestly don't know if my issue is caused by how we've configured Artifactory or if it's with `tfswitch` itself.  I lean towards the former since I don't see any issues reporting this error.

For some reason, running `tfswitch --mirror https://repo.<redacted>.com/artifactory/hashicorp-releases --list-all` results in a "cannot get list from mirror" error.

```bash
$ tfswitch --mirror https://repo.<redacted>.com/artifactory/hashicorp-releases --list-all
Cannot get list from mirror: https://repo.<redacted>.com/artifactory/hashicorp-releases
Use the arrow keys to navigate: ↓ ↑ → ←
? Select Terraform version:
▸ 1.3.4 *recent
  1.5.1 *recent
  1.5.0 *recent
```

However, I'm not running into this issue when using `tfenv`.

```bash
$ TFENV_REVERSE_REMOTE=1 tfenv list-remote
1.5.1
1.5.0
1.4.6
1.3.4
1.2.1
```

I'd like to continue using `tfswitch`.  How should I go about troubleshooting this issue?

tusv:
I think the issue is with how regex is being used to find versions.
On official terraform mirror a sample line in body looks like
`
<a href="[/terraform/1.5.1/](https://releases.hashicorp.com/terraform/1.5.1/)">terraform_1.5.1</a>
`
Whereas, in artifactory it's may be like this
`
<a href="[1.5.1/](https://repo1.uhc.com/artifactory/hashicorp-releases-cache/terraform/1.5.1/)">1.5.1/</a>    23-Jun-2023 02:03    -
`
regex being used is `\/(\d+\.\d+\.\d+)\/?"` which works for the former because of leading backslash and does not for the latter since there is no leading backslash. https://github.com/warrensbox/terraform-switcher/blob/master/lib/list_versions.go#L34

<img width="548" alt="image" src="https://github.com/warrensbox/terraform-switcher/assets/15346689/0e1865c9-eea6-49b7-b8a6-0e1ac28cd510">

Wonder if we can solve it by making leading backslash optional in regex. i.e. `\/?(\d+\.\d+\.\d+)\/?"`
[...]

PR #313, opened 2023-07-12
Make leading slash optional in regex when looking for versions

tusv:
Trying to solve https://github.com/warrensbox/terraform-switcher/issues/311

### DIFF
--- a/lib/list_versions.go
+++ b/lib/list_versions.go
@@ -27,11 +27,11 @@ func GetTFList(mirrorURL string, preRelease bool) ([]string, error) {
 	var semver string
 	if preRelease == true {
 		// Getting versions from body; should return match /X.X.X-@/ where X is a number,@ is a word character between a-z or A-Z
-		semver = `\/(\d+\.\d+\.\d+)(-[a-zA-z]+\d*)?/?"`
+		semver = `\/?(\d+\.\d+\.\d+)(-[a-zA-z]+\d*)?/?"`
 	} else if preRelease == false {
 		// Getting versions from body; should return match /X.X.X/ where X is a number
 		// without the ending '"' pre-release folders would be tried and break.
-		semver = `\/(\d+\.\d+\.\d+)\/?"`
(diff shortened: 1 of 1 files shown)


yermulnik (commented):
Might probably make sense to apply custom regex for when custom mirror endpoint is passed by means of cmdline flag. Ideally identify whether endpoint is Artifactory and use custom regex to only apply in this case.
Does it sounds sensible?

tusv:
@yermulnik I could do that. But, in this case since the fix does not break existing behavior I am hoping we can keep it simple.

I am also guessing that this may not be artifactory specific and any other repository/mirror that relies on auto index browsing is likely to behave this way.

yermulnik:
@tusv Then first of the suggested options fits this use case. My only intention is to keep as much compatibility as possible, hence applying custom regex for when `mirror` cmdline option is specified sounds reasonable to me.

Hope @warrensbox and @jukie would have an opportunity to weigh too though.

MatrixCrawler:
This looks good to me @yermulnik 
The ? quantifier in the regex requires zero or one / so i don't see that this would break existing behvaiour.

MatrixCrawler (approved):

yermulnik (approved):

yermulnik:
> This looks good to me @yermulnik The ? quantifier in the regex requires zero or one / so i don't see that this would break existing behvaiour.

👍🏻 Agreed.
Are we okay to merge this one?

yermulnik:
Asking to double check as I'm a bit out of context with the old PRs =(

tusv:
> > This looks good to me @yermulnik The ? quantifier in the regex requires zero or one / so i don't see that this would break existing behvaiour.
> 
> 👍🏻 Agreed. Are we okay to merge this one?

Yes, I just double checked just to be sure.

yermulnik:
> > 👍🏻 Agreed. Are we okay to merge this one?
> 
> Yes, I just double checked just to be sure.

Apologies for the confusion: I intended to ask @MatrixCrawler

Issue #117, opened 2020-12-07
snap package cannot use "~/.terraform.version" and "/usr/local/bin"


strowk:
I've installed package from snap and running `tfswitch -b $HOME/bin` , it gives me 
```
Error while creating /home/user/.terraform.versions/terraform_0.13.5_linux_amd64.zip - open /home/user/.terraform.versions/terraform_0.13.5_linux_amd64.zip: permission denied
```
Running just `tfswitch --latest` complains about `/usr/local/bin` directory . Both places are not available from usual snap packages.

warrensbox:
1. Are you running `tfswitch -b $HOME/bin/terraform` or `tfswitch -b $HOME/bin` ?
2. Do you have permission to write to the `$HOME/bin` directory (what do you see when you run `ls -l` on $HOME?

strowk:
1. both give same result
1. I am owner  of $HOME/bin:
```
whoami
tim

ls -la $HOME | grep ' bin'
drwxrwxr-x  3 tim  tim   4096 Dez 14 16:50 bin

```

Note:

I've uninstalled tfswitch installed using Troubleshooting guilde and installed snap package again to reproduce it again

strowk:
From what I see here 
https://forum.snapcraft.io/t/access-to-specific-hidden-file-path-in-users-home/6948
it looks like 
1. problem is in ~/.terraform.versions 
2. snap does not allow to use hidden dirs in home

So either tfswitch in snap package should automatically use different folder (maybe ~/snap/tfswitch ? I see this folder has been created by snap) or at least allow to customize it.

warrensbox:
@strowk  perhaps I can allow users to customize it.
Thank you for bringing this up. I will find some time to do this before the end of the year. Hopefully, I can release this fix in January. 
In the source code: -
```
getInstallLocation() //only downloads the terraform bins into the home directory in .terraform.versions
.
.
.
installLocation = usr.HomeDir + installPath //where installPath = "/.terraform.versions/"
```
I can make this `getInstallLocation()` customizable where users can provide a location for all terraform binaries. Again, I can only get to this by the end of the year. If you know `go` and would like to contribute, I would be happy to review and merge your code as well.

I will let you know once I have implemented this. Thank you for bringing this up.

strowk:
[...]

PR #314, opened 2023-08-25
Use go-homedir.Dir() everywhere

kim0:
This can also close https://github.com/warrensbox/terraform-switcher/issues/117 since the binary will now respect $HOME which can be set to any location instead of ~/.terraform.versions


yermulnik:
Hey @kim0 
Thank you for the contribution. I have two questions:
1. I'm not quite really familiar with Golang internals and hence why is so huge removal change set for `go.sum` file?
2. Looking at description and code of `os.UserHomeDir()`, I can see it relies upon `HOME` env var exclusively and returns error if it is not defined, which is quite rare though still viable condition (when `HOME` env var (or non-Linux OS alternative) is not defined), whilst `go-homedir` does its best to dig to the bottom of it and (despite running external commands) tries to determine user home dir much better than `os.UserHomeDir()` — what's the rationale behind this PR given the comparison on how these two determine user home dir? Thanks (and sorry if the question is dumb — I'm not a Golang folk 🤷🏻)

ps: this PR may also become stale as both owner (@warrensbox) and another contributor (@jukie) are more of not with this project rather than with it 😢

kim0:
Hello @yermulnik 

Thanks a lot for your feedback. I'm not that great with Golang either, I didn't think `os.UserHomeDir()` only looks at the `HOME` env var, but looking at its code, that does seem to be the case! A bit surprising since that is supposed to be the recommended way to work. Anyway, for now, I've switched everything to use `go-homedir` which does seem to do a better job!

It's sorta sad that the maintainers are no longer active mostly. My use-case involves running tfswitch with a read-only home directory on AWS lambda, which causes errors since tfswitch insists on writing to the home directory and does not respect `$HOME`. I hope I don't end up forking this repo and building releases! Really hope the maintainers can get in touch soon. Thanks!

main.go
```diff
@@ -70,7 +71,11 @@ func main() {
 	getopt.Parse()
 	args := getopt.Args()
 
-	homedir := lib.GetHomeDirectory()
```
yermulnik, line 73: It's probably worth of 1) either switching to `lib.GetHomeDirectory()` for all `homedir.Dir()`+`log.Fatal()` occurrences (there are only a few use cases in current tests where `os.Exist(1)` is not used for failures getting user's home dir) 2) or removing `lib.GetHomeDirectory()` at all as not used anymore.
On one hand the former seems to make sense to unify use of the same code snippet across code base, though on the other hand some say it is best practice to avoid use of `log.Fatal` and, when used, should better be used directly from `main.go` only (as opposite to using it from non-main).

What do you think?

yermulnik:
> My use-case involves running tfswitch with a read-only home directory on AWS lambda, which causes errors since tfswitch insists on writing to the home directory and does not respect `$HOME`.

Would the https://github.com/warrensbox/terraform-switcher#use-tfswitchtoml-file--for-non-admin---users-with-limited-privilege-on-their-computers be an option for your use case? You also can provide binary path as command line arg.

> I hope I don't end up forking this repo and building releases! Really hope the maintainers can get in touch soon. Thanks!

🤞🏻

kim0:
> be an option for your use case?

Thanks a lot for the reply @yermulnik .. wrt the suggested `tfswitch -b` or conf file, unfortunately this does not help. This is bec `tfswitch` is very stubborn wanting to write to the home directory. This was the reason for this PR. I confirmed `-b` doesn't help, by running in a container with the home directory `/root` mounted as read-only. You can see it fail below

```bash
root@5bc5fcb6879a:~# pwd
/root
root@5bc5fcb6879a:~# touch foo
touch: cannot touch 'foo': Read-only file system
root@5bc5fcb6879a:~# tfswitch -u -b /tmp/terraform
Creating directory for terraform binary at: /root/.terraform.versions
Unable to create directory for terraform binary at: /root/.terraform.versionspanic: mkdir /root/.terraform.versions: read-only file system

goroutine 1 [running]:
github.com/warrensbox/terraform-switcher/lib.CreateDirIfNotExist({0xc000327300, 0x19})
        /home/circleci/go/src/github.com/warrensbox/terraform-switcher/lib/files.go:112 +0x147
github.com/warrensbox/terraform-switcher/lib.GetInstallLocation()
        /home/circleci/go/src/github.com/warrensbox/terraform-switcher/lib/install.go:72 +0xca
github.com/warrensbox/terraform-switcher/lib.Install({0xc000300651, 0x5}, {0x7ffdda050f10?, 0xc0000920f0?}, {0x9e004d, 0x28})
        /home/circleci/go/src/github.com/warrensbox/terraform-switcher/lib/install.go:94 +0x70
main.installLatestVersion(0xc00016acb0, 0xc00016ad90)
        /home/circleci/go/src/github.com/warrensbox/terraform-switcher/main.go:237 +0x50
main.main()
        /home/circleci/go/src/github.com/warrensbox/terraform-switcher/main.go:179 +0xc9f
root@5bc5fcb6879a:~# echo $?
2
root@5bc5fcb6879a:~# ls -l /tmp/terraform
ls: cannot access '/tmp/terraform': No such file or directory
```

So given this failure, are you interested in merging this PR please? Let me know, thanks!

main.go
kim0: I would prefer to delete the `lib.GetHomeDirectory()` function, since it does not add much over the base library. I'll push that update now. Hope with this update, we can get this merged. Thanks!

kim0:
Sorry @yermulnik I confused you with being a maintainer .. Guess I'll have to wait more or find some other solution

yermulnik (approved):

MatrixCrawler (approved):

PR #320, opened 2023-12-13
Add -e flag to force semver lookup at releases.hashicorp.com

adam-arnold:
When using artifactory as a mirror, semvers are unable to be looked up when using a version constraint due to the way that artifactory will cache a terraform file version after its been downloaded.  The semver lookup needs to have a range of versions present already. Adds new -e flag to force this lookup at releases.hashicorp.com/terraform when -m (mirrorURL) flag is specified

Workaround for https://github.com/warrensbox/terraform-switcher/issues/311

Some autoformatting of the comments that can be removed if desired


adam-arnold:
@warrensbox, could i get a review on this?

MatrixCrawler:
Closed this as the issue for that was closed.
If this problem still persists, please create a new issue report.

Thanks

Issue #290, opened 2023-02-21
Does tfswitch verifies the sha256sums?


vikas027:
When `tfswitch` downloads a `terraform` binary, does it verifies the `sha256sum` advertised on hashicorp release site (https://releases.hashicorp.com/terraform)?
If yes, it will be good to advertise the same in the `README.md` :)

apogrebnyak:
As far as I can tell it does not.

The download is done by this method -> https://github.com/warrensbox/terraform-switcher/blob/d7dfd1b44605b095937e94b981d24305b858ff8c/lib/download.go#L13

It is called from here and I don't see any signature verification until the method exits -> https://github.com/warrensbox/terraform-switcher/blob/master/lib/install.go#L136

I am not sure if `sha256` check will do anything to ensure clean source.  Only signature check against trusted key would.
It is not done here either.

MatrixCrawler:
Duplicates #160

Issue #274, opened 2022-10-05
mirror keyword does nothing in .tfswitch.toml


amavlyanov:
I was trying to lock tf mirror instead of writing it each time I need to switch terraform version, but the '`mirror`' keyword seem to not workin in `$HOME/.tfswitch.toml` (I was trying the following config):

```toml
bin = "$HOME/.local/bin/terraform"
mirror = "https://releases.comcloud.xyz/terraform/"
```

Can you please provide us with a clear documentation of what options available via .toml config, and please, add a '`mirror`' keyword there.

warrensbox:
@amavlyanov 
I don't think there's an option in the toml file to add that.
At this time, you need to manually pass from from the command line. For example:
`tfswitch --mirror https://example.jfrog.io/artifactory/hashicorp`

However, I can add this feature (on the toml file) to the next release of tfswitch.

PR #334, opened 2024-03-28
Checksum check for TF Binaries

### LINKED ISSUE #160: Verify Downloads Checksum & Signature
Due to the increasing frequency of supply chain attacks, it would be nice if tfswitch verified the checksums and gpg signatures of terraform when it downloads it for the first time, as described here:  
https://www.hashicorp.com/security#template-page-security:~:text=Release%20Archive%20Checksum%20Verification

Here's some example bash code that may provide a useful starting point for doing the equivalent in Go:
https://gist.github.com/markmsmith/cda59d5f24a812bea66fb3dbd7612397

MatrixCrawler:
This fixes #160 and #290 

- implement checksum check for terraform binaries
- implement signature check on terraform binaries
- added .gitignore for test with windows systems
- added .github action for testing with multiple go scenarios


lib/download.go
```diff
@@ -1,47 +1,142 @@
...
 	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
 )
 
-// DownloadFromURL : Downloads the binary from the source url
-func DownloadFromURL(installLocation string, url string) (string, error) {
+const (
+	pubKeyId     = "72D7468F"
```
yermulnik, line 15: Would it make sense to allow users to customize the value?
E.g. if or when Hashicorp get the key rotated so that users don't depend on the new release of `tfswitch`?

lib/download.go
```diff
@@ -1,47 +1,142 @@
...
+	pubKeyPrefix = "hashicorp_"
+	pubKeySuffix = ".asc"
+	pubKeyUri    = "https://www.hashicorp.com/.well-known/pgp-key.txt"
```
yermulnik, line 18: Same point as for `pubKeyId` above: would it make sense to make it adjustable by users?
Additionally users may want to have their own mirror (or e.g. buildfarms w/o access to external internet) and even re-packed (built by themselves) binaries.

So it looks like it makes sense to allow verification to be 1) customizable and 2) switchable. What do you think?

lib/download.go
```diff
@@ -1,47 +1,142 @@
...
+// DownloadFromURL : Downloads the terraform binary and its hash from the source url
+func DownloadFromURL(installLocation string, mirrorURL string, tfversion string, versionPrefix string, goos string, goarch string) (string, error) {
+	pubKeyFilename := filepath.Join(installLocation, "/", pubKeyPrefix+pubKeyId+pubKeySuffix)
+	zipUrl := mirrorURL + tfversion + "/" + versionPrefix + tfversion + "_" + goos + "_" + goarch + ".zip"
+	hashUrl := mirrorURL + tfversion + "/" + versionPrefix + tfversion + "_SHA256SUMS"
+	hashSignatureUrl := mirrorURL + tfversion + "/" + versionPrefix + tfversion + "_SHA256SUMS." + pubKeyId + ".sig"
+
+	err := downloadPublicKey(installLocation, pubKeyFilename)
+	if err != nil {
+		log.Fatal("[Error]: Could not download public key file")
```
yermulnik, line 30: Same `log.Fatal()` vs `fmt.Println()` question: why different approaches?

lib/download.go
```diff
@@ -1,47 +1,142 @@
...
+
+	publicKeyFile, err := os.Open(pubKeyFilename)
+	if err != nil {
+		log.Fatal("[Error]: Could not open the public key")
+		return "", err
+	}
+
+	signatureFile, err := os.Open(hashSigFilePath)
+	if err != nil {
+		log.Fatal("[Error]: Could not open the public key")
```
yermulnik, line 63:
```suggestion
		log.Fatal("[Error]: Could not open the signature file")
```

lib/download.go
```diff
@@ -1,47 +1,142 @@
...
+		return "", err
+	}
+
+	targetFile, err := os.Open(zipFilePath)
+	if err != nil {
+		log.Fatal("[Error]: Could not open the terraform binary for signature verification.")
```
yermulnik, line 69: It probs makes sense to remove trailing dot for consistency with the rest of messages
```suggestion
		log.Fatal("[Error]: Could not open the Terraform binary archive for signature verification")
```

test-data/checksum-check-file
```diff
@@ -0,0 +1,7 @@
...
+
+Ut wisi enim ad minim veniam, quis nostrud exerci tation ullamcorper suscipit lobortis nisl ut aliquip ex ea commodo consequat. Duis autem vel eum iriure dolor in hendrerit in vulputate velit esse molestie consequat, vel illum dolore eu feugiat nulla facilisis at vero eros et accumsan et iusto odio dignissim qui blandit praesent luptatum zzril delenit augue duis dolore te feugait nulla facilisi.
+
+Nam liber tempor cum soluta nobis eleifend option congue nihil imperdiet doming id quod mazim placerat facer possim assum. Lorem ipsum dolor sit amet, consectetuer adipiscing elit, sed diam nonummy nibh euismod tincidunt ut laoreet dolore magna aliquam erat volutpat. Ut wisi enim ad minim veniam, quis nostrud exerci tation ullamcorper suscipit lobortis nisl ut aliquip ex ea commodo consequat.
```
yermulnik, line 7: It's best to add proper EOF (linefeed) please
image

lib/checksum.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	return true
+}
+
+// checkSignatureOfChecksums THis will verify the signature of the file containing the hash sums
+func checkSignatureOfChecksums(keyRingReader *os.File, hashFile *os.File, signatureFile *os.File) bool {
+	log.Println("Verifying signature of checksum file...")
+	keyring, err := openpgp.ReadArmoredKeyRing(keyRingReader)
+	if err != nil {
+		log.Fatal("[Error]: Read armored key ring: " + err.Error())
```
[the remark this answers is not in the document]
MatrixCrawler: I was actually wondering mysql why we don't use the log class overall.
log.Fatal will automatically end with an sys.exit(1) and stops the execution. At this point it makes no sense to continue with execution.
I tried to consolidate all fmt.Println to log.Println or log.Fatal

lib/download.go
MatrixCrawler: I think in this case we should cover that with a new release.

lib/checksum.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	return "", nil
+}
+
+// checkChecksumMatches This will calculate and compare the check sum of the downloaded zip file.
+func checkChecksumMatches(hashFile string, targetFile *os.File) bool {
+	_, fileName := filepath.Split(targetFile.Name())
+	expectedChecksum, err := getChecksumFromHashFile(hashFile, fileName)
+	if err != nil {
+		fmt.Println("[Error]: could not get expected checksum from file: " + err.Error())
```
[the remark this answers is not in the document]
yermulnik: > but maybe we should talk about how to print the output in general.

💯 
I personally would prefer `[ERROR] Brief sum up: err text`
Same for info, warnings, etc: `[INFO] ...`, `[WARN] ...`, so on

lib/checksum.go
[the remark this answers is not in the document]
yermulnik: I might be mixing up, though I seem to think it was @warrensbox's preference.
I'm all for `log.Fatal()` for fatal errors and `log.Println()` for the rest.

lib/download.go
yermulnik: Agree. My point is that we're better off not introducing features that may block users as this had happened e.g. when Hashi moved to another site to distribute releases — that block all of the `tfswitch` users.

lib/utils.go
```diff
@@ -0,0 +1,12 @@
+package lib
```
yermulnik, line 1: Would it make sense to name it according to the file name?
```suggestion
package utils
```

lib/checksum.go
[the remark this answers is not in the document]
MatrixCrawler: We could have a look into https://github.com/op/go-logging for a loglevel based logging

lib/utils.go
MatrixCrawler: Then we would move this file into a new package. i don't think this would be necessary at this moment.
You might be mixing something up here

lib/checksum.go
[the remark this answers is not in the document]
yermulnik: Yeah, that looks neat.
Would you be up to look into this when you get a chance? 😺

lib/utils.go
yermulnik: Okay, I'm not very familiar with this bits in Go. Just thought `utils` would better match helper functions.

lib/checksum.go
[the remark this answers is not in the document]
MatrixCrawler: I will but I will create a new issue for that because this will be a lot of refactoring.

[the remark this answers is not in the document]
yermulnik: Yep-yep, that's my point. Appreciate this.

lib/download.go
MatrixCrawler: I just had a look into the possible configurations and i see a lot of consolidation potential there.
For example: the toml file only supports bin and version param (see #274), and i don't even see a test case for that (the test-data/test_tfswitchtoml/.tfswitch.toml) is unreferenced throughout the project.
i just started fiddling around with refactoring the parameter parsing but this is definitely out of scope of this ticket.

yermulnik: Let me re-iterate and I'll leave this to you guys to decide then =)
https://github.com/warrensbox/terraform-switcher/pull/334#discussion_r1543158173 — we already were through the issue where `tfswitch` came up short for all of the users trying to download `terraform` because of the tiny change on the Hashicorp site end and users were absolutely blocked because they had no option to override the default. This is solely point that I frankly would not want to be introduced again.
While making this option customizable by users **looks to be** a scope of a separate ticket in general, I reckon the addition of a new functionality (which is the scope of this current PR) should better be accompanied with an option for users to override default values that configure this new functionality's behavior.
Thanks.

MatrixCrawler: @yermulnik i made the options for the public key configurable. Will you pleas have a look at the changes?

crablab: That's for the extra context @yermulnik - I vaguely remember that now. 

I think there's definitely a bigger thing, as @MatrixCrawler says, around configuration options that we could look at in a separate item.

crablab (approved):

lib/download.go
yermulnik: @MatrixCrawler Looking. Would you mind resolving merge conflict meanwhile?

.github/workflows/go.yml
yermulnik, line 1: This file seems redundant as of existence of https://github.com/warrensbox/terraform-switcher/blob/master/.github/workflows/build.yml, right? 🤔

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+### Added
+- added verification of the checksums for downloaded terraform files.
```
yermulnik, line 9: Do we have any automation around this? 🤔

lib/checksum.go
```diff
@@ -0,0 +1,73 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// getChecksumFromFile Extract the checksum from the signature file
+func getChecksumFromHashFile(signatureFilePath string, terraformFileName string) (string, error) {
+	readFile, err := os.Open(signatureFilePath)
+	if err != nil {
+		fmt.Println("[Error] : Could not open ", signatureFilePath)
```
yermulnik, line 20: I think it makes sense to sync with https://github.com/warrensbox/terraform-switcher/pull/350 so that we don't get a mess again when #350 is merged.
Would it make sense to complete #350 first and then get back to this PR?

yermulnik (changes requested):
Let's complete #350 first to avoid discrepancies with logging again. WDT?

lib/download.go
yermulnik: I'be partially reviewed and I think we should not slap this and #350 together and prefer #350 to get merged first.

CHANGELOG.md
crablab: I don't think so - probably not something for this PR, right?

yermulnik: Looks like yes. Let's remove this bit for now.
Ww'll need to investigate options on changelog automation later.

lib/checksum.go
```diff
@@ -0,0 +1,73 @@
...
+	return true
+}
+
+// checkSignatureOfChecksums THis will verify the signature of the file containing the hash sums
+func checkSignatureOfChecksums(keyRingReader *os.File, hashFile *os.File, signatureFile *os.File) bool {
+	log.Println("Verifying signature of checksum file...")
+	keyring, err := openpgp.ReadArmoredKeyRing(keyRingReader)
+	if err != nil {
+		log.Fatal("[Error] : Read armored key ring: " + err.Error())
+		return false
```
MatthewJohn, line 63: Is there any value in these return values, if log.Fatal immediately quits?

lib/download.go
```diff
@@ -1,47 +1,137 @@
...
 	defer response.Body.Close()
 
 	if response.StatusCode != 200 {
 		//Sometimes hashicorp terraform file names are not consistent
 		//For example 0.12.0-alpha4 naming convention in the release repo is not consistent
-		return "", fmt.Errorf("[Error] : Unable to download from %s", url)
+		log.Fatalf("[Error] : Unable to download from %s", url)
 	}
 
 	zipFile := filepath.Join(installLocation, fileName)
```
MatthewJohn, line 102: If this function is downloading arbitrary files now, zipFile probably isn't too accurate.

lib/download.go
```diff
@@ -1,47 +1,137 @@
...
+		log.Fatal("[Error]: Could not open the terraform binary for checksum verification.", zipFilePath)
+	}
+
+	hashFile, err := os.Open(hashFilePath)
+	if err != nil {
+		log.Fatal("[Error]: Could not open the terraform checksum file for signature verification.", hashFilePath)
+	}
+	verified := checkSignatureOfChecksums(publicKeyFile, hashFile, signatureFile)
+	if !verified {
+		return "", errors.New("signature of checksum files could not be verified")
```
MatthewJohn, line 75: I'm not sure how much it matters - but I'm quite aware that if we download the public key, hash file and the zip file and the verification fails - all of these files are left in the install directory. Is it worth tidying them up after a failed download.

MatthewJohn: In fact, the returned zip file path, once extracted by `Install()` will remove the zip file, but the checksum file will remain.

MatrixCrawler: I delete the files needed for the verification after the process now.

CHANGELOG.md
MatrixCrawler: Removed this entry

lib/checksum.go
MatrixCrawler: Yeah i will have to have a look into this once we merged the logger stuff from #350

lib/download.go
MatrixCrawler: Fixed

MatrixCrawler:
@yermulnik #350 is merged now and i rebased this. Will you have a look again please?

.github/workflows/go.yml
MatrixCrawler: Jep.. this is redundant. I will remove it.

lib/checksum.go
```diff
@@ -0,0 +1,86 @@
...
+	var fileHandlersToClose []*os.File
+	fileHandlersToClose = append(fileHandlersToClose, keyRingReader)
+	fileHandlersToClose = append(fileHandlersToClose, hashFile)
+	fileHandlersToClose = append(fileHandlersToClose, signatureFile)
+
+	logger.Info("Verifying signature of checksum file...")
+	keyring, err := openpgp.ReadArmoredKeyRing(keyRingReader)
+	if err != nil {
+		closeFileHandlers(fileHandlersToClose)
+		logger.Errorf("Read armored key ring: %q", err.Error())
```
yermulnik, line 73:
```suggestion
		logger.Errorf("Could not read armored key ring: %v", err)
```
1. Would it worth to provide some hint on how a user can try and resolve this?
2. Would it work on all supported OS (especially on Windows?)?

lib/checksum.go
```diff
@@ -0,0 +1,86 @@
...
+func checkSignatureOfChecksums(keyRingReader *os.File, hashFile *os.File, signatureFile *os.File) bool {
+	var fileHandlersToClose []*os.File
+	fileHandlersToClose = append(fileHandlersToClose, keyRingReader)
+	fileHandlersToClose = append(fileHandlersToClose, hashFile)
+	fileHandlersToClose = append(fileHandlersToClose, signatureFile)
+
+	logger.Info("Verifying signature of checksum file...")
+	keyring, err := openpgp.ReadArmoredKeyRing(keyRingReader)
+	if err != nil {
+		closeFileHandlers(fileHandlersToClose)
```
yermulnik, line 72: Would `defer closeFileHandlers(fileHandlersToClose)` work to replace repeatable `closeFileHandlers(fileHandlersToClose)` calls in this function?

lib/checksum.go
```diff
@@ -0,0 +1,86 @@
...
+		logger.Errorf("Read armored key ring: %q", err.Error())
+		return false
+	}
+
+	_, err = openpgp.CheckDetachedSignature(keyring, hashFile, signatureFile)
+	if err != nil {
+		closeFileHandlers(fileHandlersToClose)
+		logger.Errorf("Checking detached signature: %q", err.Error())
```
yermulnik, line 80:
```suggestion
		logger.Errorf("Could not check detached signature: %v", err)
```

lib/checksum.go
```diff
@@ -0,0 +1,86 @@
...
+		return false
+	}
+	logger.Info("Verification successful.")
```
yermulnik, line 83:
```suggestion
	logger.Info("Checksum file signature verification successful.")
```

lib/download.go
```diff
@@ -9,15 +9,100 @@ import (
 	"strings"
 )
 
-// DownloadFromURL : Downloads the binary from the source url
-func DownloadFromURL(installLocation string, url string) (string, error) {
+var (
+	PubKeyId     = "72D7468F"
+	PubKeyPrefix = "hashicorp_"
+	PubKeyUri    = "https://www.hashicorp.com/.well-known/pgp-key.txt"
+)
```
yermulnik, line 16: What do you think of consolidating default values at https://github.com/warrensbox/terraform-switcher/blob/master/lib/defaults.go or nearby the https://github.com/warrensbox/terraform-switcher/blob/master/main.go#L40-L51 ?

lib/download.go
```diff
@@ -9,15 +9,100 @@ import (
...
+// DownloadFromURL : Downloads the terraform binary and its hash from the source url
+func DownloadFromURL(installLocation string, mirrorURL string, tfversion string, versionPrefix string, goos string, goarch string) (string, error) {
+	pubKeyFilename := filepath.Join(installLocation, "/", PubKeyPrefix+PubKeyId+pubKeySuffix)
+	zipUrl := mirrorURL + tfversion + "/" + versionPrefix + tfversion + "_" + goos + "_" + goarch + ".zip"
+	hashUrl := mirrorURL + tfversion + "/" + versionPrefix + tfversion + "_SHA256SUMS"
+	hashSignatureUrl := mirrorURL + tfversion + "/" + versionPrefix + tfversion + "_SHA256SUMS." + PubKeyId + ".sig"
+
+	err := downloadPublicKey(installLocation, pubKeyFilename)
+	if err != nil {
+		logger.Error("Could not download public key file")
```
yermulnik, line 31:
```suggestion
		logger.Error("Could not download public PGP key file.")
```

lib/download.go
```diff
@@ -9,15 +9,100 @@ import (
...
+		return "", err
+	}
+
+	logger.Infof("Downloading %q", zipUrl)
+	zipFilePath, err := downloadFromURL(installLocation, zipUrl)
+	if err != nil {
+		logger.Error("Could not download zip file")
```
yermulnik, line 38:
```suggestion
		logger.Error("Could not download zip file.")
```

lib/download.go
```diff
@@ -9,15 +9,100 @@ import (
...
+		return "", err
+	}
+
+	logger.Infof("Downloading %q", hashUrl)
+	hashFilePath, err := downloadFromURL(installLocation, hashUrl)
+	if err != nil {
+		logger.Error("Could not download hash file")
```
yermulnik, line 45:
```suggestion
		logger.Error("Could not download hash file.")
```

lib/download.go
```diff
@@ -9,15 +9,100 @@ import (
...
+		return "", err
+	}
+
+	logger.Infof("Downloading %q", hashSignatureUrl)
+	hashSigFilePath, err := downloadFromURL(installLocation, hashSignatureUrl)
+	if err != nil {
+		logger.Error("Could not download hash signature file")
```
yermulnik, line 52:
```suggestion
		logger.Error("Could not download hash signature file.")
```

lib/download.go
```diff
@@ -9,15 +9,100 @@ import (
...
+		return "", err
+	}
+
+	publicKeyFile, err := os.Open(pubKeyFilename)
+	if err != nil {
+		logger.Error("Could not open the public key")
```
yermulnik, line 58:
```suggestion
		logger.Errorf("Could not open public key %q: %v", pubKeyFilename, err)
```

lib/download.go
```diff
@@ -9,15 +9,100 @@ import (
...
+		return "", err
+	}
+
+	signatureFile, err := os.Open(hashSigFilePath)
+	if err != nil {
+		logger.Error("Could not open the public key")
```
yermulnik, line 64:
```suggestion
		logger.Errorf("Could not open hash signature file %q: %v", hashSigFilePath, err)
```

lib/download.go
```diff
@@ -9,15 +9,100 @@ import (
...
+		return "", err
+	}
+
+	targetFile, err := os.Open(zipFilePath)
+	if err != nil {
+		logger.Error("Could not open the terraform binary for signature verification.")
```
yermulnik, line 70:
```suggestion
		logger.Errorf("Could not open zip file %q: %v", zipFilePath, err)
```

lib/download.go
```diff
@@ -9,15 +9,100 @@ import (
...
+		return "", err
+	}
+
+	hashFile, err := os.Open(hashFilePath)
+	if err != nil {
+		logger.Error("Could not open the terraform binary for signature verification.")
```
yermulnik, line 76:
```suggestion
		logger.Errord("Could not open hash file %q: %v", hashFilePath, err)
```

lib/download.go
```diff
@@ -9,15 +9,100 @@ import (
...
+	}
+
+	var filesToCleanup []string
+	filesToCleanup = append(filesToCleanup, hashFilePath)
+	filesToCleanup = append(filesToCleanup, hashSigFilePath)
+
+	verified := checkSignatureOfChecksums(publicKeyFile, hashFile, signatureFile)
+	if !verified {
+		cleanup(filesToCleanup)
+		return "", errors.New("signature of checksum files could not be verified")
```
yermulnik, line 87:
```suggestion
		return "", errors.New("Signature of checksum file could not be verified")
```

lib/download.go
```diff
@@ -9,15 +9,100 @@ import (
...
+	}
+	match := checkChecksumMatches(hashFilePath, targetFile)
+	if !match {
+		cleanup(filesToCleanup)
+		return "", errors.New("checksums did not match")
```
yermulnik, line 92:
```suggestion
		return "", errors.New("Checksums did not match")
```

lib/download.go
```diff
@@ -9,15 +9,100 @@ import (
...
+	}
+	cleanup(filesToCleanup)
+	return zipFilePath, err
+}
+
+func downloadFromURL(installLocation string, url string) (string, error) {
 	tokens := strings.Split(url, "/")
 	fileName := tokens[len(tokens)-1]
-	logger.Infof("Downloading to %q", installLocation)
+	logger.Infof("Downloading to: %s", filepath.Join(installLocation, "/", fileName))
```
yermulnik, line 101:
```suggestion
	logger.Infof("Downloading to %q", filepath.Join(installLocation, "/", fileName))
```

lib/download.go
```diff
@@ -9,15 +9,100 @@ import (
...
 
 	response, err := http.Get(url)
 	if err != nil {
-		logger.Errorf("Error downloading %s: %v", url, err)
+		logger.Errorf("Error while downloading %q - %v", url, err)
```
yermulnik, line 105:
```suggestion
		logger.Errorf("Error downloading %s: %v", url, err)
```

lib/download.go
```diff
@@ -28,10 +113,10 @@ func DownloadFromURL(installLocation string, url string) (string, error) {
 		return "", errors.New("Unable to download from " + url)
 	}
 
-	zipFile := filepath.Join(installLocation, fileName)
-	output, err := os.Create(zipFile)
+	filePath := filepath.Join(installLocation, fileName)
+	output, err := os.Create(filePath)
 	if err != nil {
-		logger.Errorf("Error while creating %q: %v", zipFile, err)
+		logger.Errorf("Error while creating %q: %v", filePath, err)
```
yermulnik, line 119:
```suggestion
		logger.Errorf("Error creating %q: %v", filePath, err)
```

lib/download.go
```diff
@@ -43,5 +128,34 @@ func DownloadFromURL(installLocation string, url string) (string, error) {
...
+}
+
+func downloadPublicKey(installLocation string, targetFileName string) error {
+	logger.Debugf("Looking up public key file at %q", targetFileName)
+	publicKeyFileExists := FileExists(targetFileName)
+	if !publicKeyFileExists {
+		// Public key does not exist. Let's grab it from hashicorp
+		pubKeyFile, errDl := downloadFromURL(installLocation, PubKeyUri)
+		if errDl != nil {
+			logger.Error("Error while fetching the public key file from ", PubKeyUri)
```
yermulnik, line 141:
```suggestion
			logger.Errorf("Error fetching public key file from %s", PubKeyUri)
```

lib/download.go
```diff
@@ -43,5 +128,34 @@ func DownloadFromURL(installLocation string, url string) (string, error) {
...
+			return errDl
+		}
+		errRename := os.Rename(pubKeyFile, targetFileName)
+		if errRename != nil {
+			logger.Error("Error while renaming the public key file from ", pubKeyFile, " to ", targetFileName)
```
yermulnik, line 146:
```suggestion
			logger.Errorf("Error renaming public key file from %q to %q", pubKeyFile, targetFileName)
```

lib/download.go
```diff
@@ -43,5 +128,34 @@ func DownloadFromURL(installLocation string, url string) (string, error) {
...
+	}
+	return nil
+}
+
+func cleanup(paths []string) {
+	for _, path := range paths {
+		logger.Infof("Deleting %q", path)
+		err := os.Remove(path)
+		if err != nil {
+			logger.Error("Error deleting %q - %q", path, err)
```
yermulnik, line 158:
```suggestion
			logger.Error("Error deleting %q: %v", path, err)
```

lib/install.go
```diff
@@ -119,12 +119,12 @@ func Install(tfversion string, binPath string, mirrorURL string) {
...
 	/* proceed to download it from the hashicorp release page */
-	url := mirrorURL + tfversion + "/" + versionPrefix + tfversion + "_" + goos + "_" + goarch + ".zip"
-	zipFile, errDownload := DownloadFromURL(installLocation, url)
+	zipFile, errDownload := DownloadFromURL(installLocation, mirrorURL, tfversion, versionPrefix, goos, goarch)
 
 	/* If unable to download file from url, exit(1) immediately */
 	if errDownload != nil {
-		logger.Fatalf("Error downloading %s: %v", url, errDownload)
+		logger.Fatalf("Error downloading: %v", errDownload)
+		os.Exit(1)
```
yermulnik, line 127: `logger.Fatalf()` performs `is.Exit(1)`, doesn't it? 🤔 
```suggestion
```

lib/utils.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"os"
+)
+
+// FileExists checks if a file exists and is not a directory before we try using it to prevent further errors.
```
yermulnik, line 7: nitpick for consistency
```suggestion
// FileExists checks if a file exists and is not a directory before we try using it to prevent further errors
```

main.go
```diff
@@ -64,6 +63,10 @@ func main() {
 	chDirPath := getopt.StringLong("chdir", 'c', dir, "Switch to a different working directory before executing the given command. Ex: tfswitch --chdir terraform_project will run tfswitch in the terraform_project directory")
 	versionFlag := getopt.BoolLong("version", 'v', "Displays the version of tfswitch")
 	defaultVersion := getopt.StringLong("default", 'd', defaultLatest, "Default to this version in case no other versions could be detected. Ex: tfswitch --default 1.2.4")
+	getopt.StringVarLong(&lib.PubKeyId, "public-key-id", 'k', "The ID of the public key to check the checksums against.")
+	getopt.StringVarLong(&lib.PubKeyPrefix, "public-key-prefix", 'x', "The prefix of the public key. i.e. hashicorp_")
+	getopt.StringVarLong(&lib.PubKeyUri, "public-key-uri", 'y', "The URI to download the public key from.")
```
yermulnik, line 68: Since these are a bit different from other opts I'd separate them visually + consistency with other opts' descriptions
```suggestion

	getopt.StringVarLong(&lib.PubKeyId, "public-key-id", 'k', "The ID of the public key to check the checksums against")
	getopt.StringVarLong(&lib.PubKeyPrefix, "public-key-prefix", 'x', "The prefix of the public key. i.e. \"hashicorp_\"")
	getopt.StringVarLong(&lib.PubKeyUri, "public-key-uri", 'y', "The URI to download the public key from")
```
Also ain't public key is to verify the **signature** rather than the **checksums**? 😕

yermulnik (commented):
Among other comments, I made my best to align logging messages

lib/install.go
MatrixCrawler: I had the same phenomen. It should exit but it doesn't. Hence my manual exit calls.
Maybe I am missing a flag on the logger config or something.
I will have another look. Otherwise I will open an issue with slog

lib/checksum.go
MatrixCrawler: I actually don't have a clue 😂 
my go knowledge is not that deep. I am learning with this project 😁

MatrixCrawler: You are right. Will fix that

lib/download.go
```diff
@@ -9,11 +9,86 @@ import (
...
+		cleanup(filesToCleanup)
+		return "", errors.New("Signature of checksum file could not be verified")
+	}
+	match := checkChecksumMatches(hashFilePath, targetFile)
+	if !match {
+		cleanup(filesToCleanup)
+		return "", errors.New("Checksums did not match")
+	}
+	cleanup(filesToCleanup)
+	return zipFilePath, err
```
yermulnik, line 85: We probably can do something like the below instead of adding `cleanup(filesToCleanup)` three times? (I'm not very familiar with Go and hence might be mixing up things)
```suggestion
    defer cleanup(filesToCleanup)

	verified := checkSignatureOfChecksums(publicKeyFile, hashFile, signatureFile)
	if !verified {
		return "", errors.New("Signature of checksum file could not be verified")
	}

	match := checkChecksumMatches(hashFilePath, targetFile)
	if !match {
		return "", errors.New("Checksums did not match")
	}

	return zipFilePath, err
```

yermulnik (approved):

lib/download.go
MatrixCrawler: Will call defer cleanup which will automatically called on return

MatrixCrawler:
Thank you for all your effor @yermulnik 😄

Issue #339, opened 2024-04-01
Installation via install.sh fails for version 1.0.0


MatthewJohn:
Installation fails using install.sh after release of v1.0.0

**Describe alternatives you've considered**

**Additional context**

Script output with debug:
```
warrensbox/terraform-switcher info checking GitHub for latest tag
warrensbox/terraform-switcher debug http_download https://github.com/warrensbox/terraform-switcher/releases/latest
warrensbox/terraform-switcher info found version: 1.0.0 for 1.0.0/linux/amd64
warrensbox/terraform-switcher debug downloading files into /tmp/tmp.7kYA692ao2
warrensbox/terraform-switcher debug http_download https://github.com/warrensbox/terraform-switcher/releases/download/1.0.0/terraform-switcher_1.0.0_linux_amd64.tar.gz
warrensbox/terraform-switcher debug http_download_curl received HTTP status 404
```

MatthewJohn:
All of the artifacts contain a leading dot in the version:
```
terraform-switcher_.1.0.0_Darwin_amd64.tar.gz 
```

MatthewJohn:
It looks like the gorelease has a space in the path (trying to work out if this is the 'tag' output from the bump action), which I suspect is then being converted to a dot:
```
      • uploading to release                         file=dist/terraform-switcher_ 1.0.0_Windows_armv7.tar.gz name=terraform-switcher_ 1.0.0_Windows_armv7.tar.gz
```

arturmartins:
Related:

```
❯ brew install warrensbox/tap/tfswitch

Error: invalid attribute for formula 'warrensbox/tap/tfswitch': url ("https://github.com/warrensbox/terraform-switcher/releases/download/1.0.0/terraform-switcher_ 1.0.0_Darwin_arm64.tar.gz")
Please report this issue:
```

yermulnik:
@warrensbox FYI (also checksums file is empty 🤔)

Meanwhile I updated all achieve names by removing the extraneous dot.

@arturmartins @MatthewJohn Could you please confirm you're not having issues with downloading artifacts now?

yermulnik:
Dang, Homebrew tap has spaces in filenames 😞 Need @warrensbox here as I have no write perms to tap repo (and I'm still trying to figure out what went wrong with goreleaser)

yermulnik:
Fixed and re-uploaded checksums file (manually at this time).

patrick-steadman:
[...]

PR #346, opened 2024-04-01
fix: Downgrade GH Ubuntu runner to `20.04`

### LINKED ISSUE #344: Latest release 1.0.0 does not support glibc 2.31 or lower
The latest 1.0.0 release has a dependency on glibc 2.32 or higher which is not available on Ubuntu 20.04 LTS natively. This is not called out in the release notes, so anyone with an older OS may experience issues that were not present in the previous version of tfswitch.

Ideally backwards compatibility would be nice, but if that's not feasible it should at least be called out in the release notes for the uninformed. 

I used the tag feature to lock the version to 0.13.1308 which still works. 
```sh
curl -L https://raw.githubusercontent.com/warrensbox/terraform-switcher/release/install.sh |  bash -s -- -b  $HOME/.local/bin 0.13.1308
```
Or for those using the CI steps from documentation like me:
```sh
wget https://raw.githubusercontent.com/warrensbox/terraform-switcher/release/install.sh  #Get the installer on to your machine:

chmod 755 install.sh #Make installer executable

./install.sh -b $HOME/.bin -d 0.13.1308 #Install tfswitch in a location you have permission and lock the version to 0.13.1308 for those without a newer version of glibc:

$HOME/.bin/tfswitch #test

export PATH=$PATH:$HOME/.bin #Export your .bin into your path

#You should probably add step 4 in your `.bash_profile` in your $HOME directory.

#Next, try:
`tfswitch -b $HOME/.bin/terraform 0.11.7` 

#or simply 

`tfswitch -b $HOME/.bin/terraform`
```
**Additional context**
```sh
$ lsb_release -a
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.6 LTS
Release:        20.04
Codename:       focal

$ ldd --version | head -n1
ldd (Ubuntu GLIBC 2.31-0ubuntu9.14) 2.31

# Error
$ tfswitch
tfswitch: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by tfswitch)
tfswitch: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by tfswitch)
```

### LINKED ISSUE #345: GLIBC 2.31 - No Longer Supported ?
Version 0.13.1300 worked with glibc 2.31.

Version 1.0.0+ now require glibc 2.32 or 2.34. 

Was this an intentional change ? 

We use tfswitch primarily on an older (but still supported albeit only for another couple of months) Debian Bullseye container. As a result of this change, we can no longer use the latest version of tfswitch.  

We are planning to upgrade our build container regardless, but wanted to ensure that this was a deliberate change rather than accidental.

yermulnik:
fixes #344, fixes #345
relates to #339

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
   integration_tests_linux:
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-20.04]
         go_version: ['1.22']
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   tfswitch-release:
     name: Release
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
 
     # Checkout code from repo


warrensbox (approved):
LGTM

warrensbox:
This should be released

PR #353, opened 2024-04-02
feat: Build statically linked binaries

yermulnik:
- Switch back to `ubuntu-latest` GH runners but build statically linked Go binaries instead
  - Confirmed Linux `amd64` binary works on Linux Mint 19.3 Tricia (based on Ubuntu Bionic 18.04)
  - No penalty on the size of binaries as of look of it
- While here, brush up and align GH Workflow YAML configs a bit

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: Run test
+name: Run tests
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


warrensbox:
I need to test this first throughly. I will try to do it tonight.

yermulnik:
We should be safe with this as we don't use `net` and the use of `os/user` belongs in tests only and seemingly doesn't depend on `libc`. Though we can force Go to use Go-native implementation of `us/user` calls just to stay on the safe side (which looks to be an overhead in our use case at this moment). More info: https://mt165.co.uk/blog/static-link-go/

.github/workflows/build.yml
```diff
@@ -1,24 +1,29 @@
...
 
 on:
   push:
     branches: 
       - '*'         # matches every branch that doesn't contain a '/'
       - '*/*'       # matches every branch containing a single '/'
       - '**'        # matches every branch
-      - '!release'   # excludes master
-jobs:
+      - '!release'  # excludes `release` branch
```
warrensbox, line 9: Can you change this to master

yermulnik: Yep. Done: 2ae037f

warrensbox (approved):

warrensbox:
@yermulnik are you able to Merge this and deploy it?

yermulnik:
@warrensbox No luck 😢 https://github.com/warrensbox/terraform-switcher/actions/runs/8538222695/job/23390434393
Also the `version` bit is still acting up 🤔 
image

PR #354, opened 2024-04-03
Update Go Package Index

### LINKED ISSUE #192: Update Golang Package Version on pkg.go
Actually a very old version is in there and miss most of the existing functions, would be nice to get this updated in the pkg.go repo

MatrixCrawler:
This possibly fixes #192 
We need to add a v in front of the versioning tags though


.github/workflows/publish.yml
```diff
@@ -0,0 +1,15 @@
...
+name: Update Go Module Index
+
+jobs:
+  bump-index:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+      - name: Ping endpoint
+        run: curl "https://proxy.golang.org/github.com/warrensbox/terraform-switcher/@v/$(git describe HEAD --tags --abbrev=0).info"
```
yermulnik, line 15: Please add correct EOF (carriage return)
image

.github/workflows/release.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       id: semver-tag-dry 
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        WITH_V: false
+        WITH_V: true
```
yermulnik, line 56: While this looks good, though since this is the change to the version tagging format, I'd suggest to postpone this item to some time later as we need to consider how much this is beneficial for us to have `tfswitch` as a Go module.
@warrensbox What do you think?

crablab: I think the reason people want this updated is so `go get` works as expected. I don't really see a disbenefit, as long as changing the format doesn't breaking anything downstream

yermulnik: Unfortunately we can't guarantee none of the consumers/downstream rely upon versioning without `v` as prefix.
It's also off that they require(?) `v` whilst it is not a part of valid semver 😕

MatrixCrawler: It's a bit odd, but if you have a look into the go.mod you'll see that all the modules have versions with a "v" Prefix
So it seems to be the thing GO is versioning 🤷

yermulnik: > all the modules have versions with a "v" Prefix

`tfswitch` is an app firstly as opposite to Go module 😺 
Let's see what @warrensbox thinks of this.

crablab: > Unfortunately we can't guarantee none of the consumers/downstream rely upon versioning without v as prefix.

Sure, but this isn't breaking previous releases. It's just a breaking change going forwards if you're relying on a specific format to the version number. 
I wouldn't be very concerned about changing that convention. 

> is an app firstly as opposite to Go module 

Agreed. Not too fussed either way - maybe if someone heavily in the Go community chimes in on whether we should/should not be following the module convention as an app that'd help.

yermulnik: > It's just a breaking change going forwards if you're relying on a specific format to the version number.

Yep, this is what I mean.

> I wouldn't be very concerned about changing that convention.

If we announce that and provide some time, then we probs can break what's been in place from the very beginning...
Either way I'd leave this for @warrensbox to consider and decide.

warrensbox: I agree with @yermulnik; we should provide a heads-up in the README.md about this. That way, people who have been using it without the 'v' can adjust their workflow accordingly.

I also agree that tfswitch if more of a tool rather than a module.

Not super urgent, but we can announce that we are switching on the README, and set a target version that would include "v"

warrensbox: I still need to update the README on this

MatrixCrawler:
I updated the readme with the version note and the documentation for some missing points.
i reduced the readme file as most of the stuff is covered by the documentation and i think no one really wants to read a five page Readme ;-)
But maybe i am wrong with this.

crablab:
> I updated the readme with the version note and the documentation for some missing points.
> i reduced the readme file as most of the stuff is covered by the documentation and i think no one really wants to read a five page Readme ;-)

Please could you split these changes out into a separate PR? :sweat_smile:  We now have a couple of PRs with large deltas and which just keep getting longer and longer. That makes it really tricky to review as it's not clear when a PR is "done", and we're ending up in endless cycles of comments as new stuff is being added all the time :) 

I think it'd help with review if we can keep them tightly scoped to a particular goal, even if that means we need to come back with more PRs in the future.

MatrixCrawler:
> > I updated the readme with the version note and the documentation for some missing points.
> > i reduced the readme file as most of the stuff is covered by the documentation and i think no one really wants to read a five page Readme ;-)
> 
> Please could you split these changes out into a separate PR? 😅 We now have a couple of PRs with large deltas and which just keep getting longer and longer. That makes it really tricky to review as it's not clear when a PR is "done", and we're ending up in endless cycles of comments as new stuff is being added all the time :)
> 
> I think it'd help with review if we can keep them tightly scoped to a particular goal, even if that means we need to come back with more PRs in the future.

Moved the readme and documentation update to #376

yermulnik:
> That makes it really tricky to review as it's not clear when a PR is "done", and we're ending up in endless cycles of comments as new stuff is being added all the time :)

+1
force pushes also make it hard to track changes linked from the email notifications.
I'd suggest either to force-push right before the merge, or (preferably) rely upon GH feature to squash and merge when you press the green button to merge PR.

yermulnik:
> I will try to rebase less during the review :)

Thanks.
Though please take into consideration whet I wrote: there's actually no need to do that in most of the cases in current conditions 😺 The only meaningful time to force-push is right before you call for review.

crablab (approved):
LGTM - Warren gave a :+1:  in the comments, I believe. 

Thanks @MatrixCrawler for splitting the PR :)

PR #359, opened 2024-04-05
docs: Actualize CHANGELOG

yermulnik:
- Add changelog entries for each tag starting from the one that already existed in CHANGELOG
- Utilized [chelnak/gh-changelog](https://github.com/chelnak/gh-changelog) to automate process
  - Manually adjusted where `gh-changelog` didn't make its job good enough

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-<!-- markdownlint-disable MD024 -->
+<!-- markdownlint-disable MD013 MD024 MD043 -->
 # Changelog
 
 All notable changes to this project will be documented in this file.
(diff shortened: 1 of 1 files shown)


crablab:
Could we document how we're using this extension, since we're going to need contributors to run it on every PR. 

More generally we'll probably need some contribution guidelines.

MatrixCrawler (approved):

yermulnik:
@crablab I'm merging this at this moment and will try and compile contribution notes when we have the process more or less established.

warrensbox:
@yermulnik can you please take a look at this as well?
https://goreleaser.com/customization/changelog/
Since we are already using goreleaser

yermulnik:
> @yermulnik can you please take a look at this as well?
> [goreleaser.com/customization/changelog](https://goreleaser.com/customization/changelog/)
> Since we are already using goreleaser

Thanks for the pointer. Will look into it after weekend.

yermulnik:
> > @yermulnik can you please take a look at this as well?
> > [goreleaser.com/customization/changelog](https://goreleaser.com/customization/changelog/)
> > Since we are already using goreleaser
> 
> Thanks for the pointer. Will look into it after weekend.

So I only now had a time to look into this and.. and, only after I had changed `.goreleaser.yaml`, I found that this actually is for release changelog, rather than for `CHANGELOG.md` 🤷🏻 
https://goreleaser.com/quick-start/
изображение
Though I decided to create a PR out of this (just in case we may want it).

PR #360, opened 2024-04-05
Release Workflow fix

MatrixCrawler:
It seems that the version update only works over the PR way. With this i try to solve that problem.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,7 +33,7 @@ jobs:
       uses: actions/setup-go@v5
       with:
         go-version: '1.22' # The Go version to download (if necessary) and use
-  
+
     # Double check Go version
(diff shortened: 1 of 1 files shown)


yermulnik (approved):

MatrixCrawler:
I would like to let @warrensbox make the merge.
It might be necessary to exempt PR that are created by GitHub actions from the "needs two reviews" restrictions

warrensbox:
Let me test this first on another repo

warrensbox:
I cannot get this to work right

MatrixCrawler:
You are changing the version file in the workflow, but github does not allow pushes on the master branch directly. Changes have to be done via a PR.
This fix should create a temporary release-branch, a PR and merge this PR automatically to update the Version file.

PR #363, opened 2024-04-05
bugfix: Fix Fatal logger calls by force exit

yermulnik:
Attempt to hastily fix Fatal level calls to logger by appending `os.Exit(1)` to each `logger.Fatal()` (`logger.Fatalf()`) call, as it appeared that it doesn't exit as we'd expect it to.

Refs:
- #334 (comment)
- gookit/slog#143

### DIFF
--- a/lib/defaults.go
+++ b/lib/defaults.go
@@ -1,8 +1,10 @@
 package lib
 
 import (
-	"github.com/mitchellh/go-homedir"
+	"os"
 	"runtime"
+
+	"github.com/mitchellh/go-homedir"
 )
 
 // GetDefaultBin Get default binary path
@@ -12,6 +14,7 @@ func GetDefaultBin() string {
 		home, err := homedir.Dir()
 		if err != nil {
 			logger.Fatal("Could not detect home directory.")
+			os.Exit(1)
 		}
 		defaultBin = home + "/bin/terraform.exe"
 	}
--- a/lib/download_test.go
+++ b/lib/download_test.go
@@ -23,7 +23,6 @@ func TestDownloadFromURL_FileNameMatch(t *testing.T) {
 	home, err := homedir.Dir()
 	if err != nil {
 		logger.Fatalf("Could not detect home directory")
-		os.Exit(1)
 	}
 
 	logger.Infof("Current home directory: %q", home)
--- a/lib/download_test.go
+++ b/lib/download_test.go
@@ -2,12 +2,13 @@ package lib
 
 import (
 	"fmt"
-	"github.com/mitchellh/go-homedir"
 	"net/url"
 	"os"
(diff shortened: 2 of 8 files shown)


yermulnik:
2reviewers: I just added `os.Exit(1)` to every single `logger.Fatal()` call. Including tests. Which (the latter) may probably break some testing logic — please review them as I'm not familiar with Go at the necessary level 🤷🏻 Thank you.

MatrixCrawler (changes requested):
I don't think we need os.Exit(1) in tests (nor fatal logs)
My expectation is that the tests fails and all the remaining tests are still executed (which they are not if we call os.Exit(1))

yermulnik:
> I don't think we need os.Exit(1) in tests (nor fatal logs)
> My expectation is that the tests fails and all the remaining tests are still executed (which they are not if we call os.Exit(1))

Reverted test files: 6190226

MatrixCrawler (approved):
LGTM.
Thanks for your effort

yermulnik:
👍🏻 
@warrensbox @MatthewJohn @crablab @jukie Please let me know in case explicit exit is needed for tests and I'll add with a separate PR. Thanks.

yermulnik:
So the exit function has to be explicitly declared at logger init (https://github.com/gookit/slog/issues/143). I'll work on this change in near days. Caveat ℹ️: this would make `logger.Fatal()` calls in test to fail with exit code 1 as well.

MatrixCrawler:
I will have a look at the test logs. In tests error level should suffice.
And we'd have to look att all other fatals if we'd REALLY really want to exit at this point

yermulnik:
> I will have a look at the test loggers. In tests error level should suffice.

Thank you.
I'd also much appreciate if you could look into configuring logger as per https://github.com/gookit/slog/issues/143 — given my knowledge of Golang is of a sysadmin level (as in script simple code bits), for me this task is more of a taking wild guesses on how to do that correctly =)
Along with that the other part of this task is to simply remove all the `os.Exit(1)` statements that directly follow `logger.Fatal()` and `logger.Fatalf()` calls.

> And we'd have to look att all other fatals if we'd REALLY really want to exit at this point

When I was reviewing logging re-factoring PR and when I was adding `os.Exit(1)` to all fatals, I tried my best to re-consider whether Fatal is needed or not. So hopefully all of them (or at least vast majority of them) are correctly used.

The other related task I'd think of is to provide an option for a user to customize logging level so that they can reduce or increase amount of log messages they see in console. With that done, we might re-visit all logging statements and configure them more granular by employing more Warns, Notices, Debugs, and Traces (at the moment we mainly use Fatals, Errors and Infos, with some sprinkling with Panic, Notice and Debug I guess).

PR #375, opened 2024-04-11
Update dependabot.yml

MatrixCrawler:
Update labels for dependabot as the labels don't exists in the default.
We do, however have a dependency label that could be used.
This is just mere convenience. Before that the bot tried to set non existing labels.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,8 +13,7 @@ updates:
     commit-message:
       prefix: "go:"
     labels:
-      - c:deps
-      - golang
+      - dependencies
   - package-ecosystem: "github-actions"
     directory: "/"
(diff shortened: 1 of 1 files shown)


MatrixCrawler:
> Why do we need to change the labels it's applying? Please could you expand the description :) 

Have updated the pr description 😊

yermulnik (commented):
Should we add labels to `github-actions` ecosystem part of this config (and maybe commit message prefix — prefixes help manage sections in auto-generated changelog entries, which is WIP yet)?

MatrixCrawler:
Added prefix and labels for github actions as well as suggested from @yermulnik

yermulnik (approved):

Issue #394, opened 2024-04-25
install.sh tries to download 1.1.0 instead of v1.1.0


d33psky:
install tries this url

https://github.com/warrensbox/terraform-switcher/releases/download/v1.1.0/terraform-switcher_1.1.0_linux_amd64.tar.gz

while the url should be

https://github.com/warrensbox/terraform-switcher/releases/download/v1.1.0/terraform-switcher_v1.1.0_linux_amd64.tar.gz

note the missing v in the second 1.1.0

warrensbox:
This is fixed. Please let us know it's working now! 

Thanks for the fix @d33psky !

PR #395, opened 2024-04-25
this fixes issue #394

d33psky:


warrensbox (approved):
Nice catch. And thank you!

d33psky:
hmm while this fixes 1.1.0 it breaks for older versions

Issue #400, opened 2024-04-25
Error in v1.10 parsing version


AndyLonghurstHESA:
Since the latest version has been released, we have broken pipelines that cannot resolve the TF version. We did not have issue issue yesterday. It is complaining of a double = character. Checking the versions.tf we have "=1.7.5" specified.

`2024/04/25 10:00:56.109010 Running command tfswitch
10:00:56.160 FATAL Error parsing configuration file: "error parsing constraint: Malformed constraint: = =1.7.5"`

Versions.tf file:
`required_version = "=1.7.5"`

MatrixCrawler:
I cannot reproduce that in the moment.

My version.tf looks like this

```
terraform {
  required_version = "= 1.7.5"
}
```

which works fine.

MatrixCrawler:
I can reproduce it when there is no space between the = and the version number
Invalid version.tf
```
terraform {
  required_version = "=1.7.5"
}
```

MatrixCrawler:
The error is here
https://github.com/warrensbox/terraform-switcher/blob/master/lib/param_parsing/versiontf.go#L44-L52

We will fix that. Thanks for pointing out.

PR #403, opened 2024-04-25
Fix for #400 - Whitespaces and Version Constraints

### LINKED ISSUE #402: We are facing the issue with the latest version(v1.1.0) of tfswitcher
Our pipeline was running fine with v1.0.2, but suddenly in one step our pipeline download the latest tfswitcher(v1.1.0) and then pipeline got failed. We are facing the below error :

FATAL Error parsing configuration file: "exact constraint (\"= 1.1.9, = 1.1.9\") cannot be combined with other conditions"

MatrixCrawler:
- added fix for #400 when the comparator is not separated from the version number by a space
- possibly fixes #402 as well

### DIFF
--- a/lib/param_parsing/versiontf.go
+++ b/lib/param_parsing/versiontf.go
@@ -1,7 +1,7 @@
 package param_parsing
 
 import (
-	"fmt"
+	semver "github.com/hashicorp/go-version"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -13,7 +13,7 @@ import (
 
 func GetVersionFromVersionsTF(params Params) (Params, error) {
 	var tfConstraints []string
-	var exactConstraints []string
+	//var exactConstraints []string
 
 	curDir, err := os.Getwd()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


lib/param_parsing/versiontf.go
```diff
@@ -77,3 +78,9 @@ func isTerraformModule(params Params) bool {
 	module, err := tfconfig.LoadModule(params.ChDirPath)
 	return err == nil && len(module.RequiredCore) > 0
 }
+
+func cleanupVersionConstraints(constraint string) string {
+	regex := regexp.MustCompile(`(?P<Comparator>\D+)(?P<Version>(\d+\.\d+\.\d+)(-[a-zA-z]+\d*)?)$`)
```
yermulnik, line 72: @MatrixCrawler First of all thanks for taking care of it — I indeed missed "improperly" formatted version constraints where constraint operator isn't separated from the version =(
My point is looking up `(\d+\.\d+\.\d+)` is incorrect since version can be specified as `major`, `major.minor`, and `major.minor.patch` (with an optional suffix that should fall under semver though). Hence we probably should better look up constraint operators instead: https://developer.hashicorp.com/terraform/language/expressions/version-constraints#version-constraint-syntax — `=`, `!=`, `>`, `>=`, `<`, `<=`, `~>`. And ideally follow what TF Config Inspect does: https://github.com/hashicorp/terraform-config-inspect/blob/master/tfconfig/load_hcl.go#L71-L78 (didn't go deeper though)

MatrixCrawler: I use the hashicorp version constraint checker now. no need for manually checking.

yermulnik: Ah, just saw the ce0909ec0b8abb0d5478a944e1eabd1e23e7fcdc 👍🏻

lib/param_parsing/versiontf.go
```diff
@@ -12,7 +12,7 @@ import (
 
 func GetVersionFromVersionsTF(params Params) (Params, error) {
 	var tfConstraints []string
-	var exactConstraints []string
+	//var exactConstraints []string
```
yermulnik, line 15: We probably can drop this as it is not used anymore

yermulnik (approved):

lib/param_parsing/versiontf.go
MatrixCrawler: Thanks for finding. Deleted.

Issue #296, opened 2023-03-24
Add ability to show version (and not download) when parsing version from version.tf file


ocofaigh:
**Is your feature request related to a problem? Please describe.**
`tfswitch --show-latest` will always display the latest version. However it seems there is no way to display the version that would be installed when reading from a version.tf file without actually installing it:
```
% tfswitch              
Reading required version from terraform file
Reading required version from constraint: >= 1.3, < 1.4
Matched version: 1.3.9
```

**Describe the solution you'd like**
Looking for a flag to be able to output 1.3.9 from above example without actually downloading it

**Describe alternatives you've considered**

**Additional context**

## Required version
```sh
go version 1.13
```

### Step 1 - Create workspace
*Skip this step if you already have a github go workspace*   
Create a github workspace.
<img src="https://s3.us-east-2.amazonaws.com/kepler-images/warrensbox/tfswitch/tfswitch-workspace.gif" alt="drawing" style="width: 600px;"/>   

### Step 2 - Set GOPATH
*Skip this step if you already have a github go workspace*    
Export your GOPATH environment variable in your `go` directory.   
```sh
export GOPATH=`pwd`
```
<img src="https://s3.us-east-2.amazonaws.com/kepler-images/warrensbox/tfswitch/tfswitch-gopath.gif" alt="drawing" style="width: 600px;"/>   

### Step 3 - Clone repository
Git clone this repository.  
```sh 
git clone git@github.com:warrensbox/terraform-switcher.git
```
<img src="https://s3.us-east-2.amazonaws.com/kepler-images/warrensbox/tfswitch/tfswitch-git-clone.gif" alt="drawing" style="width: 600px;"/>  

### Step 4 - Get dependencies
Go get all the dependencies.   

```sh 
go mod download
```
```sh 
go get -v -t -d ./...
```
Test the code (optional).
```sh  
go vet -tests=false ./...
```
```sh 
go test -v ./...
```
<img src="https://s3.us-east-2.amazonaws.com/kepler-images/warrensbox/tfswitch/tfswitch-go-get.gif" alt="drawing" style="width: 600px;"/>  

### Step 5 - Build executable
Create a new branch.   
```sh 
git checkout -b feature/put-your-branch-name-here
```
Refactor and add new features to the code.  
Go build the code.   
```sh 
go build -o test-tfswitch
```
Test the code and create a new pull request!

[...]

PR #407, opened 2024-04-25
Display Terraform Version

### LINKED ISSUE #398: New version v1.1.0 does not work as expected
Hello, there is some issue with the latest version which has been released today - it determines wrongly Terraform version "09" while it was working fine until today. Please kindly fix the issue.

warrensbox/terraform-switcher info checking GitHub for latest tag
warrensbox/terraform-switcher info found version: 1.1.0 for v1.1.0/linux/amd64
warrensbox/terraform-switcher info installed /home/AzDevOps/.bin/tfswitch

Trying to read required terraform binary version from terraform files within /agent/_work/1/s/edited
TFVER=09

MatrixCrawler:
- add function to display the current version of terraform that was selected by tfswitch

Probably closes #398 and relates to #296

### DIFF
--- a/lib/param_parsing/parameters.go
+++ b/lib/param_parsing/parameters.go
@@ -21,9 +21,9 @@ type Params struct {
 	ShowLatestFlag   bool
 	ShowLatestPre    string
 	ShowLatestStable string
+	ShowSelectedFlag bool
 	Version          string
 	VersionFlag      bool
-	WhichFlag        bool
 }
 
 var logger *slog.Logger
@@ -46,8 +46,8 @@ func GetParameters() Params {
 	getopt.BoolVarLong(&params.ShowLatestFlag, "show-latest", 'U', "Show latest stable version")
 	getopt.StringVarLong(&params.ShowLatestPre, "show-latest-pre", 'P', "Show latest pre-release implicit version. Ex: tfswitch --show-latest-pre 0.13 prints 0.13.0-rc1 (latest)")
 	getopt.StringVarLong(&params.ShowLatestStable, "show-latest-stable", 'S', "Show latest implicit version. Ex: tfswitch --show-latest-stable 0.13 prints 0.13.7 (latest)")
+	getopt.BoolVarLong(&params.ShowSelectedFlag, "which", 'w', "Displays the currently selected terraform version by tfswitch")
 	getopt.BoolVarLong(&params.VersionFlag, "version", 'v', "Displays the version of tfswitch")
-	getopt.BoolVarLong(&params.WhichFlag, "which", 'w', "Displays the currently selected terraform version by tfswitch")
 
 	// Parse the command line parameters to fetch stuff like chdir
 	getopt.Parse()
@@ -97,6 +97,7 @@ func initParams(params Params) Params {
 	params.ShowLatestFlag = false
 	params.ShowLatestPre = lib.DefaultLatest
 	params.ShowLatestStable = lib.DefaultLatest
+	params.ShowSelectedFlag = false
 	params.Version = lib.DefaultLatest
 	params.VersionFlag = false
 	return params
--- a/lib/param_parsing/parameters.go
+++ b/lib/param_parsing/parameters.go
@@ -23,6 +23,7 @@ type Params struct {
 	ShowLatestStable string
 	Version          string
 	VersionFlag      bool
+	WhichFlag        bool
 }
 
 var logger *slog.Logger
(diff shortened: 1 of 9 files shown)


warrensbox:
I would think that users would simply do `terraform -v` do get the latest version

warrensbox:
Let me test this out

yermulnik:
I actually am supporting what @warrensbox says, while I also can see the use case for this feature (e.g. when `tfswtich`-managed `terraform` **symlink** is called from the custom location which is not on the `PATH` or doesn't precede bin directory with another `terraform` **binary**).

Though I think this feature needs two improvements:
* Name of the parameter: I reckon it should be more specific, like e.g. `show-selected` to keep consistency with other similar parameters from https://github.com/warrensbox/terraform-switcher/blob/feature/which-flag/lib/param_parsing/parameters.go#L46C46-L48
* The logic should not rely upon the arbitrary file that can be changed from outside `tfswitch`. Since `tfswitch` installs TF binaries with version appended to the file name, I insist we resolve `terraform` symlink from the default bin dir (and respectively from `CustomBinaryPath` if defined by the user) to the real file and derive TF version from its name (was thinking that running resolved `terraform` binary should be much more reliable, though we re-link symlink w/o doing such, so it's superfluous).

MatrixCrawler:
I will give this a thougt. Thanks for the input.

MatrixCrawler:
I changed the approach in simply executing the binary, asking for the version and put out the data.

lib/param_parsing/parameters.go
```diff
@@ -45,6 +46,7 @@ func GetParameters() Params {
 	getopt.BoolVarLong(&params.ShowLatestFlag, "show-latest", 'U', "Show latest stable version")
 	getopt.StringVarLong(&params.ShowLatestPre, "show-latest-pre", 'P', "Show latest pre-release implicit version. Ex: tfswitch --show-latest-pre 0.13 prints 0.13.0-rc1 (latest)")
 	getopt.StringVarLong(&params.ShowLatestStable, "show-latest-stable", 'S', "Show latest implicit version. Ex: tfswitch --show-latest-stable 0.13 prints 0.13.7 (latest)")
+	getopt.BoolVarLong(&params.ShowSelectedFlag, "which", 'w', "Displays the currently selected terraform version by tfswitch")
```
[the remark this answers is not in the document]
MatrixCrawler: Forgot to change the flags....

yermulnik (commented):
LGTM
If @warrensbox is Ok with the implementation (calling `terraform --version` from `installPath` from within `tfswitch`), then I'm all for it

lib/param_parsing/parameters.go
```diff
@@ -45,6 +46,7 @@ func GetParameters() Params {
 	getopt.BoolVarLong(&params.ShowLatestFlag, "show-latest", 'U', "Show latest stable version")
 	getopt.StringVarLong(&params.ShowLatestPre, "show-latest-pre", 'P', "Show latest pre-release implicit version. Ex: tfswitch --show-latest-pre 0.13 prints 0.13.0-rc1 (latest)")
 	getopt.StringVarLong(&params.ShowLatestStable, "show-latest-stable", 'S', "Show latest implicit version. Ex: tfswitch --show-latest-stable 0.13 prints 0.13.7 (latest)")
+	getopt.BoolVarLong(&params.ShowSelectedFlag, "show-selected", 'w', "Display the currently selected terraform version by tfswitch")
```
crablab, line 49:
```suggestion
	getopt.BoolVarLong(&params.ShowSelectedFlag, "show-selected", 'w', "Display the currently selected Terraform version by tfswitch")
```

crablab, line 49: Where does the `w` option come from? 

Personally I would combine this with the `-v` option so it shows both the tfswitch version, and the currently selected Terraform version. I think this pattern is fairly common, where `-v` supplies versions for related components.

warrensbox:
> LGTM If @warrensbox is Ok with the implementation (calling `terraform --version` from `installPath` from within `tfswitch`), then I'm all for it

@yermulnik @MatrixCrawler @crablab It really comes down to - "Do really need this feature". I feel like the user can simply do `terraform --version` to get the version. Also, it may be only one person's issue. If they can simply run `terraform --version`, or modify their process to get their version, we do not have to modify our code. 

For https://github.com/warrensbox/terraform-switcher/issues/296, the `tfswitch --show-latest` is actually working as expected. It should just show the latest binary that terraform has to offer.

yermulnik:
> @yermulnik @MatrixCrawler @crablab It really comes down to - "Do really need this feature".

Citing myself =)
> [!NOTE]
> I actually am supporting what @warrensbox says, while I also can see the use case for this feature […]

I'm ok with declining this feature (the use case in #398 isn't really clear to me).

crablab:
I'd agree with above - I don't think this is something we need to add. 
My comment above stands, in that if this is included I think it should be bundled with the existing `-v` option.

PR #409, opened 2024-04-25
Fix terragrunt parsing

### LINKED ISSUE #401: Any content in terragrunt.hcl other than "terraform_version_constraint" causes tfswitch to error
Since this morning the following snippet:
```
terragrunt_version_constraint = "= 0.36.2"
terraform_version_constraint = "= 0.14.0"
```

Causes this output:
```
tfswitch --log-level=DEBUG
12:10:34.729 INFO [terragrunt.go:20,GetVersionFromTerragrunt] Reading configuration from "/redacted/terragrunt.hcl"  
12:10:34.729 FATAL [parameters.go:70,GetParameters] Error parsing configuration file: "could not decode body of HCL file \"/redacted/terragrunt.hcl\""  
```

However running like so:
```
terraform_version_constraint = "= 0.14.0"
```

Works as expected:
```
tfswitch --log-level=DEBUG                                                              ─╯
12:09:11.331 INFO [terragrunt.go:20,GetVersionFromTerragrunt] Reading configuration from "/redacted/terragrunt.hcl"  
12:09:11.331 DEBUG [list_versions.go:19,getTFList] Get list of terraform versions  
12:09:11.420 INFO [semver.go:14,GetSemver] Reading required version from constraint: "= 0.14.0"  
12:09:11.421 INFO [semver.go:42,SemVerParser] Matched version: "0.14.0"  
12:09:11.421 INFO [install.go:257,installableBinLocation] Installing terraform at "/home/redacted/bin"  
12:09:11.421 INFO [install.go:307,InstallVersion] Switched terraform to version "0.14.0" 
```
I've seen similar issues here this morning but they seem to be different outputs

MatrixCrawler:
This will fix #401 

* removed diagnostic.hasError check as it is expected to have an error every time something is present that we are not looking explicitly for.
* introduced test if a terraform_version_constraint can be found in hcl file.

### DIFF
--- a/lib/param_parsing/terragrunt.go
+++ b/lib/param_parsing/terragrunt.go
@@ -25,8 +25,8 @@ func GetVersionFromTerragrunt(params Params) (Params, error) {
 		}
 		var versionFromTerragrunt terragruntVersionConstraints
 		diagnostics = gohcl.DecodeBody(hclFile.Body, nil, &versionFromTerragrunt)
-		if diagnostics.HasErrors() {
-			return params, fmt.Errorf("could not decode body of HCL file %q", filePath)
+		if versionFromTerragrunt.TerraformVersionConstraint == "" {
+			return params, fmt.Errorf("could not find terraform_version_constraint in file %q", filePath)
 		}
 		version, err := lib.GetSemver(versionFromTerragrunt.TerraformVersionConstraint, params.MirrorURL)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


lib/param_parsing/terragrunt.go
```diff
@@ -25,8 +25,8 @@ func GetVersionFromTerragrunt(params Params) (Params, error) {
 		}
 		var versionFromTerragrunt terragruntVersionConstraints
 		diagnostics = gohcl.DecodeBody(hclFile.Body, nil, &versionFromTerragrunt)
-		if diagnostics.HasErrors() {
-			return params, fmt.Errorf("could not decode body of HCL file %q", filePath)
+		if versionFromTerragrunt.TerraformVersionConstraint == "" {
+			return params, fmt.Errorf("could not find terraform_version_constraint in file %q", filePath)
```
yermulnik, line 29: I might be missing something, but why would we want to return error if TG file has not explicit TF constraint? Should just ignore TG file instead in this case? 🤔

MatrixCrawler: Good idea. Implemented that

warrensbox (approved):
I tested this code and looks like everything is working as the issue pointed out.

PR #419, opened 2024-04-26
docs: Streamline/fix docu site files and README

yermulnik:


www/docs/usage/general.md
```diff
@@ -1,11 +1,14 @@
-## Order of variable precedence
+## Order of Terraform version definition precedence
```
warrensbox, line 1: @yermulnik based on our discussion, does this need to be updated?
https://github.com/warrensbox/terraform-switcher/discussions/420

yermulnik: Yes, this needs to. Though the current docu version (incl this PR) actually reflects current (latest) implementation and the change to reverse the order per discussion from https://github.com/warrensbox/terraform-switcher/discussions/420 has to be done in scope of PR to update the behavior and precedence I think (to avoid discrepancy between code and documentation).
What do you think?

yermulnik: I mean that a PR that will bring the change, should also introduce a change to the docu, rather than changing documentation ahead of time.

crablab: I agree @yermulnik. It also makes it easier to trace back when changes are made alongside the relevant documentation

PR #429, opened 2024-05-09
Check version sources based on whether verison has been set

MatthewJohn:
Update logic for reading sources for obtaining version to read from next source explicitly if version has not been set. This handles situations where a source exists (e.g. TOML file, terraform files), but a version constraint cannot be found from the source. This avoids duplicate checks for existance of sources (such as Terragrunt file), as this was performed in GetParamters and also in the methods to read the config from the source.

Issue #420


lib/param_parsing/parameters.go
```diff
@@ -56,17 +56,20 @@ func GetParameters() Params {
 	logger = lib.InitLogger(params.LogLevel)
 	var err error
 	// Read configuration files
-	if tomlFileExists(params) {
-		params, err = getParamsTOML(params)
-	} else if tfSwitchFileExists(params) {
+	params, err = getParamsTOML(params)
+	if params.Version == "" {
```
MatthewJohn, line 60: This feels fine*ish* if we _know_ the only param value we are obtaining is `Version`.
In the (unlikely?) event that we wish to obtain additional params from these, perhaps a better way is to obtain in order of least-to-highest priority, without the conditions, allowing each higher priority method to override the values from the previous. But, as long as we only care about `Version`, this is probably a more efficient method.

MatthewJohn: I'm also wondering if it's nicer to move the "obtain version" to a separate function, which calls (e.g. `GetVersionFromTerragrunt`), changing these GetVersion* methods to only return a version strring and then return-fast, e.g.:
```
func GetParams() {
...
  params = getParamsToml(params)
  if params.Version == "" {
     params.Version = getParamsFromSources(params)
   }
...
}

func getParamsFromSources(params) string {
    version := ""
    if version = GetVersionFromTfSwitch(params); version != "" {
        return version
    }
    if version = GetParamsFromTerraformVersion(params); version != "" {
        return version
    }
    ...
    return ""
}
```

MatthewJohn: Continued this discussion as it may play a role in https://github.com/warrensbox/terraform-switcher/discussions/420#discussioncomment-9364783

MatthewJohn:
I initially started looking at the, now, failing tests...
The first one I spotted was due to the TOML file failing to override the env variable for version - this is obviously testing the opposite of what was agreed in #420 (env variables have highest priority), so I'm going to leave these failing for now to get some feedback on the suggestion :)

lib/param_parsing/parameters.go
```diff
@@ -56,19 +56,52 @@ func GetParameters() Params {
...
+	// TOML from Homedir
+	if tomlFileExists(params, lib.GetHomeDirectory()) {
+		params, err = getParamsTOML(params, lib.GetHomeDirectory())
+		if err != nil {
+			logger.Fatalf("Failed to obtain settings from TOML config in home directory: %q", err)
+		}
+	}
+
+	// TOML from ChDirPath
+	if tomlFileExists(params, params.ChDirPath) {
```
MatthewJohn, line 68: I really don't like this - I did initially remove it and leave to getParamsTOML to take care of dealing with "does the file exist", since they both currently do it (not keen on the split of responsibility).
The _main_ reason I added it back in is that the GetVersionFromTerraformVerison (I think), doesn't have similar checks and if decoding Terraform fails, the entire thing fails.
I'm wondering if all of these methods (GetParamsFrom*) should:
 * Always be run (they can take care of checking if files exist, if applicable)
 * Always fail gracefully (maybe barring the TOML) and maybe just print out a warning if we fail to parse the source.

Mainly because we can't always guarantee that the way we're parsing them is correct (perhaps an old or newer version of the external application) OR the user might have additional templating etc (and be relying on the TOML file).

warrensbox:
Still reviewing...

lib/param_parsing/toml.go
warrensbox, line 1: // getParamsTOML parses everything in the toml file, return required version and bin path
func getParamsTOML(params Params) (Params, error) {
	tomlPath := filepath.Join(params.ChDirPath, tfSwitchTOMLFileName)
	if tomlFileExists(params.ChDirPath) {
		logger.Infof("Reading configuration from %q", tomlPath)
		configfileName := lib.GetFileName(tfSwitchTOMLFileName)
		viperParser := viper.New()
		viperParser.SetConfigType("toml")
		viperParser.SetConfigName(configfileName)
		viperParser.AddConfigPath(params.ChDirPath)

		errs := viperParser.ReadInConfig() // Find and read the config file
		if errs != nil {
			logger.Errorf("Could not to read %q: %v", tomlPath, errs)
			return params, errs
		}

		if viperParser.Get("bin") != nil {
			params.CustomBinaryPath = viperParser.GetString("bin")
		}
		if viperParser.Get("log-level") != nil {
			params.LogLevel = viperParser.GetString("log-level")
		}
		if viperParser.Get("version") != nil {
			params.Version = viperParser.GetString("version")
		}
	}
	return params, nil
}

func tomlFileExists(dir string) bool {
	tomlPath := filepath.Join(dir, tfSwitchTOMLFileName)
	if lib.CheckFileExist(tomlPath) {
		fmt.Println("FOunf toml")
	}
	return lib.CheckFileExist(tomlPath)
}

lib/param_parsing/parameters.go
```diff
@@ -56,19 +56,52 @@ func GetParameters() Params {
 	logger = lib.InitLogger(params.LogLevel)
 	var err error
 	// Read configuration files
-	if tomlFileExists(params) {
-		params, err = getParamsTOML(params)
-	} else if tfSwitchFileExists(params) {
+	// TOML from Homedir
+	if tomlFileExists(params, lib.GetHomeDirectory()) {
```
warrensbox, line 60: We do not need to pass params and lib.GetHomeDirectory()

lib/param_parsing/parameters.go
warrensbox, line 68: Same thing. See proposed changes on toml.go

lib/param_parsing/toml.go
```diff
@@ -38,7 +39,7 @@ func getParamsTOML(params Params) (Params, error) {
 	return params, nil
 }
 
-func tomlFileExists(params Params) bool {
-	tomlPath := filepath.Join(params.ChDirPath, tfSwitchTOMLFileName)
+func tomlFileExists(params Params, baseDir string) bool {
```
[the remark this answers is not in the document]
warrensbox: We are not really using params

lib/param_parsing/parameters.go
MatthewJohn: @warrensbox do you have any thoughts on these options? :)

lib/param_parsing/parameters.go
```diff
@@ -56,19 +56,52 @@ func GetParameters() Params {
...
 	var err error
 	// Read configuration files
-	if tomlFileExists(params) {
-		params, err = getParamsTOML(params)
-	} else if tfSwitchFileExists(params) {
+	// TOML from Homedir
+	if tomlFileExists(lib.GetHomeDirectory()) {
+		params, err = getParamsTOML(params, lib.GetHomeDirectory())
+		if err != nil {
+			logger.Fatalf("Failed to obtain settings from TOML config in home directory: %q", err)
```
yermulnik, line 63:
```suggestion
			logger.Fatalf("Failed to obtain settings from TOML config in home directory: %v", err)
```

lib/param_parsing/parameters.go
```diff
@@ -56,19 +56,52 @@ func GetParameters() Params {
...
+		}
+	}
+
+	// TOML from ChDirPath
+	if tomlFileExists(params.ChDirPath) {
+		params, err = getParamsTOML(params, params.ChDirPath)
+		if err != nil {
+			logger.Fatalf("Failed to obtain settings from TOML config in directory %s: %q", params.ChDirPath, err)
```
yermulnik, line 71:
```suggestion
			logger.Fatalf("Failed to obtain settings from TOML config in directory %q: %v", params.ChDirPath, err)
```

lib/param_parsing/parameters.go
```diff
@@ -56,19 +56,52 @@ func GetParameters() Params {
...
-	} else if tfSwitchFileExists(params) {
+	// TOML from Homedir
+	if tomlFileExists(lib.GetHomeDirectory()) {
+		params, err = getParamsTOML(params, lib.GetHomeDirectory())
+		if err != nil {
+			logger.Fatalf("Failed to obtain settings from TOML config in home directory: %q", err)
+		}
+	}
+
+	// TOML from ChDirPath
```
yermulnik, line 67: Based on https://github.com/warrensbox/terraform-switcher/discussions/420#discussioncomment-9369690 do we really want TOML in ChDirPath? 🤔

lib/param_parsing/parameters.go
```diff
@@ -56,19 +56,52 @@ func GetParameters() Params {
...
+		if err != nil {
+			logger.Fatalf("Failed to obtain settings from TOML config in directory %s: %q", params.ChDirPath, err)
+		}
+	}
+
+	if tfSwitchFileExists(params) {
 		params, err = GetParamsFromTfSwitch(params)
-	} else if terraformVersionFileExists(params) {
+		if err != nil {
+			logger.Fatalf("Failed to obtain settings from tfswitch config: %q", err)
```
yermulnik, line 78:
```suggestion
			logger.Fatalf("Failed to obtain settings from \".tfswitch\" file: %v", err)
```

lib/param_parsing/parameters.go
```diff
@@ -56,19 +56,52 @@ func GetParameters() Params {
...
+		}
+	}
+
+	if terraformVersionFileExists(params) {
 		params, err = GetParamsFromTerraformVersion(params)
-	} else if isTerraformModule(params) {
+		if err != nil {
+			logger.Fatalf("Failed to obtain settings from Terraform version: %q", err)
```
yermulnik, line 85:
```suggestion
			logger.Fatalf("Failed to obtain settings from \".terraform-version\" file: %v", err)
```

lib/param_parsing/parameters.go
```diff
@@ -56,19 +56,52 @@ func GetParameters() Params {
...
+		}
+	}
+
+	if isTerraformModule(params) {
 		params, err = GetVersionFromVersionsTF(params)
-	} else if terraGruntFileExists(params) {
+		if err != nil {
+			logger.Fatalf("Failed to obtain settings from .tfversions: %q", err)
```
yermulnik, line 92:
```suggestion
			logger.Fatalf("Failed to obtain settings from Terraform module: %v", err)
```

lib/param_parsing/parameters.go
```diff
@@ -56,19 +56,52 @@ func GetParameters() Params {
...
+		}
+	}
+
+	if terraGruntFileExists(params) {
 		params, err = GetVersionFromTerragrunt(params)
-	} else {
-		params = GetParamsFromEnvironment(params)
+		if err != nil {
+			logger.Fatalf("Error parsing configuration file: %q", err)
```
yermulnik, line 99:
```suggestion
			logger.Fatalf("Failed to obtain settings from Terragrunt configuration: %v", err)
```

lib/param_parsing/parameters.go
```diff
@@ -56,19 +56,52 @@ func GetParameters() Params {
...
+		}
 	}
+
+	params = GetParamsFromEnvironment(params)
+
 	if err != nil {
 		logger.Fatalf("Error parsing configuration file: %q", err)
 	}
```
yermulnik, line 107: Which `err` this relies upon eventually? Seems like this error handling is sort of redundant, isn't it? 🤔 
If it is not redundant, please use `%v` for errors instead of `%q`.

lib/param_parsing/parameters_test.go
```diff
@@ -91,17 +92,76 @@ func TestGetParameters_dry_run_wont_download_anything(t *testing.T) {
 	})
 }
 
-func TestGetParameters_check_config_precedence(t *testing.T) {
+func writeTestFile(t *testing.T, basePath string, fileName string, fileContent string) {
+	fullPath := filepath.Join(basePath, fileName)
+	if err := os.WriteFile(fullPath, []byte(fileContent), 0666); err != nil {
```
yermulnik, line 97: The `0666` is too permissive IMHO.

lib/param_parsing/parameters.go
MatthewJohn: I guess we could do with some weight in from @warrensbox , as this seemed quite specific in the main description:
```
But what is the .toml file is not in the HOME directory. That's where the option to pass the -c came in.

SO we still need to make the toml file readable as well incorporating the ability to have the version overridden.
```

lib/param_parsing/parameters.go
MatthewJohn: I agree. I was wondering if _any_ of this should actually cause an error.
If I'm writing terraform in a directory and want to get a terraform version initialised, if my terraform is invalid, should it mean that tfswitch errors out completely?

lib/param_parsing/parameters.go
yermulnik: 👍🏻 
I personally see no reason for TOML to exist out of HOME (or XDG Config) dir since this adds ambiguity as TOML defines not just version parameter.
Referencing back to you https://github.com/warrensbox/terraform-switcher/discussions/420#discussioncomment-9371421 also =)
There are other methods to define per-dir (incl `ChDirPath`) TF version and I'd prefer TOML to exist only in HOME (or XDG) to eliminate any kind of confusion that may add complexity understanding where the hiccup came from for users and for us when debugging issues.

lib/param_parsing/parameters.go
yermulnik: I'd support `tfswitch` to fail if any of config file is broken in any way (unreadable because of the the file perms, incorrect parameter or value, so on). As in ignore specific config file only if it doesn't exist.
So we actually could be ignoring "broken" config files to kinda "improve" user experience by trying to guess what they want, though I'm basing off of the point that if the user expects specific config file to provide TF version, we should explicitly fail and let user know they made a typo or any other error and hence they don't get the outcome they're expecting.

lib/param_parsing/parameters_test.go
MatthewJohn: I can change it for sure. Only bearing in mind that it's just for the tests and the file is created in the git tree, so if someone is using VSCode with a dev container which is running as root and the test fails to clear up the files, removing them from the host OS as a standard user would cause a little friction

yermulnik: I might be running a bit "too much", but write permission for group and others is a potential security breach that allows arbitrary actors to write to a file even during a relatively short period of time 🤷🏻 My take is always use what's defined by `umask` (incl user's override for system default)

MatthewJohn: I'm happy to make the change to make it more restrictive.
Though, from what I read in the docs for the functions:
 * WriteFile (https://pkg.go.dev/os#WriteFile) - indicates that, if the file does not exist, it creates the file with the "required permission (before umask)"

However, `io.Create`, the function that was also used in tfswitche's files.go states:
`if the file does not exist, it is created with mode 0666 (before umask)`
and this does appear to take the umask into account.

yermulnik: > Though, from what I read in the docs for the functions:

Ah, my bad, I just re-read the docu and tested it and it indeed takes my custom umask value into account.
So the `perm` looks to be a fallback, though my understanding of "sane" differs =)
I'd appreciate if you changed it to `0600` please as it's apparently us to look after security when using these bits in Go 🤪

lib/param_parsing/parameters.go
warrensbox: Here are some options:

@yermulnik @MatthewJohn 
- users cannot specify a **version** in the TOML file. The TOML is used for all other configurations except **version**
- only the TOML file in HOME can have a version (which later can be overriden)

Thoughts?

warrensbox: @yermulnik the TOML file is useful mostly for including the bin location. 
`bin: <path-to-install-tf>
If the user does not have access to the HOME directory, how would they specify the `bin` directory?

What is your suggestion?
- TOML only available in $HOME
- if TOML is used in ChDirPath, they can only include version ?

warrensbox: >Though, I can imagine any changes to these could be fairly catastrophic - tfswitch provides a lot of ways of passing in versions and you could almost guarantee that each of the methods is used by at least one person :D

Haha yes this has always been the case - “The Needs of the Many Outweigh the Needs of the Few” 🖖

yermulnik: > the TOML file is useful mostly for including the bin location.

Also `log-level` (which is not working properly when defined with TOML file at the moment, though hopefully this PR will get this fixed 🤞🏻) and potentially other features we may add in future.

> What is your suggestion?

The «_TOML only available in $HOME_» is my pref.
The «_if TOML is used in ChDirPath, they can only include version ?_» is what I can accept if others think this option makes sense.

MatthewJohn: One idea I did have is to completely ignore any TOML files in Chdir but instead to add a variable for TOML file location, which defaults to the homedir TOML path.
This would add the logical separation:
 * `Chdir` is to specify the location of your _code_ which tfswitch will check for versions
 * `TomlFile` is to simply change where the TOML file can is

We would only attempt to read the TOML file once (in the Tomlfile param path)

warrensbox: @yermulnik why can't the toml file in `ChDirPath` include `bin` as well?
I might be missing something here.

yermulnik: Actually I don't see a common use case for the TOML file to reside in `ChDirPath`. IMHO this would add complexity investigating issues and make debugging convoluted. I mean exactly non-`version` configuration parameters from TOML file.
If users need to override `bin` or `log-level` per use case, then they can (and should) use cmdline options. IMHO.
My vote goes for **explicit** (cmdline args) rather than **implicit** (TOML in `ChdriPath`).

Along with that what @MatthewJohn suggests above re the option to redefine TOML file location using cmdline option, absolutely makes sense to me and is a pretty common approach.

ps: I'm not pushing my opinion as an absolute one. If majority of us see TOML in `ChDirPath` as a good option, I'm fine with it. My only intention is to reduce failure and complexity points 🤝

lib/param_parsing/parameters.go
warrensbox: I would like tfswitch to fail if the configuration file is incorrect. This is because some of these switchings are done by robots or automation, and it's better if tfswitch fails in this case. We don't want to presume what the user wants if tfswitch is being run by a robot or automation.

lib/param_parsing/parameters.go
MatthewJohn: From the comment in #420 from Warren, that we should no longer support ChDirPath TOML file.. I've realised that the tests around the TOML file make heavy use of this - meaning that the only way we can test integration with the TOML file is to modify the TOML file in homedir - which isn't great for running tests.
Any suggestions for this? (I could leave the Toml directory argument in `getParamsTOML`, allowing the tests to test `getParamsTOML` passing in a test directory - but not sure this is fantastic

lib/param_parsing/parameters.go
MatthewJohn: Bringing this back up, from a discussion in #441, we actually ignore any terraform parsing failures and don't even tell the user - do we wish to continue doing this?
Couple of thoughts around this, based on #441 and my own personal experience:
 * I have seen uses where CI/CD pipelines use it and adhere to whatever the user has asked for. If they expect it to happen in terraform and they have a small syntax error (say during CI-part of CI/CD), do we want tfswitch to hang-around waiting for user-input?
 * If there's bad terraform, do we want to fail entirely - maybe there's a new syntax added that we don't support and the parsing fails? (though I guess this would become HCL3 vs HCL2 - but more importantly, are there syntax changes between HCL1 and HCL2 that are not valid and would this stop users of terraform <0.11 from being able to use tfswitch at all?!

lib/param_parsing/parameters.go
MatthewJohn: Any concerns with the changes? If not, I'll resolve the comment :)

lib/param_parsing/parameters.go
MatthewJohn: @warrensbox did you get a chance to ponder the above? Can this "conversation" be resolved? :)

MatthewJohn: or @yermulnik

lib/param_parsing/parameters.go
yermulnik: > do we want tfswitch to hang-around waiting for user-input?

For sure heavy "NO" if this is about CI.
We probably can check whether tty/vty is allocated and only come up with the interactive selection then and explicitly fail otherwise to force CI pipeline to fail too.

> If there's bad terraform, do we want to fail entirely

That's a hard question. I **personally* would prefer `tfswitch` to fail explicitly on bad terraform code (when it fails to load module). While others may prefer `tfswitch` to be less harsh I'd presume. And I'm leaning to think the latter is what ordinary users would prefer most.
Based on our experience from #441 I'd instead like `tfswitch` to log informational messages (hidden by default unless user increases verbosity) for when (and why) it ignored some specific source of TF version (this would allow us to know the reason for #441 w/o conducting "investigation").

yermulnik: > log informational messages (hidden by default unless user increases verbosity) for when (and why) it ignored some specific source of TF version

And, as you correctly noticed in the other thread, print explicit log message if the source is discarded because of the incorrect syntax whatsoever.

lib/param_parsing/parameters.go
yermulnik: @MatthewJohn Apologies my inattention, though I can't get what input you're expecting?
I did re-read you comment several times and I still am clueless on what you'd like to hear 🤷🏻

lib/param_parsing/parameters.go
MatthewJohn: I guess, finally, should this be covered in this PR, or should we re-open #441 and add way more logging around the fact?

MatthewJohn: Though, I will say, hanging CI/CD is _probably_ worse than it erroring saying "invalid terraform - use <version> arg to specify a version".

> We probably can check whether tty/vty is allocated 
I did start to look down this in golang, very briefly before posting and nothing jumped out immediately - I'll take another look

yermulnik: > I guess, finally, should this be covered in this PR

I'm not a dev and I prefer less friction, hence I choose this option, though I'd like to relay this decision to you so that's you to decide either to work on this in a standalone PR or in this current one.

> hough, I will say, hanging CI/CD is _probably_ worse than it erroring saying "invalid terraform - use arg to specify a version".

100% agree.

> I did start to look down this in golang, very briefly before posting and nothing jumped out immediately

Probably https://github.com/mattn/go-isatty ?

MatthewJohn: . Probably https://github.com/mattn/go-isatty or https://pkg.go.dev/golang.org/x/term#IsTerminal? (didn't compare them, though we need to ensure that would work on all the supported OS)

ooo - I was _clearly_ googling the wrong thing :facepalm: 

> 100% agree.

:+1: 


> I'm not a dev and I prefer less friction, hence I choose this option, though I'd like to relay this decision to you so that's you to decide either to work on this in a standalone PR or in this current one.

I'm just aware of the amount of changes that go in - if something breaks post-release, do we roll forward or back (and remove a load of other stuff). There's also other bug reports that are pending the PR.
Historically, at least in my small amount of OSS that people actually use :P (so ignore most of it ;) ), I do like smaller things - release fast (release per issue) etc. But I'm open :)  I'm very much aware of @warrensbox bringing us into his project and don't want to dictate things too much (so I'm always up for his suggestions/opinions)

yermulnik: If that's a big portion of work, then defo go with standalone PR.

MatthewJohn: Maybe not a big portion of work.. but should be considered and carefully tested and I'd be worried about both of these if it's "tacked on" to push through a PR, if you see what I mean :)

MatthewJohn: On that note, I do fancy a release branch, if we're aiming to release several PRs in one release.
We can get the PRs merged into the branch, so they're all in one place and line up any smaller changes that we can get in, without a big PR to to be "reviewed" (in the literal line-by-line idea), because they've already been reviewed in individual PRs

yermulnik: Given existence of any kind of hesitation/worrisome bits/loveless phase of moon 🤪 , defo go with standalone PR =)
I worded that weirdly and effectively I meant if it's the "2-4 lines of code" when I said I prefer this to be part of the current PR 😂

MatthewJohn: > loveless phase of moon
:laughing: :laughing: :laughing: :laughing:

lib/param_parsing/parameters.go
MatthewJohn: Ah sorry - well, if you're happy with how it's currently implemented, then all good :)

lib/param_parsing/parameters.go
MatthewJohn: I've added a warning when we fail to parse Terraform in b2a5916

example:
```
➜  tmp git:(420-how-to-deal-with-toml-file) ✗ go run ../
16:41:37.128 WARN Error whilst parsing Terraform: "Unclosed configuration block: There is no closing brace for this block before the end of the file. This may be caused by incorrect brace nesting elsewhere in this file."  
? Select Terraform version: 
  ▸ 1.8.4 *recent
    1.3.6 *recent

```

lib/param_parsing/parameters.go
yermulnik: If it works as intended an as expected, then I'm all good with it unless you think it has to be re-factored or re-worked 😄

lib/param_parsing/parameters.go
MatthewJohn: Have created https://github.com/warrensbox/terraform-switcher/issues/451 for the TTY detection etc.

lib/param_parsing/versiontf.go
```diff
@@ -64,5 +64,8 @@ func GetVersionFromVersionsTF(params Params) (Params, error) {
 
 func isTerraformModule(params Params) bool {
 	module, err := tfconfig.LoadModule(params.ChDirPath)
+	if err != nil {
+		logger.Warnf("Error whilst parsing Terraform: %q", err)
```
yermulnik, line 68: Consistency nit:
```suggestion
		logger.Warnf("Error parsing Terraform module: %v", err)
```

.github/workflows/build.yml
```diff
@@ -68,6 +68,12 @@ jobs:
         go build -v -o build/tfswitch
         mkdir `pwd`/bin/
         find ./test-data/integration-tests/* -type d -print0 | while read -r -d $'\0' TEST_PATH; do
+            if test -f "${TEST_PATH}/.tfswitch.toml"
+            then
+                cp "${TEST_PATH}/.tfswitch.toml" ~/
+            else
+                { test -f ~/.tfswitch.toml && rm ~/.tfswitch.toml; } || true
+            fi
```
yermulnik, line 76: What this if-conditional goal is?
If test TOML exists, then copy it to HOME, otherwise delete TOML from HOME 😲

lib/param_parsing/parameters.go
```diff
@@ -56,23 +61,44 @@ func GetParameters() Params {
...
+			logger.Fatalf("Failed to obtain settings from Terraform module: %v", err)
+		}
 	}
-	if err != nil {
-		logger.Fatalf("Error parsing configuration file: %q", err)
+
+	if terraGruntFileExists(params) {
+		params, err = GetVersionFromTerragrunt(params)
+		if err != nil {
+			logger.Fatalf("Failed to obtain settings from Terragrunt configuration: %v", err)
```
yermulnik, line 96: Instead of hardcoding and wording version sources, could we probably move those from `lib/param_parsing/*.go` files to a more global place and re-use?

.github/workflows/build.yml
MatthewJohn: Exactly that - it's handled the fact that a test has a TOML file, but we ignore ChDiR (CWD) now, but I wanted to keep the test... so copied to HOMEDIR so it continued to work

yermulnik: Ah, I lean to think I got it =)
You should be good to replace `{ test -f ~/.tfswitch.toml && rm ~/.tfswitch.toml; } || true` with a simpler `rm -f ~/.tfswitch.toml`

PR #432, opened 2024-05-22
docs: Fix links to `install.sh`

### LINKED ISSUE #431: Install script issue
Called install.sh script using following command by passing version tag but it fails to execute entire script. fails somewhere in execute function. Unable to figure it out exact which line it is failing (Version i am passing is 1.0.2)

`curl -L https://raw.githubusercontent.com/warrensbox/terraform-switcher/release/install.sh | bash -s -- $VERSION`

image

Apparently it is failing somewhere in execute function but there is no way to debug it.

yermulnik:
Fix links to `install.sh` after moving from `release` to `master` branch as main branch

fixes #431

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ See the [release page](https://github.com/warrensbox/terraform-switcher/releases
 
 For macOS or various Linux distributions, Homebrew offers the simplest installation process. <a href="https://brew.sh/" target="_blank">If you do not have homebrew installed, click here</a>.
 
-```ruby
+```shell
 brew install warrensbox/tap/tfswitch
 ```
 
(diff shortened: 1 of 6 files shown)


MatthewJohn:
I also noted a day ago, whilst trying to use the installer, that the installer doesn't work for any releases prior to v1.1.0 (so only supports the latest 2 versions) - or is that due to the old version of the script being used by the instructions, which this PR fixes? :)

MatthewJohn:
> I also noted a day ago, whilst trying to use the installer, that the installer doesn't work for any releases prior to v1.1.0 (so only supports the latest 2 versions) - or is that due to the old version of the script being used by the instructions, which this PR fixes? :)

Ah, it does - ignore me :)

MatthewJohn (approved):
Looks good :)

yermulnik:
JFYI: pushed WWW docs to https://tfswitch.warrensbox.com/ manually

Issue #414, opened 2024-04-26
Terraform `1.8.2` ships additional file within tarball


MatrixCrawler:
### Discussed in https://github.com/warrensbox/terraform-switcher/discussions/411

<div type='discussions-op-text'>

<sup>Originally posted by **yermulnik** April 26, 2024</sup>
Should we be taking care of this (e.g. to not vomit `.terraform.versions` dir with arbitrary file)?

https://github.com/hashicorp/terraform/releases/tag/v1.8.2
NOTE:

Starting with this release, we are including a copy of our license file in all packaged versions of our releases, such as the release .zip files. If you are consuming these files directly and would prefer to extract the one terraform file instead of extracting everything, you need to add an extra argument specifying the file to extract, like this:

```shell
unzip terraform_1.8.2_linux_amd64.zip terraform
```

We should make sure, that only the terraform binary is extracted from the archive.

PR #433, opened 2024-05-24
fix: Only extract terraform file from Terraform release archive

MatthewJohn:
Update tests to ensure only terraform file is extracted Add 'terraform' file, along with 'README' and 'LICENSE' fiel to test-data.zip

Issue #414


MatrixCrawler (approved):
LGTM

Issue #315, opened 2023-08-29
OpenTofu Support


cameronattard:
**Is your feature request related to a problem? Please describe.**
Per https://opentf.org/announcement, a new fork of Terraform, OpenTF is being created.

**Describe the solution you'd like**
Allow using `tfswitch` to install and manage OpenTF versions in addition to Terraform versions.

nitrocode:
Wouldn't something like the `--mirror` flag work for opentf ?

https://github.com/warrensbox/terraform-switcher#use-custom-mirror

roni-frantchi:
> Wouldn't something like the --mirror flag work for opentf ?

Hey @nitrocode - not quite.  
Although we are looking to be fully compatible with legacy Terraform, the binary name, command (`terraform` vs `opentf`) are all going to be different.

nitrocode:
Same issue that affects listing versions in mirrors will also affect opentofu since there isn't a single non-paginated url containing every opentofu release.

Related issue
- https://github.com/warrensbox/terraform-switcher/issues/311

nitrocode:
An option might be to switch from downloading terraform releases from the hashicorp page and download instead directly from github. Then it would be fairly simple to download the files from opentofu instead. The version check would then have to hit the github rest api or have an index url file passed in as an argument.

https://github.com/opentofu/opentofu/releases
https://github.com/hashicorp/terraform/releases

nitrocode:
I made a page here that seems to line up similar to the terraform release page

https://nitrocode.github.io/opentofu/

nitrocode:
We'll probably need a way of overriding the `versionPrefix` since opentofu's prefix is `tofu_`

It might also be nice to override the rest of these too.

https://github.com/warrensbox/terraform-switcher/blob/d7dfd1b44605b095937e94b981d24305b858ff8c/main.go#L40-L49

MatthewJohn:
I'm interested in this issue, as we use tfswitch (or terraform-switcher ;) ) in Terrareg.

Once concern I could imagine is that, you're using Hashicorp's HCL library for obtaining the constraint from a `terraform {}` block: https://github.com/warrensbox/terraform-switcher/blob/d7dfd1b44605b095937e94b981d24305b858ff8c/main.go#L433

[...]

PR #435, opened 2024-05-26
OpenTofu support and JSON recentFile support

MatthewJohn:
Issue #315

https://github.com/warrensbox/terraform-switcher/discussions/434


lib/symlink.go
```diff
@@ -89,9 +89,9 @@ func CheckSymlink(symlinkPath string) bool {
 }
 
 // ChangeSymlink : move symlink to existing binary
-func ChangeSymlink(binVersionPath string, binPath string) {
+func ChangeSymlink(product Product, binVersionPath string, binPath string) {
```
MatthewJohn, line 92: This breaks a public method

lib/files.go
```diff
@@ -49,7 +45,7 @@ func CheckFileExist(file string) bool {
 
 // Unzip will decompress a zip archive, moving all files and folders
 // within the zip file (parameter 1) to an output directory (parameter 2).
-func Unzip(src string, dest string) ([]string, error) {
+func Unzip(src string, dest string, fileToUnzip string) ([]string, error) {
```
MatthewJohn, line 48: This breaks a public method *and* it's being used!
https://github.com/hazelops/ize/blob/97d7c0e48f116054b7493dfcaa15bd0460bae016/internal/terraform/local.go#L274

lib/products.go
```diff
@@ -0,0 +1,151 @@
...
+			PublicKeyId:    "72D7468F",
+			PublicKeyUrl:   "https://www.hashicorp.com/.well-known/pgp-key.txt",
+		},
+	},
+	OpenTofuProduct{
+		ProductDetails{
+			ID:                    "opentofu",
+			Name:                  "OpenTofu",
+			DefaultMirror:         "https://get.opentofu.org/tofu",
+			DefaultDownloadMirror: "https://github.com/opentofu/opentofu/releases/download",
```
MatthewJohn, line 122: This cannot currently be overriden.
If we do allow it, it's not (currently) used by Terraform.
But, if it is, then the "default download mirror" URL would need to default to releases.hashicorp.com for Terraform and Github got OpenTofu.
However, since they have different layouts:
 * terraform/1.0.0/terraform_1.0x
 * tags/v1.0.0/tofu/
We'd need to make it clear how someone would create their own mirror if they wish to host them!

lib/list_versions.go
```diff
@@ -14,15 +14,7 @@ type tfVersionList struct {
...
-// getTFList :  Get the list of available terraform version given the hashicorp url
-func getTFList(mirrorURL string, preRelease bool) ([]string, error) {
-	logger.Debugf("Get list of terraform versions")
-	result, err := getTFURLBody(mirrorURL)
-	if err != nil {
-		return nil, err
-	}
-
-	var tfVersionList tfVersionList
+func getVersionsFromBody(body string, preRelease bool, tfVersionList *tfVersionList) {
```
yermulnik, line 17: Just a wild suggestion: would it make sense to employ a HTML parsing lib (e.g. `net/html` or similar) to parse HTML? 🤔 
Might probably be a scope for standalone PR though.

lib/param_parsing/parameters.go
```diff
@@ -43,10 +54,12 @@ func GetParameters() Params {
...
 	getopt.StringVarLong(&params.LogLevel, "log-level", 'g', "Set loglevel for tfswitch. One of (INFO, NOTICE, DEBUG, TRACE)")
-	getopt.StringVarLong(&params.MirrorURL, "mirror", 'm', "install from a remote API other than the default. Default: "+lib.DefaultMirror)
+
+	getopt.StringVarLong(&params.MirrorURL, "mirror", 'm', "install from a remote API other than the default. Default (based on product):\n"+strings.Join(defaultMirrors, "\n"))
 	getopt.BoolVarLong(&params.ShowLatestFlag, "show-latest", 'U', "Show latest stable version")
 	getopt.StringVarLong(&params.ShowLatestPre, "show-latest-pre", 'P', "Show latest pre-release implicit version. Ex: tfswitch --show-latest-pre 0.13 prints 0.13.0-rc1 (latest)")
 	getopt.StringVarLong(&params.ShowLatestStable, "show-latest-stable", 'S', "Show latest implicit version. Ex: tfswitch --show-latest-stable 0.13 prints 0.13.7 (latest)")
+	getopt.StringVarLong(&params.Product, "product", 'q', fmt.Sprintf("Specifies which product to use. Ex: `tfswitch --product terraform` will install Terraform. Options: (%s)", strings.Join(productIds, ", ")))
```
yermulnik, line 62: Assuming Terraform is what we install by default (when this arg is not provided), I'd suggest to 1) mention this defaults to Terraform and 2) provide example with OpenTofu instead.

Also while I'm amid the review process, I'll just leave this question here (apologies if I'll find answer down the changeset from this PR): is it possible to specific default Product using TOML file? I guess this would be handy.

lib/param_parsing/parameters.go
```diff
@@ -73,6 +86,14 @@ func GetParameters() Params {
 		logger.Fatalf("Error parsing configuration file: %q", err)
 	}
 
+	// Set defaults based on product
+	product := lib.GetProductById(params.Product)
+	if product == nil {
+		logger.Fatalf("Invalid product: " + params.Product)
```
yermulnik, line 92:
```suggestion
		logger.Fatalf("Invalid \"product\" configuration value: %q", params.Product)
```

lib/param_parsing/parameters.go
```diff
@@ -43,10 +54,12 @@ func GetParameters() Params {
 	getopt.StringVarLong(&params.LatestStable, "latest-stable", 's', "Latest implicit version based on a constraint. Ex: tfswitch --latest-stable 0.13.0 downloads 0.13.7 and 0.13 downloads 0.15.5 (latest)")
 	getopt.BoolVarLong(&params.ListAllFlag, "list-all", 'l', "List all versions of terraform - including beta and rc")
 	getopt.StringVarLong(&params.LogLevel, "log-level", 'g', "Set loglevel for tfswitch. One of (INFO, NOTICE, DEBUG, TRACE)")
-	getopt.StringVarLong(&params.MirrorURL, "mirror", 'm', "install from a remote API other than the default. Default: "+lib.DefaultMirror)
+
```
yermulnik, line 57: nit: why empty line?

lib/products.go
```diff
@@ -0,0 +1,151 @@
...
+	return p.VersionPrefix
+}
+func (p TerraformProduct) GetExecutableName() string {
+	return p.ExecutableName
+}
+func (p TerraformProduct) GetArchivePrefix() string {
+	return p.ArchivePrefix
+}
+func (p TerraformProduct) GetArtifactUrl(mirrorURL string, version string) string {
+	return fmt.Sprintf("%s%s", mirrorURL, version)
```
yermulnik, line 59: This assumes `mirrorURL` is always suffixed with `/`, right?
Would it make sense to 1) trim trailing slash (if any) and 2) use `%s/%s` instead to ensure we glue URL parts together correctly? 🤔

lib/products.go
```diff
@@ -0,0 +1,151 @@
...
+	TerraformProduct{
+		ProductDetails{
+			ID:             "terraform",
+			Name:           "Terraform",
+			DefaultMirror:  "https://releases.hashicorp.com/terraform",
+			VersionPrefix:  "terraform_",
+			ExecutableName: "terraform",
+			ArchivePrefix:  "terraform_",
+			PublicKeyId:    "72D7468F",
+			PublicKeyUrl:   "https://www.hashicorp.com/.well-known/pgp-key.txt",
```
yermulnik, line 114: Thoughts aloud: would really love this two to be overridable by cmdline args. Or maybe an option to turn off GPG verification. For use cases where user wants to use their own mirror with repacked archives.

lib/products.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func GetAllProducts() []Product {
+	return products
+}
+
+func getLegacyProduct() Product {
+	product := GetProductById(legacyProductId)
+	if product == nil {
+		logger.Fatalf("Default product could not be found")
```
yermulnik, line 148: Trailing `f` in `logger.Fatalf` seems to be not needed.

lib/symlink.go
yermulnik: As I already commented above, maybe have a fallback for Product to provide backward compatibility? 🤔

main.go
```diff
@@ -31,6 +31,13 @@ var version string
 
 func main() {
 
+	product := lib.GetProductById(parameters.Product)
+	if product == nil {
+		fmt.Printf("Invalid product specified")
```
yermulnik, line 36: Trailing `f` in `fmt.Printf` seems redundant.
Also maybe print out available values right away to ease user's experience?

main.go
```diff
@@ -31,6 +31,13 @@ var version string
...
+		lib.UsageMessage()
+		os.Exit(0)
```
yermulnik, line 38: It's probably a matter of personal preference, though I think it is odd to exit successfully on invalid user input.

lib/download.go
```diff
@@ -12,14 +12,17 @@ import (
 
 // DownloadFromURL : Downloads the terraform binary and its hash from the source url
 func DownloadFromURL(installLocation, mirrorURL, tfversion, versionPrefix, goos, goarch string) (string, error) {
+	product := getLegacyProduct()
```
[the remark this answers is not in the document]
MatthewJohn: I was going to say the following, but your point about tfswitch being for terraform is certainly changing how I saw it :D 

My original thinking process was that the "legacy" aspect was purely for the external methods that I was trying to deprecate and replace with those that take a product.
If (assuming we're generally happy about the new methods) wanted to keep both version around for the long-term, then I agree.. but this is where "legacy" came from.
The "main" execution path (at least, from the cmd PoV) doesn't use this and uses the GetProductById methods

lib/list_versions.go
```diff
@@ -33,13 +25,27 @@ func getTFList(mirrorURL string, preRelease bool) ([]string, error) {
...
 	}
+	for _, match := range matches {
+		trimstr := strings.Trim(match, "/\"") //remove '/' or '"' from /X.X.X/" or /X.X.X"
+		tfVersionList.tflist = append(tfVersionList.tflist, trimstr)
+	}
+}
+
+// getTFList :  Get the list of available terraform version given the hashicorp url
+func getTFList(mirrorURL string, preRelease bool) ([]string, error) {
+	logger.Debugf("Get list of terraform versions")
```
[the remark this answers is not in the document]
MatthewJohn: hmm, I guess "terraform" should be replaced by the product name - seems like a waste to pass the product object in just for the name (at least it's an internal function) :P

lib/param_parsing/parameters.go
MatthewJohn: > Assuming Terraform is what we install by default (when this arg is not provided), I'd suggest to 1) mention this defaults to Terraform and 2) provide example with OpenTofu instead.

Good idea - I'll do that

> Also while I'm amid the review process, I'll just leave this question here (apologies if I'll find answer down the changeset from this PR): is it possible to specific default Product using TOML file? I guess this would be handy.

Also a very good idea - I'll do this as well :)

lib/param_parsing/parameters.go
MatthewJohn: Ah, this was a refactoring thing - mirror URL had a bunch of the logic that has since been moved to L39 and I'd forgotten to clear it up

lib/products.go
MatthewJohn: Without looking (I'll take another look at the code base tomorrow morning), I saw we always prepend the trailing slash - I had at some point had the seperator here, which broke the requests because of double slashes.
I wasn't keen on it at the time, so maybe it's better to check when the URL is being used.
I'll look at cleaning this up

lib/products.go
MatthewJohn: Probably both would be good, yeh - and probably TOML support :D

lib/list_versions.go
MatthewJohn: From the discussion in #434, I'd assume (hope) that maybe we'd look at moving to the JSON endpoints for Terraform/OpenTofu (if we create a PR there) and wouldn't have a need for it.
I guess using a HML parser would also mean knowing the exact layout of the page and traversing it accordingly - as brittle as using regex on HTML to obtain the versions is, I would imagine HTML parsing it _could_ be worse in this sense?

lib/download.go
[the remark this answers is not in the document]
yermulnik: Yea, your point absolutely makes sense too =)
Given the «_was trying to deprecate and replace with those that take a product_» statement, I'd say just drop a comment to let readers know this bit is targeted to be decommissioned/replaced sooner or later and hence it's "legacy".

For me it's like "if your peer don't get it, someone else may miss the point to".

lib/list_versions.go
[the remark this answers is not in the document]
yermulnik: > I guess "terraform" should be replaced by the product name

I wasn't that thorough to understand whether this function is TF-specific or is used for the other Product to and leant to think it was the former 🤷🏻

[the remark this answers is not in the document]
yermulnik: > a waste to pass the product object in just for the name (at least it's an internal function)

This is something that user can face and I'd make it as much clear as possible for ordinary people.
Re the "waste" — is there a noticeable speed or memory penalty on passing the object?

lib/param_parsing/parameters.go
yermulnik: Apologies for the typo, I obviously meant `specify` rather than `specific`. Though this couldn't make you misinterpret for sure =) Thanks.

lib/products.go
yermulnik: Thanks.
This is actually what I try to implement always for this kind of _base_ URL values: trim trailing slash if any and append slash unconditionally then — this way I'm always sure there's no doubleslash at this place.

lib/products.go
yermulnik: And, yep, the scope of another issue/PR obviously =)

lib/list_versions.go
yermulnik: Your point makes sense to me. In spite of moving to JSON there's no point in re-factoring to HTML parsing using lib 🤝 
As for the parsing using regex vs parsing using purpose-built lib, I'd say why don't we parse JSON using regexp then? (it is possible. I tried it and it worked) =))

lib/files.go
MatthewJohn: I'm afraid I'm quite new to golang.. though I've dealt with default argument values before - I can't quite remember how (I had a feeling it was using a (without knowing a better word for it) expansion technique (`(params ...string)`).
If this is the "best" way, I guess it would need to do something like:

```
func Unzip(src string, dest, string, fileToUnzipSlice ...string) {
  fileToUnzip := "terraform"
  if len(fileToUnzipSlice) > 0 {
    fileToUnzip = fileToUnzipSlice[0]
  }
}
```
?

lib/list_versions.go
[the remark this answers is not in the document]
MatthewJohn: I'm tempted to suggest:
```
logger.Debug("Getting list of versions")
```
as this method is used by quite a few more functions and will affect even more public functions, which would need signature changes.
What do you think?

main.go
MatthewJohn: Yes, I wasn't quite sure how to handle this...
`GetParameters` already obtains the product and ensures it's not null. Ideally this should be done once - but I'm not sure if GetParameters or main should do do it

main.go
MatthewJohn: This is mostly why I used UsageMessage afterwards. Though, this functionality should _never_ be called (at least at the moment) because GetParameters already obtains the product and verifies it.(see https://github.com/warrensbox/terraform-switcher/pull/435#discussion_r1615273953).

But in this case, do you think saying `Invalid product specified by "-q"` and printing out usage would be good enough, since the usage provides the available parameters?
I wonder if there's a way to print usage, but mark as an argument as having an invalid value, so that the usage tells the user this.

lib/param_parsing/parameters.go
MatthewJohn: > Assuming Terraform is what we install by default (when this arg is not provided), I'd suggest to 1) mention this defaults to Terraform and 2) provide example with OpenTofu instead.

This is now done

> Also while I'm amid the review process, I'll just leave this question here (apologies if I'll find answer down the changeset from this PR): is it possible to specific default Product using TOML file? I guess this would be handy.

TODO

lib/param_parsing/parameters.go
```diff
@@ -43,10 +54,11 @@ func GetParameters() Params {
...
 	getopt.StringVarLong(&params.LogLevel, "log-level", 'g', "Set loglevel for tfswitch. One of (INFO, NOTICE, DEBUG, TRACE)")
-	getopt.StringVarLong(&params.MirrorURL, "mirror", 'm', "install from a remote API other than the default. Default: "+lib.DefaultMirror)
+	getopt.StringVarLong(&params.MirrorURL, "mirror", 'm', "install from a remote API other than the default. Default (based on product):\n"+strings.Join(defaultMirrors, "\n"))
 	getopt.BoolVarLong(&params.ShowLatestFlag, "show-latest", 'U', "Show latest stable version")
 	getopt.StringVarLong(&params.ShowLatestPre, "show-latest-pre", 'P', "Show latest pre-release implicit version. Ex: tfswitch --show-latest-pre 0.13 prints 0.13.0-rc1 (latest)")
 	getopt.StringVarLong(&params.ShowLatestStable, "show-latest-stable", 'S', "Show latest implicit version. Ex: tfswitch --show-latest-stable 0.13 prints 0.13.7 (latest)")
+	getopt.StringVarLong(&params.Product, "product", 'q', fmt.Sprintf("Specifies which product to use. Ex: `tfswitch --product opentofu` will install Terraform. Options: (%s). Default: %s", strings.Join(productIds, ", "), lib.DefaultProductId))
```
MatthewJohn, line 61: The reason I ended up with `-q` is that that I tried to use `-p` and `-P`, but these are already taken!
Happy for any suggestions as to a good character :) I'm also wary that `-q` is common for `--quiet`

lib/files.go
yermulnik: > I'm afraid I'm quite new to golang..

Whereas I'm a way not a Golang guy at all 🤣 But, yeah, using ellipsis looks easiest way.
I'd only say let's check the number of args to prevent consumers passing improper count of args by an accident or a human mistake. Plus an explanatory comment. Like this:
```go
// Make 3rd arg optional for backward compatibility
func Unzip(src string, dest string, fileToUnzipSlice ...string) {
  fileToUnzip := "terraform"
  if len(fileToUnzipSlice) > 1 {
    log.Fatal("Too many args")
  } else if len(fileToUnzipSlice) == 1 {
    fileToUnzip = fileToUnzipSlice[0]
  }
}
```

yermulnik: I also assume `fileToUnzip := "terraform"` would refer not `terraform` verbatim, but would be a ref to a var that holds default product's executable filename.

lib/list_versions.go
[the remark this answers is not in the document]
yermulnik: Works for me 👍🏻 Also probably adjust the comment to this function respectively. Thanks.

main.go
yermulnik: > guess, ideally, `GetParameters` would store the Product that it obtains and main wouldn't need to check

+1 👍🏻

main.go
yermulnik: > But in this case, do you think saying `Invalid product specified by "-q"` and printing out usage would be good enough, since the usage provides the available parameters?

Yep, agree, should be sufficient to simplify code (and is a common approach I guess).

lib/param_parsing/parameters.go
yermulnik: Ah, good shout re the `--quiet`. That indeed makes `-q` looking odd.
What about `-t`? Like the first char of `tool`, but also as a last char of `--product` longopt =)

lib/files.go
MatthewJohn: Done :)

lib/download.go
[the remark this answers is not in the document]
MatthewJohn: Have added some comments around the "legacy" methods - hope this looks okay :)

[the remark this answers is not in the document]
yermulnik: Absolutely. Appreciate this ❤️

lib/install.go
```diff
@@ -260,13 +267,13 @@ func installableBinLocation(userBinPath string) string {
 			homeBinExist := CheckDirExist(filepath.Join(homedir, "bin")) //check to see if ~/bin exist
 			if homeBinExist {                                            //if ~/bin exist, install at ~/bin/terraform
 				logger.Infof("Installing terraform at %q", filepath.Join(homedir, "bin"))
-				return filepath.Join(homedir, "bin", "terraform")
+				return filepath.Join(homedir, "bin", product.GetExecutableName())
```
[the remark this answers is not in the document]
MatthewJohn: Fixed :)

lib/files.go
```diff
@@ -125,6 +132,7 @@ func WriteLines(lines []string, path string) (err error) {
 }
 
 // ReadLines : Read a whole file into the memory and store it as array of lines
+// @TODO This isn't used - should we keep it?
```
MatthewJohn, line 135: From https://github.com/warrensbox/terraform-switcher/pull/436#discussion_r1615333045

Must be considered in review - either remove the TODO or remove the method - same with WriteLines

MatthewJohn: From @yermulnik 

As we've learned there are loads (or at least some 🤪) of the projects that use `tfswitch` as a lib or re-use its code as-is, we may probably need to announce removal of functions/methods/whatever in advance and give consumers time to at least next release to migrate/replace.
So what I'd suggest is to leave this as is and get this removed any time later after any major release and in case any of use stumble upon it by an accident or on purpose (and hence maybe rephrase it like `... - should remove anytime after 2.0.0`). IMHO.

lib/recent.go
```diff
@@ -0,0 +1,100 @@
...
+	// Check for requestedVersion in versions list
+	for versionIndex, versionVal := range versions {
+		if versionVal == requestedVersion {
+			versions = append(versions[:versionIndex], versions[versionIndex+1:]...)
+		}
+	}
+
+	// Add new version to start of slice
+	versions = append([]string{requestedVersion}, versions...)
+	if len(versions) > 3 {
```
MatthewJohn, line 25: From https://github.com/warrensbox/terraform-switcher/pull/436#discussion_r1615278360
@yermulnik:

This pushed me to think that we may allow user to customize number of recent versions we keep for them. WDYT?

lib/param_parsing/parameters.go
MatthewJohn: Configuration in TOML file has been added in 147fd939ebbf3d7890128d997e0a1e19f98d0a93

main.go
MatthewJohn: I have added the product entity to Params in 67433dbaf7d6fb94c7f51486edb0a2b0cf1f659e and removed this code ;)

lib/param_parsing/parameters.go
MatthewJohn: Have updated in 9ab0409af79194d714ea65523e2351aa5c6dc0a6

lib/symlink.go
MatthewJohn: Since this is not a simple string argument and is taking the "product" (similar to a lot of other public methods that we've created deprecated stubs for) - I wonder if it's better to do the same here.
I'll do so, post the commit and see what you think :)

yermulnik: > I've modified in [d22a046](https://github.com/warrensbox/terraform-switcher/commit/d22a046234da91c69f6acdc4d9a5bcd2a8f530cc) - what you think :)

The "trick" looks good =) Maybe also add to comment that this method is still there for backward compatibility as it's public one? 🤔

MatthewJohn:
I've started writing some tests for DownloadProductFromUrl and faced some issues.
Have raised: https://github.com/warrensbox/terraform-switcher/issues/439
I wonder if these tests can immediately start using Proton Mail's library to see if it resolves the issues

lib/files.go
warrensbox: Yeah maybe a depreciation note. 
Perhaps Release 2.0

yermulnik:
> I wonder if these tests can immediately start using Proton Mail's library to see if it resolves the issues

I'd say go ahead with it.
@warrensbox ?

lib/products.go
warrensbox: There is a --mirror option with the current version.
Please double check to see if the `--mirror` is still usable after this upgrade

lib/symlink.go
MatthewJohn: Different to:
```
// This function has been deprecated in favor of ChangeProductSymlink
```
?

MatthewJohn:
> > I wonder if these tests can immediately start using Proton Mail's library to see if it resolves the issues
> 
> I'd say go ahead with it. @warrensbox ?

I moved away, only for the purpose of writing some end-to-end tests that mock external services and it works :)

warrensbox:
> > > I wonder if these tests can immediately start using Proton Mail's library to see if it resolves the issues
> > 
> > 
> > I'd say go ahead with it. @warrensbox ?
> 
> I moved away, only for the purpose of writing some end-to-end tests that mock external services and it works :)

Yeah go for it. It's better to stay up-to-date! Thanks @MatthewJohn

lib/download.go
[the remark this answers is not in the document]
warrensbox: Yeah, we may have people still using these functions. It's good to leave it there for now

lib/files.go
```diff
@@ -49,9 +45,20 @@ func CheckFileExist(file string) bool {
 
 // Unzip will decompress a zip archive, moving all files and folders
 // within the zip file (parameter 1) to an output directory (parameter 2).
-func Unzip(src string, dest string) ([]string, error) {
+// fileToUnzip (parameter 3) specifices the file within the zipfile to be extracted.
```
warrensbox, line 48: Minor spelling error:
specifices = specifies

lib/param_parsing/parameters.go
```diff
@@ -56,23 +69,67 @@ func GetParameters() Params {
...
+	// TOML from Homedir
+	if tomlFileExists(lib.GetHomeDirectory()) {
+		params, err = getParamsTOML(params, lib.GetHomeDirectory())
+		if err != nil {
+			logger.Fatalf("Failed to obtain settings from TOML config in home directory: %v", err)
+		}
+	}
+
+	// TOML from ChDirPath
+	if tomlFileExists(params.ChDirPath) {
```
warrensbox, line 81: I believe this is also your work at:

https://github.com/warrensbox/terraform-switcher/pull/429

For now, we can exclude: `TOML from ChDirPath` .
See: https://github.com/warrensbox/terraform-switcher/discussions/420#discussioncomment-9624774

lib/param_parsing/toml.go
```diff
@@ -1,23 +1,24 @@
...
-	"path/filepath"
 )
 
 const tfSwitchTOMLFileName = ".tfswitch.toml"
 
 // getParamsTOML parses everything in the toml file, return required version and bin path
-func getParamsTOML(params Params) (Params, error) {
-	tomlPath := filepath.Join(params.ChDirPath, tfSwitchTOMLFileName)
-	if tomlFileExists(params) {
+func getParamsTOML(params Params, baseDir string) (Params, error) {
```
warrensbox, line 13: I still think this can be better?

`getParamsTOML(params Params, baseDir string)`

for:

params.ChDirPath = lib.GetHomeDirectory() //set path to home
params, err = getParamsTOML(params, lib.GetHomeDirectory())

that way we only need `getParamsTOML(params)`?

lib/recent.go
warrensbox: Instead of letting users customize, lets leave it at 5 for now (increase from 3). I don't think we should over-engineer this. Unless multiple people opens a PR for this.

lib/param_parsing/parameters.go
MatthewJohn: Yep! :) Because of the way the product selection was playing with the GetParameters (and I assumed the TOML file changes would most likely beat these ones), I merged that branch into here for simplicity

lib/param_parsing/toml.go
MatthewJohn: If we don't respect the ChDirPath that the user passes, wouldn't it it be odd to manipulate the variable to use it for something that it's not intended for?

lib/files.go
MatthewJohn: Fixed in 6766145

lib/products.go
MatthewJohn: @warrensbox It is not currently usable for the actual download source of OpenTofu - it will currently always use Github

lib/products.go
MatthewJohn: This is now fixed :)

lib/files.go
MatthewJohn: Done :)

lib/recent.go
MatthewJohn: This has been replaced by the work in #437

lib/install.go
```diff
@@ -74,7 +74,7 @@ func install(tfversion string, binPath string, installPath string, mirrorURL str
 	}
 
 	/* check if selected version already downloaded */
-	installFileVersionPath := ConvertExecutableExt(filepath.Join(installLocation, VersionPrefix+tfversion))
+	installFileVersionPath := ConvertExecutableExt(filepath.Join(installLocation, product.GetVersionPrefix()+tfversion))
```
yermulnik, line 77: readability nit:
```suggestion
	installFileVersionPath := ConvertExecutableExt(filepath.Join(installLocation, product.GetVersionPrefix() + tfversion))
```

lib/install.go
```diff
@@ -191,42 +188,66 @@ func installableBinLocation(userBinPath string) string {
...
+func InstallProductVersion(product Product, dryRun bool, version, customBinaryPath, installPath, mirrorURL string) {
 	logger.Debugf("Install version %s. Dry run: %s", version, strconv.FormatBool(dryRun))
 	if !dryRun {
 		if validVersionFormat(version) {
 			requestedVersion := version
 
 			//check to see if the requested version has been downloaded before
 			installLocation := GetInstallLocation(installPath)
-			installFileVersionPath := ConvertExecutableExt(filepath.Join(installLocation, VersionPrefix+requestedVersion))
+			installFileVersionPath := ConvertExecutableExt(filepath.Join(installLocation, product.GetVersionPrefix()+requestedVersion))
```
yermulnik, line 245:
```suggestion
			installFileVersionPath := ConvertExecutableExt(filepath.Join(installLocation, product.GetVersionPrefix() + requestedVersion))
```

lib/param_parsing/environment.go
```diff
@@ -3,6 +3,11 @@ package param_parsing
 import "os"
 
 func GetParamsFromEnvironment(params Params) Params {
-	params.Version = os.Getenv("TF_VERSION")
+	if envVersion := os.Getenv("TF_VERSION"); envVersion != "" {
+		params.Version = envVersion
+	}
+	if envProduct := os.Getenv("TF_PRODUCT"); envProduct != "" {
```
yermulnik, line 9: Should the option to use Env var be reflected in documentation?
Ref: https://tfswitch.warrensbox.com/usage/commandline/#use-environment-variables

lib/param_parsing/parameters.go
```diff
@@ -43,10 +60,11 @@ func GetParameters() Params {
 	getopt.StringVarLong(&params.LatestStable, "latest-stable", 's', "Latest implicit version based on a constraint. Ex: tfswitch --latest-stable 0.13.0 downloads 0.13.7 and 0.13 downloads 0.15.5 (latest)")
 	getopt.BoolVarLong(&params.ListAllFlag, "list-all", 'l', "List all versions of terraform - including beta and rc")
 	getopt.StringVarLong(&params.LogLevel, "log-level", 'g', "Set loglevel for tfswitch. One of (ERROR, INFO, NOTICE, DEBUG, TRACE)")
-	getopt.StringVarLong(&params.MirrorURL, "mirror", 'm', "install from a remote API other than the default. Default: "+lib.DefaultMirror)
+	getopt.StringVarLong(&params.MirrorURL, "mirror", 'm', "install from a remote API other than the default. Default (based on product):\n"+strings.Join(defaultMirrors, "\n"))
```
yermulnik, line 63:
```suggestion
	getopt.StringVarLong(&params.MirrorURL, "mirror", 'm', "install from a remote API other than the default. Default (based on product):\n" + strings.Join(defaultMirrors, "\n"))
```

lib/param_parsing/parameters.go
```diff
@@ -43,10 +60,11 @@ func GetParameters() Params {
...
 	getopt.BoolVarLong(&params.ShowLatestFlag, "show-latest", 'U', "Show latest stable version")
 	getopt.StringVarLong(&params.ShowLatestPre, "show-latest-pre", 'P', "Show latest pre-release implicit version. Ex: tfswitch --show-latest-pre 0.13 prints 0.13.0-rc1 (latest)")
 	getopt.StringVarLong(&params.ShowLatestStable, "show-latest-stable", 'S', "Show latest implicit version. Ex: tfswitch --show-latest-stable 0.13 prints 0.13.7 (latest)")
+	getopt.StringVarLong(&params.Product, "product", 't', fmt.Sprintf("Specifies which product to use. Ex: `tfswitch --product opentofu` will install Terraform. Options: (%s). Default: %s", strings.Join(productIds, ", "), lib.DefaultProductId))
```
yermulnik, line 67:
```suggestion
	getopt.StringVarLong(&params.Product, "product", 't', fmt.Sprintf("Specifies which product to use. Ex: `tfswitch --product opentofu` will install OpenTofu. Options: (%s). Default: %s", strings.Join(productIds, ", "), lib.DefaultProductId))
```

lib/param_parsing/toml.go
```diff
@@ -34,11 +35,14 @@ func getParamsTOML(params Params) (Params, error) {
 		if viperParser.Get("version") != nil {
 			params.Version = viperParser.GetString("version")
 		}
+		if configKey := "product"; viperParser.Get(configKey) != nil {
```
yermulnik, line 38: Maybe add a note to https://tfswitch.warrensbox.com/usage/config-files/#use-tfswitchtoml-file-for-non-admin-users-with-limited-privilege-on-their-computers ?

lib/products.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+	}
+	return nil
+}
+
+func GetAllProducts() []Product {
+	return products
+}
+
+// Obtain produced used by deprecated public methods that
```
yermulnik, line 163:
```suggestion
// Obtain product used by deprecated public methods that
```

test-data/integration-tests/test_tfswitchtoml/.tfswitch.toml
```diff
@@ -1,3 +1,4 @@
 bin = "/usr/local/bin/terraform_from_toml"
-version = "0.11.4"
+version = "1.6.2"
 log-level = "NOTICE"
+product = "opentofu"
```
yermulnik, line 4: Very much nitpick, but still: please add correct EOF
image

lib/install.go
MatthewJohn: VSCode does this for me! If I change, it just changes it straight back! (I assumed it was a common golang styling thing)

yermulnik: I also resolved other commit suggestions of the same kind.

MatthewJohn: Thank you :)

lib/param_parsing/environment.go
MatthewJohn: Done in 1af481d

test-data/integration-tests/test_tfswitchtoml/.tfswitch.toml
MatthewJohn: Fixed in 2a3ab61

lib/param_parsing/toml.go
MatthewJohn: Fixed in e8e0707

yermulnik (approved):
LGTM
Maybe have @warrensbox to sign this off 🤔

warrensbox:
@MatthewJohn I need to test a few more cases before merging. Will get it done by 12pm PST on Jun 5.

warrensbox:
> @MatthewJohn I need to test a few more cases before merging. Will get it done by 12pm PST on Jun 6.

@MatthewJohn @yermulnik please see remarks on #420

lib/recent.go
```diff
@@ -7,61 +7,44 @@ import (
...
+	product.SetRecentVersionProduct(r, sliceToCheck)
 }
 
-func getRecentVersions(installPath string, dist string) ([]string, error) {
+func getRecentVersions(installPath string, product Product) ([]string, error) {
 	installLocation := GetInstallLocation(installPath)
 	recentFilePath := filepath.Join(installLocation, recentFile)
-	var recentFileData RecentFiles
+	var recentFileData RecentFile
 	unmarshalRecentFileData(recentFilePath, &recentFileData)
```
warrensbox, line 46: Can you check if the file exist before unmarshalling or reading?
 If a first time user downloads tfswitch, they won't have a RECENT file. They will be greeted with a error message (even though, we will create a RECENT file for them later)
 
 
 
 something like :
 
 
 
 ```
 
 	if CheckFileExist(recentFilePath) {
		unmarshalRecentFileData(recentFilePath, &recentFileData)
	}
	
	```

MatthewJohn: Since we call `unmarshalRecentFileData` in a couple of places, would you be happy to have the check inside `unmarshalRecentFileData`, so it just returns if the file doesn't exist?

MatthewJohn: Have done so in 0d34959, let me know what you think :)

lib/param_parsing/parameters.go
warrensbox: @MatthewJohn this still show Terraform instead of OpenTofu

PR #436, opened 2024-05-26
Convert recent file to json

MatthewJohn:
Replace methods for interacting with recent file with two methods:
 * get recent versions (which ignores all errors and always returns a list - only populated if the data could be successfully read)
 * add recent version, which reads the current versions and prepends to new version to add

Issue #315

### DIFF
--- a/lib/files.go
+++ b/lib/files.go
@@ -121,6 +121,7 @@ func WriteLines(lines []string, path string) (err error) {
 }
 
 // ReadLines : Read a whole file into the memory and store it as array of lines
+// @TODO This isn't used - should we keep it?
 func ReadLines(path string) (lines []string, err error) {
 	var (
 		file   *os.File
--- a/lib/install.go
+++ b/lib/install.go
@@ -1,7 +1,9 @@
 package lib
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -37,10 +39,6 @@ func initialize(binPath string) {
 	if symlinkExist {
 		RemoveSymlink(binPath)
 	}
-}
-
-func getRecentFileName(product Product) string {
-	return recentFilePrefix + product.GetId()
 }
 
 // GetInstallLocation : get location where the terraform binary will be installed,
@@ -94,7 +92,7 @@ func install(product Product, tfversion string, binPath string, installPath stri
 		/* set symlink to desired version */
 		CreateSymlink(installFileVersionPath, binPath)
 		logger.Infof("Switched terraform to version %q", tfversion)
-		addRecent(product, tfversion, installPath) //add to recent file for faster lookup
+		addRecentVersion(product, tfversion, installPath) //add to recent file for faster lookup
 		return
 	}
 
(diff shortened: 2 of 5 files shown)


lib/install.go
```diff
@@ -138,97 +136,78 @@ func install(product Product, tfversion string, binPath string, installPath stri
...
-	addRecent(product, tfversion, installPath) //add to recent file for faster lookup
+	addRecentVersion(product, tfversion, installPath) //add to recent file for faster lookup
 	return
 }
 
-// addRecent : add to recent file
-func addRecent(product Product, requestedVersion string, installPath string) {
+// getRecentFileName : Obtain recent version filename for product
+func getRecentFileName(product Product) string {
+	return recentFilePrefix + product.GetId() + ".json"
```
yermulnik, line 145: To make user's experience smooth and to maintain backward compatibility, we probably need to add 1) a code to migrate old plain TXT file to JSON seamlessly to user and 2) a notice to let user know the migration is being done from plain TXT to JSON when TXT version of file exists and no JSON version exists yet.

Also are there any obstacles or reasoning that prevents us from using the same file for any Product?
I think it's more versatile to keep the single file to track recent versions of any product.

lib/install.go
```diff
@@ -138,97 +136,78 @@ func install(product Product, tfversion string, binPath string, installPath stri
...
+			versions = append(versions[:versionIndex], versions[versionIndex+1:]...)
 		}
+	}
 
-		versionExist := versionExist(requestedVersion, lines)
+	// Add new version to start of slice
+	versions = append([]string{requestedVersion}, versions...)
+	if len(versions) > 3 {
+		versions = versions[0:2]
+	}
```
yermulnik, line 167: This pushed me to think that we may allow user to customize number of recent versions we keep for them. WDYT?

lib/install.go
```diff
@@ -138,97 +136,78 @@ func install(product Product, tfversion string, binPath string, installPath stri
...
 
-		// @TODO Does this not duplicate the behavoir of CreateRecentFile, called above? (possibly )
-		if !versionExist {
-			if len(lines) >= 3 {
-				_, lines = lines[len(lines)-1], lines[:len(lines)-1]
+	// Write new versions back to recent files
+	recentVersionFh, err := os.OpenFile(recentFilePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
```
yermulnik, line 170: I'm one of those nerds who has custom umask set to not allow group and others by default. I'm from minority of users though. But I still would prefer file creation operations to take into account `umask` value that I define and fallback to sane default of `0644` otherwise.

lib/install.go
```diff
@@ -138,97 +136,78 @@ func install(product Product, tfversion string, binPath string, installPath stri
...
+	if err != nil {
+		logger.Errorf("Error to open %q file for writing: %v", recentFilePath, err)
```
yermulnik, line 172:
```suggestion
		logger.Errorf("Failed to open recent versions file (%q) for writing: %v", recentFilePath, err)
```

lib/install.go
```diff
@@ -138,97 +136,78 @@ func install(product Product, tfversion string, binPath string, installPath stri
...
-				_ = WriteLines(lines, recentFilePath)
-			} else {
-				lines = append([]string{requestedVersion}, lines...)
-				_ = WriteLines(lines, recentFilePath)
-			}
-		}
+	// Marhsall data and write to file
+	jsonData, err := json.Marshal(versions)
+	if err != nil {
+		logger.Errorf("Failed to marshal recent versions data: %v", err)
```
yermulnik, line 180:
```suggestion
		logger.Errorf("Failed to marshal recent versions data from %q file: %v", recentFilePath, err)
```

lib/install.go
```diff
@@ -138,97 +136,78 @@ func install(product Product, tfversion string, binPath string, installPath stri
...
+	}
 
-	} else {
-		CreateRecentFile(requestedVersion, installPath, recentFilePath)
+	_, err = recentVersionFh.Write(jsonData)
+	if err != nil {
+		logger.Errorf("Failed to write recent versions file data: %v", err)
```
yermulnik, line 185:
```suggestion
		logger.Errorf("Failed to write recent versions data to %q file: %v", recentFilePath, err)
```

lib/install.go
```diff
@@ -138,97 +136,78 @@ func install(product Product, tfversion string, binPath string, installPath stri
...
-
-		lines, errRead := ReadLines(versionFile)
-		var outputRecent []string
-
-		if errRead != nil {
-			logger.Errorf("Error reading %q file: %f", versionFile, errRead)
-			return nil, errRead
+		content, err := ioutil.ReadFile(recentVersionFile)
+		if err != nil {
+			logger.Warnf("Error when opening recent version file (%s): %s", recentVersionFile, err)
```
yermulnik, line 200:
```suggestion
			logger.Warnf("Error opening recent versions file (%q): %v. Ignoring", recentVersionFile, err)
```

lib/install.go
```diff
@@ -138,97 +136,78 @@ func install(product Product, tfversion string, binPath string, installPath stri
...
-				return nil, errRead
-			}
-
-			/* 	output can be confusing since it displays the 3 most recent used terraform version
-			append the string *recent to the output to make it more user friendly
-			*/
-			outputRecent = append(outputRecent, fmt.Sprintf("%s *recent", line))
+		err = json.Unmarshal(content, &outputRecent)
+		if err != nil {
+			logger.Warnf("Error during unmarshalling recentVersionFile: ", err)
```
yermulnik, line 206:
```suggestion
			logger.Warnf("Error during unmarshalling recent versions data from %s file: %v. Ignoring", recentVersionFile, err)
```

lib/install.go
```diff
@@ -363,28 +342,54 @@ func InstallOption(listAll, dryRun bool, customBinaryPath, installPath string, m
...
 
-	if len(tflist) == 0 {
+	if len(selectVersions) == 0 {
 		logger.Fatalf("Terraform version list is empty: %s", mirrorURL)
 		os.Exit(1)
 	}
 
 	/* prompt user to select version of terraform */
 	prompt := promptui.Select{
 		Label: "Select Terraform version",
```
yermulnik, line 380: Add `TODO` note to replace `Terraform` with name of Product when PR for OpenTofu support is approved and merged? 🤔

MatthewJohn: :+1: Because of the changes in the OpenTofu branch, this a branch from that branch, so all of this is immediately available to implement.
As part of the other branch, I'll take a wider look at how many places "Terraform" is still being used in messages etc

lib/files.go
```diff
@@ -121,6 +121,7 @@ func WriteLines(lines []string, path string) (err error) {
 }
 
 // ReadLines : Read a whole file into the memory and store it as array of lines
+// @TODO This isn't used - should we keep it?
```
MatthewJohn, line 124: Must be considered in review - either remove the TODO or remove the method - same with WriteLines

lib/install.go
MatthewJohn: > Also are there any obstacles or reasoning that prevents us from using the same file for any Product?
> I think it's more versatile to keep the single file to track recent versions of any product.

Sounds good

lib/install.go
yermulnik: > this a branch from that branch, so all of this is immediately available to implement.

ouch, I had a gut feeling akin to deja vu 🤣 

> As part of the other branch, I'll take a wider look at how many places "Terraform" is still being used in logs etc

Appreciate that 🙏🏻

lib/files.go
yermulnik: As we've learned there are loads (or at least some 🤪) of the projects that use `tfswitch` as a lib or re-use its code as-is, we may probably need to announce removal of functions/methods/whatever in advance and give consumers time to at least next release to migrate/replace.
So what I'd suggest is to leave this as is and get this removed any time later after any major release and in case any of use stumble upon it by an accident or on purpose (and hence maybe rephrase it like `... - should remove anytime after `2.0.0`. IMHO.

lib/install.go
MatrixCrawler: I would suggest to keep the data in one file and use a dict/list combination to store it.
It could look like this
```
{
 "terraform": [
    "filename1",
    "filename2",
  ],
 "tofu": [
    "tofuname1",
    "tofuname2"
  ]
}
```

This would be a struct like this

```
type RecentFiles struct {
  Terraform []string `json:"terraform"`
  Tofu []string `json:"tofu"`
}
```

We could traverse/lookup easily through the lists and determine whether we have the local file or not.

lib/install.go
MatthewJohn: As mentioned in the previous PR, I suspect this is the case?
Based on: `it is created with mode perm (before umask)` (https://pkg.go.dev/github.com/pborman/getopt#section-readme)

lib/install.go
MatthewJohn: Thanks @MatrixCrawler,

Apart from having a "tigher" definition, is there any additional benefit over using a `map[string][]string`?

Because the pseudo code would look more like:

```
func getRecentVersions(product, file) []string {
  rawContent := readFile(file)
  recentVersionData := Unmarshal(rawContent)
  if product.Id in recentVersionData {
    return recentVersionData[product.Id]
  }
  return []
}
```

As opposed to:
```
func getRecentVersions(product, file) []string {
  rawContent := readFile(file)
  recentVersionData := Unmarshal(rawContent)
  if product.Id == "terraform" {
    return recentVersionData.Terraform
  }
  if product.Id == "opentofu" {
    return recentVersionData.OpenTofu
  }
  return []
}
```
which (I guess theoretically due to chance of this happening) would require up-keep to match the available products.

Though, I guess, to put the ownership of this back to the product itself, it could look more like:
```
func getRecentVersions(product, file) []string {
  rawContent := readFile(file)
  recentVersionData := Unmarshal(rawContent)
  return product.GetRecentVersionFromContent(recentVersionData)
}

func (p TerraformProduct) GetRecentVersionFromContent(recentVersionData RecentFiles) {
  return recentVersionData.Terraform
}
func (p OpenTofuProduct) GetRecentVersionFromContent(recentVersionData RecentFiles) {
  return recentVersionData.OpenTofu
}
```

MatrixCrawler: I am a fan of "clear" definitions. Sure, when you break it down this is nothing else than a map[string][]string but i think the code is better readable that way.
I have a suggestion for the conversion of the recent file to json
See https://github.com/warrensbox/terraform-switcher/pull/437 if you'd like.

MatthewJohn: Thanks @MatrixCrawler

What do you think of the final suggestion in the previous comment, where we use a concrete structure, as you suggested, but the responsibility for returning the correct attribute is delegated to the product itself?

MatthewJohn: Thanks @MatrixCrawler

What do you think of the final suggestion in the previous comment, where we use a concrete structure, as you suggested, but the responsibility for returning the correct attribute is delegated to the product itself?

Thanks for #437 - it looks how I'd have imagined - though seeing how you handled the data migration was good to see. 
Were you hoping to merge this into somewhere or purely as a demonstration? :)

MatrixCrawler: I did not want to override your suggestion. But if you are okay with that we could go with my solution. I need to add some tests though before we can merge that.

MatrixCrawler: > What do you think of the final suggestion in the previous comment, where we use a concrete structure, as you suggested, but the responsibility for returning the correct attribute is delegated to the product itself?
> 

That looks like a nice approach using an interface.
I went with a solution that simply returns exactly the same as the old functions did for simpicity. But i am not opposed to your solution at all.

MatthewJohn: Can do - I had a local version using the new structure, but happy to go with yours :)

MatrixCrawler: Then i will add some more tests

MatthewJohn: I'll just need to work on integrating those changes into #435, as this PR was based on the branch for that :)
I guess, adding the changes for OpenTofu might be a little.. odd, if #437 is merged before that.

MatthewJohn: If you did want to merge your PR prior to #435 , then I think it makes sense, but if it does depend on it and/or makes more sense to ship together, is it worth either branching from #435 and making the changes or even branching from this PR to make the suggested changes?

MatrixCrawler: My branch is based on master an can be merged prior to #435. It should be fully backwards compatible.

lib/install.go
yermulnik: Yep, the other thread resolved my doubts. Thanks. I'm resolving this thread.

lib/install.go
MatthewJohn: @yermulnik Just so you see the thread :)

yermulnik: Thanks for highlighting it. It looks good to me (despite all those Golang bits, I have no understanding of, you guys are discussing here 😀)

yermulnik: @MatrixCrawler Also the comments you added to #437 all make perfect sense (lib naming, ordering, pruning old items, so on) — thanks for taking thorough look 🤝

lib/install.go
MatthewJohn: Fixed :)

lib/install.go
MatthewJohn: I'm planning to merge this PR (which will go into #435) and I'll re-raise this comment there :)

lib/files.go
MatthewJohn: I'm going to merge this PR into #435 and I'll re-raise the comment there :)

lib/install.go
MatthewJohn: Since yours closely resembles this PR and there's quite a few changes that will result in needing to be re-made for the #435 , which contains the "product" changes. I'll merge this PR (into #435) and I will deal with the fun merge of main :D)

lib/install.go
yermulnik: Thank you

lib/install.go
yermulnik: > and I will deal with the fun merge of main :D)

What an outstanding example of bravery =) Kudos to you man!

MatthewJohn: IF this plan goes horrendously wrong and #437 causes turmoil - I can always back the merge :D

lib/install.go
warrensbox: Instead of letting users customize, lets leave it at 5 for now (increase from 3). I don't think we should over-engineer this.

PR #437, opened 2024-05-27
RECENT to JSON Proposal

MatrixCrawler:
suggestion for implementing storing the recent versions in json format

- added conversion of existing file
- added test cases


lib/recent.go
```diff
@@ -0,0 +1,89 @@
+package lib
```
MatthewJohn, line 1: I actually started going down this route, before deciding against it to make the diffing easier (at least initially) and only _just_ removed the draft file :D 
Screenshot from 2024-05-27 08-38-04

lib/recent.go
```diff
@@ -0,0 +1,92 @@
...
+		if s != "" {
+			recentFileData.Terraform = append(recentFileData.Terraform, s)
+		}
+	}
+}
+
+func saveFile(data RecentFiles, path string) {
+	bytes, err := json.Marshal(data)
+	if err != nil {
+		logger.Errorf("Could not Marshal data to json. %q", err)
```
yermulnik, line 86:
```suggestion
		logger.Errorf("Could not marshal data to JSON: %v", err)
```

lib/recent.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	err = os.WriteFile(path, bytes, 0644)
+	if err != nil {
+		logger.Errorf("Could not save file %v", path)
```
yermulnik, line 90:
```suggestion
		logger.Errorf("Could not save file %q: %v", path, err)
```

lib/recent.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func getRecentVersions(installPath string, dist string) ([]string, error) {
+	installLocation := GetInstallLocation(installPath)
+	recentFilePath := filepath.Join(installLocation, recentFile)
+	var recentFileData RecentFiles
+	unmarshal(recentFilePath, &recentFileData)
+	if dist == distTerraform {
+		return recentFileData.Terraform, nil
+	} else if dist == distTofu {
```
MatthewJohn, line 53: For visibility in this PR (from https://github.com/warrensbox/terraform-switcher/pull/436):
I'll just need to work on integrating these changes into https://github.com/warrensbox/terraform-switcher/pull/435, as this PR was based on the branch for that :)

I guess, adding the changes for OpenTofu might be a little.. odd, if https://github.com/warrensbox/terraform-switcher/pull/437 is merged before that.

If you did want to merge this PR prior to https://github.com/warrensbox/terraform-switcher/pull/435 , then I think it makes sense (though the references to OpenTofu are possibly strange), but if it does depend on it and/or makes more sense to ship together, is it worth either branching from https://github.com/warrensbox/terraform-switcher/pull/435 and making the changes or even branching from this PR to make the suggested changes?

lib/recent_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+func Test_convertData(t *testing.T) {
+	recentFileContent, err := os.ReadFile("../test-data/recent/RECENT_OLD_FORMAT")
+	if err != nil {
+		logger.Error("Could not open file ../test-data/recent/RECENT_OLD_FORMAT")
+	}
+
+	var recentFileData RecentFiles
+	convertData(recentFileContent, &recentFileData)
+	assert.Equal(t, "1.5.6", recentFileData.Terraform[0])
```
MatthewJohn, line 18: Probably worth asserting the length of `recentFileData.Terraform` before referencing slice indexes

lib/recent_test.go
```diff
@@ -0,0 +1,89 @@
...
+	assert.Equal(t, 0, len(recentFileData.Tofu))
+}
+
+func Test_saveFile(t *testing.T) {
+	var recentFileData = RecentFiles{
+		Terraform: []string{"1.2.3", "4.5.6"},
+		Tofu:      []string{"6.6.6"},
+	}
+	temp, err := os.MkdirTemp("", "recent-test")
+	defer func(path string) {
```
MatthewJohn, line 31: Would this be simpler as:
```
	defer func() {
		_ = os.RemoveAll(temp)
	}()
```
?

lib/recent_test.go
```diff
@@ -0,0 +1,89 @@
...
+		_ = os.RemoveAll(path)
+	}(temp)
+	if err != nil {
```
MatthewJohn, line 34: The error should be checked before performing the defer, I think

lib/recent_test.go
```diff
@@ -0,0 +1,89 @@
...
+	strings, err := getRecentVersions("../test-data/recent/recent_as_json/", distTerraform)
+	if err != nil {
+		t.Error("Unable to get versions from recent file")
+	}
+	assert.Equal(t, []string{"1.2.3", "4.5.6"}, strings)
+}
+
+func Test_getRecentVersionsGenericForTofu(t *testing.T) {
+	logger = InitLogger("DEBUG")
+	strings, err := getRecentVersions("../test-data/recent/recent_as_json/", distTofu)
```
MatthewJohn, line 58:
```suggestion
	strings, err := getRecentVersions("../test-data/recent/recent_as_json", distTofu)
```
The training slash feels strange

lib/recent_test.go
```diff
@@ -0,0 +1,89 @@
...
+	saveFile(recentFileData, pathToTempFile)
+
+	content, err := os.ReadFile(pathToTempFile)
+	if err != nil {
+		t.Errorf("Could not read converted file %v", pathToTempFile)
+	}
+	assert.Equal(t, "{\"terraform\":[\"1.2.3\",\"4.5.6\"],\"tofu\":[\"6.6.6\"]}", string(content))
+}
+
+func Test_getRecentVersionsGenericForTerraform(t *testing.T) {
```
MatthewJohn, line 47: The function isn't called Generic anymore :)

lib/recent_test.go
```diff
@@ -0,0 +1,89 @@
...
+	logger = InitLogger("DEBUG")
+	temp, err := os.MkdirTemp("", "recent-test")
+	defer func(path string) {
+		_ = os.RemoveAll(path)
+	}(temp)
+	if err != nil {
+		t.Errorf("Could not create temporary directory")
+	}
+	addRecent("3.7.0", temp, distTerraform)
+	bytes, err := os.ReadFile(filepath.Join(temp, ".terraform.versions/RECENT"))
```
MatthewJohn, line 75:
```suggestion
	bytes, err := os.ReadFile(filepath.Join(temp, ".terraform.versions", "RECENT"))
```
The `filepath.Join` to be useful should be used for joining all parts of the path

lib/recent_test.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		t.Error("Could not open file")
+		t.Error(err)
+	}
+	assert.Equal(t, "{\"terraform\":[\"3.7.0\"],\"tofu\":null}", string(bytes))
+
+	addRecent("1.1.1", temp, distTofu)
+	bytes, err = os.ReadFile(filepath.Join(temp, ".terraform.versions/RECENT"))
```
MatthewJohn, line 83:
```suggestion
	bytes, err = os.ReadFile(filepath.Join(temp, ".terraform.versions", "RECENT"))
```

lib/recent.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	if dist == distTerraform {
+		recentFileData.Terraform = append(recentFileData.Terraform, requestedVersion)
+	} else if dist == distTofu {
+		recentFileData.Tofu = append(recentFileData.Tofu, requestedVersion)
+	}
+	saveFile(recentFileData, recentFilePath)
+}
+
+func getRecentVersions(installPath string, dist string) ([]string, error) {
```
MatthewJohn, line 46: `getRecentVersions` previously added the suffix " *recent", for display purposes.
I wasn't keen and addressed this in my PR, but the functionality will be lost with this change

MatthewJohn (changes requested):

lib/defaults.go
```diff
@@ -27,8 +23,11 @@ func GetDefaultBin() string {
 const (
 	DefaultMirror             = "https://releases.hashicorp.com/terraform"
 	DefaultLatest             = ""
+	distTerraform             = "terraform"
+	distTofu                  = "tofu"
```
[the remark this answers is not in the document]
MatthewJohn: distribution, I'm assuming? Though, I wonder if people would get annoyed if we called Terraform/OpenTofu distributions, since it's sort of implies two different variants/releases from a common source (at least, my interpretation) :D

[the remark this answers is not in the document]
MatrixCrawler: yes distribution it is. just needed a variable name. if you have something better i am all ears 😁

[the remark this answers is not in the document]
yermulnik: Maybe when @MatthewJohn gets this PR incorporated into #435 we can use https://github.com/warrensbox/terraform-switcher/blob/315-opentofu-support/lib/products.go for this too? 🤔 
Otherwise I'd think of something like `binTerraform` and `binOpenTofu` (not ideal but at least eliminates `dist` confusion for me 🤷🏻)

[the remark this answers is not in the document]
yermulnik: > (not ideal but at least eliminates `dist` confusion for me 🤷🏻)

I'm assuming myself as one of the ordinary users that sometimes look into code 😺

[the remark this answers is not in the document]
MatthewJohn: > Maybe when @MatthewJohn gets this PR incorporated into https://github.com/warrensbox/terraform-switcher/pull/435 we can use https://github.com/warrensbox/terraform-switcher/blob/315-opentofu-support/lib/products.go for this too? 🤔

Yep :)

lib/recent.go
```diff
@@ -0,0 +1,92 @@
...
+		sliceToCheck = recentFileData.Tofu
+	}
+	for _, v := range sliceToCheck {
+		if v == requestedVersion {
+			// entry already exists. Nothing to do
+			return
+		}
+	}
+	if dist == distTerraform {
+		recentFileData.Terraform = append(recentFileData.Terraform, requestedVersion)
```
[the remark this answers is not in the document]
MatrixCrawler: I will have a look into this. Thanks for pointing out.

lib/recent_test.go
MatrixCrawler: No idea :)
I just let IntelliJ generate thos stubs. But maybe you are right.

lib/recent.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	var sliceToCheck []string
+	if dist == distTerraform {
+		sliceToCheck = recentFileData.Terraform
+	} else if dist == distTofu {
+		sliceToCheck = recentFileData.Tofu
+	}
+	for _, v := range sliceToCheck {
+		if v == requestedVersion {
+			// entry already exists. Nothing to do
```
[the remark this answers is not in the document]
MatrixCrawler: Solved that problem. Existing will be moved first.

lib/recent.go
MatrixCrawler: added that functionality.

lib/recent_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"github.com/stretchr/testify/assert"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func Test_convertData(t *testing.T) {
+	recentFileContent, err := os.ReadFile("../test-data/recent/RECENT_OLD_FORMAT")
```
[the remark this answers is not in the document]
MatrixCrawler: Agreed. Done that.

lib/recent_test.go
MatrixCrawler: Done

lib/recent_test.go
MatrixCrawler: Done

lib/recent.go
```diff
@@ -0,0 +1,119 @@
...
+		convertOldRecentFile(recentFileContent, recentFileData)
+	} else {
+		err = json.Unmarshal(recentFileContent, &recentFileData)
+		if err != nil {
+			logger.Errorf("Could not unmarshal recent versions content from %q file", recentFilePath)
+		}
+	}
+}
+
+func convertOldRecentFile(content []byte, recentFileData *RecentFiles) {
```
warrensbox, line 101: Nice addition!

lib/defaults.go
```diff
@@ -27,9 +23,12 @@ func GetDefaultBin() string {
...
 	DefaultLatest             = ""
+	distributionTerraform     = "terraform"
+	distributionOpenTofu      = "opentofu"
 	installFile               = "terraform"
 	InstallDir                = ".terraform.versions"
+	pubKeySuffix              = ".asc"
 	recentFile                = "RECENT"
+	TerraformPrefix           = "terraform_"
 	tfDarwinArm64StartVersion = "1.0.2"
-	VersionPrefix             = "terraform_"
```
MatthewJohn, line 34: This was a public constant, we maybe we need to consider if it was being used externally

lib/recent.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	for versionIndex, versionValue := range sliceToCheck {
+		if versionValue == version {
+			sliceToCheck = append(sliceToCheck[:versionIndex], sliceToCheck[versionIndex+1:]...)
+		}
+	}
+	sliceToCheck = append([]string{version}, sliceToCheck...)
+
+	if len(sliceToCheck) > 3 {
+		deleteDownloadedBinaries(installPath, distribution, sliceToCheck[3:])
```
MatthewJohn, line 45: Do we *want* to delete these binaries. If tfswitch sets an arbitrary limit of 3 versions, does that mean that user's should *only* cache 3 versions.
I'd had thought the recentVersions was _only_ a UI thing and not related to how many versions were being cached.

lib/recent.go
```diff
@@ -0,0 +1,119 @@
...
+	"strings"
+)
+
+type RecentFiles struct {
+	Terraform []string `json:"terraform"`
+	OpenTofu  []string `json:"openTofu"`
+}
+
+func addRecent(requestedVersion string, installPath string, distribution string) {
+	if !validVersionFormat(requestedVersion) {
```
MatthewJohn, line 16: I'm not sure I really understand the need for this check.
Since it's an internal method, it can't be called externally.
Do we really expect to have code paths that would allow an invalid version to be passed in, especially since it's called post-install

lib/recent.go
MatthewJohn: For reference, at least for me, I use quite a few versions of terraform regularly (more than 3 for sure) and it's nice to always use local caches

lib/recent_test.go
```diff
@@ -0,0 +1,150 @@
...
+func Test_saveFile(t *testing.T) {
+	var recentFileData = RecentFiles{
+		Terraform: []string{"1.2.3", "4.5.6"},
+		OpenTofu:  []string{"6.6.6"},
+	}
+	temp, err := os.MkdirTemp("", "recent-test")
+	if err != nil {
+		t.Errorf("Could not create temporary directory")
+	}
+	defer func(path string) {
```
MatthewJohn, line 33: I'm not 100% sure, but I think that simply
```
  defer os.RemoveAll(temp)
```
would provide the same functionality

lib/recent_test.go
MatthewJohn: Ah, I'd lost this in the diff-view, I mentioned that I think:
```
  defer os.RemoveAll(temp)
```
might be simpler :)

lib/recent_test.go
MatrixCrawler: It would but would not ignore errors explicitely. At least my IDE states thats.

lib/recent.go
MatrixCrawler: I just added the check again as it already was in the old function that was adding versions to the recent file.

lib/recent.go
MatrixCrawler: I agree. I wasn't aware that this was only a UI thing (why?) I thought this method missing. I can remove it

MatthewJohn: I assume just to avoid cluttering up the UI. @warrensbox ?

warrensbox: The RECENT version is for user to switch easily between their most used version.

RECENT is only a UI thing. We should not delete other version. It's nice to have the cache there.
@MatrixCrawler

warrensbox:
Once you resolve the comments, It's ready to merge

MatrixCrawler:
@MatthewJohn requested a change that i cannot resolve.

lib/recent.go
```diff
@@ -0,0 +1,107 @@
...
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+type RecentFiles struct {
+	Terraform []string `json:"terraform"`
+	OpenTofu  []string `json:"openTofu"`
```
yermulnik, line 12: It's probably worth of lowercasing `openTofu` for consistency with `terraform`

lib/recent.go
```diff
@@ -0,0 +1,107 @@
...
+	saveRecentFile(recentFileData, recentFilePath)
+}
+
+func prependRecentVersionToList(version, distribution string, r *RecentFiles) {
+	var sliceToCheck []string
+	if distribution == distributionTerraform {
+		sliceToCheck = r.Terraform
+	} else if distribution == distributionOpenTofu {
+		sliceToCheck = r.OpenTofu
+	}
```
yermulnik, line 36: Wouldn't `switch` be a bit more streamlined here? 🤔

lib/recent.go
```diff
@@ -0,0 +1,107 @@
...
+			sliceToCheck = append(sliceToCheck[:versionIndex], sliceToCheck[versionIndex+1:]...)
+		}
+	}
+	sliceToCheck = append([]string{version}, sliceToCheck...)
+
+	if distribution == distributionTerraform {
+		r.Terraform = sliceToCheck
+	} else if distribution == distributionOpenTofu {
+		r.OpenTofu = sliceToCheck
+	}
```
yermulnik, line 48: Wouldn't `switch` be a bit more streamlined here? 🤔

lib/recent.go
```diff
@@ -0,0 +1,107 @@
...
+	installLocation := GetInstallLocation(installPath)
+	recentFilePath := filepath.Join(installLocation, recentFile)
+	var recentFileData RecentFiles
+	unmarshalRecentFileData(recentFilePath, &recentFileData)
+	var listOfRecentVersions []string
+	if dist == distributionTerraform {
+		listOfRecentVersions = recentFileData.Terraform
+	} else if dist == distributionOpenTofu {
+		listOfRecentVersions = recentFileData.OpenTofu
+	}
```
yermulnik, line 61: Wouldn't `switch` be a bit more streamlined here? 🤔

lib/recent.go
```diff
@@ -0,0 +1,107 @@
...
+	var maxCount int
+	if len(listOfRecentVersions) >= 3 {
+		maxCount = 3
```
yermulnik, line 64: I might be mixing up things, though I seem to saw somewhere else(?) that @warrensbox said 5 would be better. Am I mixing different things?

lib/recent.go
MatrixCrawler: Will do

lib/recent.go
MatrixCrawler: Changed to switch

lib/recent.go
MatrixCrawler: Changed to switch

lib/recent.go
MatrixCrawler: Changed to switch

lib/recent.go
MatrixCrawler: I just took over the existing value. If we want to change it, i missed it. @warrensbox

MatthewJohn: I think it was mentioned in the other PR IIRC!

MatthewJohn:
> @MatthewJohn requested a change that i cannot resolve.

What do you mean, sorry? As in it's a difficult problem or Github doesn't let you resolve it?

MatthewJohn (approved):

lib/recent.go
yermulnik: Let's probably have @warrensbox confirm whether to go with the existing value or change it to 5 while we're here?

yermulnik (approved):

lib/recent.go
MatrixCrawler: https://github.com/warrensbox/terraform-switcher/pull/435/files#r1623089267
This is the comment where he saif to set it to 5.
Will do so

PR #446, opened 2024-06-04
skip any parsing of config files if the -v flag is provided

### LINKED ISSUE #443: tfswitch -v should not apply anything - it should only print the tfswitch version
Currently the tfswitch -v reads config files and applies them when you pass `-v`. Previously, it only printed the tfswitch version

tfswitch -v should not apply anything - it should only print the tfswitch version

Current:
```
warrenv@mass ~/WarrensBox/go/src/github.com/warrensbox/tfswitch/test-data/integration-tests/test_versiontf --> tfswitch -v
13:39:12.681 INFO Reading version from Terraform module at "."
13:39:12.754 INFO Reading required version from constraint: "1.6.3"
13:39:12.755 INFO Matched version: "1.6.3"
Version: v1.1.1
```

Needs to look like:
```
Version: v1.1.1
```

MatrixCrawler:
Fixes #443 
- skip any parsing of config files if the -v flag is provided

### DIFF
--- a/lib/param_parsing/parameters.go
+++ b/lib/param_parsing/parameters.go
@@ -52,32 +52,33 @@ func GetParameters() Params {
 	// Parse the command line parameters to fetch stuff like chdir
 	getopt.Parse()
 
-	oldLogLevel := params.LogLevel
-	logger = lib.InitLogger(params.LogLevel)
-	var err error
-	// Read configuration files
-	if tomlFileExists(params) {
-		params, err = getParamsTOML(params)
-	} else if tfSwitchFileExists(params) {
-		params, err = GetParamsFromTfSwitch(params)
-	} else if terraformVersionFileExists(params) {
-		params, err = GetParamsFromTerraformVersion(params)
-	} else if isTerraformModule(params) {
-		params, err = GetVersionFromVersionsTF(params)
-	} else if terraGruntFileExists(params) {
-		params, err = GetVersionFromTerragrunt(params)
-	} else {
-		params = GetParamsFromEnvironment(params)
-	}
-	if err != nil {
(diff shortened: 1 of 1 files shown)


lib/param_parsing/parameters.go
```diff
@@ -52,32 +52,33 @@ func GetParameters() Params {
...
-	} else {
-		params = GetParamsFromEnvironment(params)
-	}
-	if err != nil {
-		logger.Fatalf("Error parsing configuration file: %q", err)
-	}
-
-	// Logger config was changed by the config files. Reinitialise.
-	if params.LogLevel != oldLogLevel {
+	if !params.VersionFlag {
```
yermulnik, line 55: Should we add all of the params I listed at https://github.com/warrensbox/terraform-switcher/issues/443#issuecomment-2147257097 ? 🤔

MatrixCrawler: That would be complicated. We would need to parse the options a third time.
I don't see a problem in parsing the other variables though.

yermulnik: > That would be complicated.

Ah, that's weird 😢 
* `params.HelpFlag`
* `params.LatestFlag`
* `params.LatestPre`
* `params.LatestStable`
* `params.ListAllFlag`
* `params.ShowLatestFlag`
* `params.ShowLatestPre`
* `params.ShowLatestStable`
* `params.VersionFlag`

It's just pointless to find TF version for these options.

Since this would require more effort, let's go with baby steps then =)

yermulnik (approved):

warrensbox:
Thanks @MatrixCrawler !

Issue #422, opened 2024-04-30
How do I turn off logging?


grimm26:
Since upgrading from 1.0.2 to 1.1.1 tfswitch spews all sorts of INFO level logging. I only want to see that if I enable it. How do I turn it off?

warrensbox:
Can you show an example of too much logs please?

grimm26:
![image](https://github.com/warrensbox/terraform-switcher/assets/127401/cadb112c-c7d0-424c-b261-26c0866d8036)

grimm26:
that last entry is because I have a hook to switch tf version based on something besides the version constraint. So I get 3 lines from tfswitch's automagic switch and then one for the switch I actually want.

warrensbox:
@grimm26 what was working in the previous releases that has stopped working in this release?

Maybe that would help us understand what you need?

grimm26:
Nothing stopped working, I just have a lot of spam scrolling by on my screen when I switch versions and there doesn't seem to be a way to turn it off

MatrixCrawler:
So you want something like a --no-log flag or a config parameter in the TOML File that allows you to run tfswitch in silent mode. Did i understand that correctly?

yermulnik:
> So you want something like a --no-log flag or a config parameter in the TOML File that allows you to run tfswitch in silent mode.

This kind of option already exists: `log-level` — we only need to ensure it works as expected and tune it by adding more options to reduce logging to only Errors, Fatals, Panics I guess.

MatrixCrawler:
Added a --log-level=ERROR with #447

PR #447, opened 2024-06-04
Added ERRORlog-level

MatrixCrawler:
- added --log-level=ERROR option to supress all non-error logging.

Likely solves #422 

image


MatrixCrawler:
> I'm pretty sure we shouldn't be suppressing Errors, Fatals, and Panics.

Right you are. The is not a wise decision. Reverted the changes to introduce an ERROR log level

yermulnik (approved):

PR #457, opened 2024-06-08
Add error return values to signatures of public methods to allow migrating to returning errors rather than Fatals in future

MatthewJohn:
See conversation in #452


lib/install.go
```diff
@@ -347,4 +349,5 @@ func InstallProductOption(product Product, listAll, dryRun bool, customBinaryPat
 		install(product, selectVersions[selectedItx].Version, customBinaryPath, installPath, mirrorURL)
 	}
 	os.Exit(0)
+	return nil
```
yermulnik, line 352: `return` after `os.Exit()` makes no sense. if this is in preparation to future changes, please leave an explanatory comment about this.

lib/install.go
```diff
@@ -268,6 +269,7 @@ func InstallProductVersion(product Product, dryRun bool, version, customBinaryPa
 			os.Exit(1)
 		}
 	}
+	return nil
```
yermulnik, line 272: I guess it's best to leave a comment that `os.Exit()` is to be changed to `return` probably.

warrensbox: The reason we are doing this is because we want our libraries to be accessible by other users. Exit probably does not provide enough info to users using our library. 

However, we may need to Exit(1), so bots that are using our tool knows that something went wrong. I suggest we Exit(1) at the parent function that is calling these functions

lib/install.go
warrensbox: @yermulnik has a valid point.

@MatthewJohn as long as the parent function exits appropriately, we should be good. The parent can accept the return value (nil, not nil, something else), and then decide to exit (1 or 0)

MatthewJohn: Yes, I agree - this PR was purely the bare minimum to get the function definition update for future improvement (and even after an os.Exit, it seems go still requires a return value), so that when these new methods are released, we don't need to update the definitions, but can slowly start to remove the other means of handling errors/exiting (so we always return to the calling function with an error (or success)

MatthewJohn: Have improved this - let me know what you think :)

lib/install.go
```diff
@@ -338,13 +340,13 @@ func InstallProductOption(product Product, listAll, dryRun bool, customBinaryPat
 	if errPrompt != nil {
 		if errPrompt.Error() == "^C" {
 			// Cancel execution
-			os.Exit(1)
+			return fmt.Errorf("user interupt")
```
yermulnik, line 343:
```suggestion
			return fmt.Errorf("user interrupt")
```

lib/install.go
MatthewJohn: This should be done :)

main.go
```diff
@@ -44,32 +45,36 @@ func main() {
...
-		lib.InstallProductVersion(parameters.ProductEntity, parameters.DryRun, parameters.DefaultVersion, parameters.CustomBinaryPath, parameters.InstallPath, parameters.MirrorURL)
+		err = lib.InstallProductVersion(parameters.ProductEntity, parameters.DryRun, parameters.DefaultVersion, parameters.CustomBinaryPath, parameters.InstallPath, parameters.MirrorURL)
 	default:
 		// Set list all false - only official release will be displayed
-		lib.InstallProductOption(parameters.ProductEntity, false, parameters.DryRun, parameters.CustomBinaryPath, parameters.InstallPath, parameters.MirrorURL)
+		err = lib.InstallProductOption(parameters.ProductEntity, false, parameters.DryRun, parameters.CustomBinaryPath, parameters.InstallPath, parameters.MirrorURL)
+	}
+	if err != nil {
+		logger.Error(err)
+		os.Exit(1)
```
yermulnik, line 78: Why not `logger.Fatal(err)`?

PS: I'm a bit hesitant on whether Error+Exit or Fatal (which implies Exit in current logger configuration) 🤔 The former seems to fit fine, though if any of `case` statements from the `main` return error, ain't that mean something weird happened and `main` should "escalate" from an ordinary Error to a more explicit Fatal? 🤔

@warrensbox WDYT?

yermulnik (approved):

warrensbox (approved):

PR #458, opened 2024-06-08
feat: Allow for case-insensitve matching of products

MatthewJohn:


yermulnik (approved):
Nice solution 👍🏻
Re the PR title: I'd say it's a feature, rather than a fix.

PR #464, opened 2024-06-13
Add debug logging when successfully obtaining parameter from location…

MatthewJohn:
… or when terraform exists but does not provide a version

PR #468, opened 2024-06-15
Fix user bin path on windows

PierreTechoueyres:
Allow use of `bin` keyword in toml config file

### DIFF
--- a/lib/install.go
+++ b/lib/install.go
@@ -133,6 +133,7 @@ func ConvertExecutableExt(fpath string) string {
 
 // installableBinLocation : Checks if terraform is installable in the location provided by the user.
 // If not, create $HOME/bin. Ask users to add  $HOME/bin to $PATH and return $HOME/bin as install location
+// Deprecated: This function has been deprecated and will be removed in v2.0.0
 func installableBinLocation(product Product, userBinPath string) string {
 	homedir := GetHomeDirectory()         //get user's home directory
 	binDir := Path(userBinPath)           //get path directory from binary path
--- a/lib/install.go
+++ b/lib/install.go
@@ -58,10 +58,7 @@ func install(product Product, tfversion string, binPath string, installPath stri
 	installFileVersionPath := ConvertExecutableExt(filepath.Join(installLocation, product.GetVersionPrefix()+tfversion))
 	recentDownloadFile := CheckFileExist(installFileVersionPath)
 	if recentDownloadFile {
-		ChangeProductSymlink(product, installFileVersionPath, binPath)
-		logger.Infof("Switched %s to version %q", product.GetName(), tfversion)
-		addRecent(tfversion, installPath, product) //add to recent file for faster lookup
-		return nil
+		return switchToVersion(product, tfversion, binPath, installPath, installFileVersionPath)
 	}
 
 	// If the requested version had not been downloaded before
@@ -70,7 +67,7 @@ func install(product Product, tfversion string, binPath string, installPath stri
 	exist := versionExist(tfversion, tflist) // Check if version exists before downloading it
 
 	if !exist {
-		return fmt.Errorf("the provided terraform version does not exist: %q.\n Try `tfswitch -l` to see all available versions", tfversion)
+		return fmt.Errorf("the provided %s version does not exist: %q.\n Try `tfswitch -l` to see all available versions", product.GetId(), tfversion)
 	}
 
 	goarch := runtime.GOARCH
@@ -107,7 +104,15 @@ func install(product Product, tfversion string, binPath string, installPath stri
 	/* remove zipped file to clear clutter */
 	RemoveFiles(zipFile)
 
-	ChangeProductSymlink(product, installFileVersionPath, binPath)
+	return switchToVersion(product, tfversion, binPath, installPath, installFileVersionPath)
+}
+
+func switchToVersion(product Product, tfversion string, binPath string, installPath string, installFileVersionPath string) error {
+	err := ChangeProductSymlink(product, installFileVersionPath, binPath)
+	if err != nil {
+		return err
+	}
+
 	logger.Infof("Switched %s to version %q", product.GetName(), tfversion)
 	addRecent(tfversion, installPath, product) //add to recent file for faster lookup
 	return nil
--- a/lib/install.go
+++ b/lib/install.go
@@ -6,7 +6,6 @@ import (
 	"path/filepath"
 	"runtime"
 	"strconv"
-	"strings"
 	"sync"
 
 	"github.com/manifoldco/promptui"
@@ -51,17 +50,28 @@ func GetInstallLocation(installPath string) string {
 }
 
 // install : install the provided version in the argument
-func install(product Product, tfversion string, binPath string, installPath string, mirrorURL string) {
+func install(product Product, tfversion string, binPath string, installPath string, mirrorURL string) error {
 	var wg sync.WaitGroup
-	/* Check to see if user has permission to the default bin location which is  "/usr/local/bin/terraform"
-	 * If user does not have permission to default bin location, proceed to create $HOME/bin and install the tfswitch there
-	 * Inform user that they don't have permission to default location, therefore tfswitch was installed in $HOME/bin
-	 * Tell users to add $HOME/bin to their path
-	 */
-	binPath = installableBinLocation(product, binPath)
 
-	initialize(binPath)                               //initialize path
-	installLocation = GetInstallLocation(installPath) //get installation location -  this is where we will put our terraform binary file
+	//check to see if the requested version has been downloaded before
+	installLocation := GetInstallLocation(installPath)
+	installFileVersionPath := ConvertExecutableExt(filepath.Join(installLocation, product.GetVersionPrefix()+tfversion))
+	recentDownloadFile := CheckFileExist(installFileVersionPath)
+	if recentDownloadFile {
+		ChangeProductSymlink(product, installFileVersionPath, binPath)
+		logger.Infof("Switched %s to version %q", product.GetName(), tfversion)
+		addRecent(tfversion, installPath, product) //add to recent file for faster lookup
+		return nil
+	}
+
+	// If the requested version had not been downloaded before
+	// Set list all true - all versions including beta and rc will be displayed
+	tflist, _ := getTFList(mirrorURL, true)  // Get list of versions
+	exist := versionExist(tfversion, tflist) // Check if version exists before downloading it
+
+	if !exist {
+		return fmt.Errorf("the provided terraform version does not exist: %q.\n Try `tfswitch -l` to see all available versions", tfversion)
+	}
 
 	goarch := runtime.GOARCH
 	goos := runtime.GOOS
@@ -71,33 +81,6 @@ func install(product Product, tfversion string, binPath string, installPath stri
 	tf102, _ := version.NewVersion(tfDarwinArm64StartVersion)
 	if goos == "darwin" && goarch == "arm64" && tfver.LessThan(tf102) {
 		goarch = "amd64"
-	}
-
-	/* check if selected version already downloaded */
-	installFileVersionPath := ConvertExecutableExt(filepath.Join(installLocation, product.GetVersionPrefix()+tfversion))
-	fileExist := CheckFileExist(installFileVersionPath)
-
-	/* if selected version already exist, */
-	if fileExist {
-
-		/* remove current symlink if exist*/
-		symlinkExist := CheckSymlink(binPath)
-
-		if symlinkExist {
-			RemoveSymlink(binPath)
-		}
-
-		/* set symlink to desired version */
-		CreateSymlink(installFileVersionPath, binPath)
-		logger.Infof("Switched %s to version %q", product.GetName(), tfversion)
-		addRecent(tfversion, installPath, product) //add to recent file for faster lookup
-		return
-	}
-
-	//if does not have slash - append slash
-	hasSlash := strings.HasSuffix(mirrorURL, "/")
-	if !hasSlash {
-		mirrorURL = fmt.Sprintf("%s/", mirrorURL)
 	}
 
 	/* if selected version already exist, */
@@ -124,18 +107,10 @@ func install(product Product, tfversion string, binPath string, installPath stri
 	/* remove zipped file to clear clutter */
 	RemoveFiles(zipFile)
 
-	/* remove current symlink if exist*/
-	symlinkExist := CheckSymlink(binPath)
-
-	if symlinkExist {
-		RemoveSymlink(binPath)
-	}
-
-	/* set symlink to desired version */
-	CreateSymlink(installFileVersionPath, binPath)
+	ChangeProductSymlink(product, installFileVersionPath, binPath)
 	logger.Infof("Switched %s to version %q", product.GetName(), tfversion)
 	addRecent(tfversion, installPath, product) //add to recent file for faster lookup
-	return
+	return nil
 }
 
 // ConvertExecutableExt : convert executable with local OS extension
@@ -199,7 +174,7 @@ func InstallLatestVersion(dryRun bool, customBinaryPath, installPath string, mir
 func InstallLatestProductVersion(product Product, dryRun bool, customBinaryPath, installPath string, mirrorURL string) error {
 	tfversion, _ := getTFLatest(mirrorURL)
 	if !dryRun {
-		install(product, tfversion, customBinaryPath, installPath, mirrorURL)
+		return install(product, tfversion, customBinaryPath, installPath, mirrorURL)
 	}
 	return nil
 }
@@ -242,28 +217,7 @@ func InstallProductVersion(product Product, dryRun bool, version, customBinaryPa
 	if !dryRun {
 		if validVersionFormat(version) {
 			requestedVersion := version
-
-			//check to see if the requested version has been downloaded before
-			installLocation := GetInstallLocation(installPath)
-			installFileVersionPath := ConvertExecutableExt(filepath.Join(installLocation, product.GetVersionPrefix()+requestedVersion))
-			recentDownloadFile := CheckFileExist(installFileVersionPath)
-			if recentDownloadFile {
-				ChangeProductSymlink(product, installFileVersionPath, customBinaryPath)
-				logger.Infof("Switched %s to version %q", product.GetName(), requestedVersion)
-				addRecent(requestedVersion, installPath, product) //add to recent file for faster lookup
-				return nil
-			}
-
-			// If the requested version had not been downloaded before
-			// Set list all true - all versions including beta and rc will be displayed
-			tflist, _ := getTFList(mirrorURL, true)         // Get list of versions
-			exist := versionExist(requestedVersion, tflist) // Check if version exists before downloading it
-
-			if exist {
-				install(product, requestedVersion, customBinaryPath, installPath, mirrorURL)
-			} else {
-				return fmt.Errorf("the provided terraform version does not exist: %q.\n Try `tfswitch -l` to see all available versions", requestedVersion)
-			}
+			return install(product, requestedVersion, customBinaryPath, installPath, mirrorURL)
 		} else {
 			PrintInvalidTFVersion()
 			UsageMessage()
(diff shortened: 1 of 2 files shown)


yermulnik:
@PierreTechoueyres Thank you for the contribution.
Would you mind elaborating on details of what and how this PR resolves?
Thanks.

PierreTechoueyres:
Without b86f96b8858aa1f9b65fa1968cf5536fea1d6a78 the binary is always installed in %USERPROFILE%\bin whatever you setup in your ` .tfswitch.toml`.

For example I have an ` .tfswitch.toml` like the one bellow
```
bin = "C:/Users/MyName/go/bin"
version = "1.8.4"
```

And without my patch, tfswitch always install the terraform binary in `C:/Users/MyName/bin`.

PierreTechoueyres:
I think that my commit b86f96b8858aa1f9b65fa1968cf5536fea1d6a78 could be simplified. 
```diff
-		binPathWritable := false //assume bin path is not writable
-		if runtime.GOOS != "windows" {
-			binPathWritable = CheckDirWritable(binDir) //check if is writable on ( only works on LINUX)
-		}
+		binPathWritable := runtime.GOOS == "windows" || CheckDirWritable(binDir) //check if is writable on ( only works on LINUX)
```
 ##What do you think  of it ?

yermulnik:
> Without [b86f96b](https://github.com/warrensbox/terraform-switcher/commit/b86f96b8858aa1f9b65fa1968cf5536fea1d6a78) the binary is always installed in %USERPROFILE%\bin whatever you setup in your ` .tfswitch.toml`.

This should be presumably fixed in the `v1.2.0-alpha` (and consequently in the nearest upcoming release): https://github.com/warrensbox/terraform-switcher/releases/tag/v1.2.0-alpha — can you please try this alpha and see whether `bin` value from TOML is respected correctly?
Thanks.

PierreTechoueyres:
I've already checked with latest mater (6b0b1901ddb37319b018dbce44d48f3943f3f2b1) and created my patch upon the code at it.

yermulnik:
Ah, I see 👍🏻 
* Please revert logging additions and changes as they partially overlap with #464 and partially add excessive logging I guess.
* Also do I get it right that you're suggesting to assume `bin` path on Windows is writable unconditionally (as opposite to existing logic that assumes Windows paths are not writable by default):  https://github.com/warrensbox/terraform-switcher/pull/468/files#diff-90697d6eae2e2416dcb80680a07d02ed87ac625142983a3f4ce392f236f4be54R163 — did you have  a chance to validate this change for paths outside your home directory on Windows? (unfortunately we have limited capacity to verify all of the changes on Windows 😢)
  * I suppose that the "_custom bin path is not writable on Windows_" bit was added on purpose and hence to allow custom `bin` paths on Windows we probably need to go a bit further and implement something like this https://stackoverflow.com/a/76285404/5093149 so that `tfswitch` doesn't just assume custom `bin` path is writable, but also ensures it is indeed writable before proceeding 🤔 

Cc: @warrensbox @MatthewJohn @MatrixCrawler

PierreTechoueyres:
>     * Please revert logging additions and changes as they partially overlap with [Add debug logging when successfully obtaining parameter from location… #464](https://github.com/warrensbox/terraform-switcher/pull/464) and partially add excessive logging I guess.

Done.

>     * Also do I get it right that you're suggesting to assume `bin` path on Windows is writable unconditionally (as opposite to existing logic that assumes Windows paths are not writable by default):  https://github.com/warrensbox/terraform-switcher/pull/468/files#diff-90697d6eae2e2416dcb80680a07d02ed87ac625142983a3f4ce392f236f4be54R163 — did you have  a chance to validate this change for paths outside your home directory on Windows? (unfortunately we have limited capacity to verify all of the changes on Windows 😢)

No, I didn't validate that a path outside the home directory on Windows was writable. *But* I think it's user responsibility to check he has the write access on this directory.
I admit this isn't an ideal situation. But the other way is worse. Currently the option is simply ignored on Window, with this patch it let you a chance to use it.
       
>       * I suppose that the "_custom bin path is not writable on Windows_" bit was added on purpose and hence to allow custom `bin` paths on Windows we probably need to go a bit further and implement something like this https://stackoverflow.com/a/76285404/5093149 so that `tfswitch` doesn't just assume custom `bin` path is writable, but also ensures it is indeed writable before proceeding 🤔
That's would definitely be a better way (and an OS agnostic one).
I see there is already an implementation of `CheckDirWritable` in `lib/dir_perm_windows.go`. Do you know why it is discarded in current code ?

PierreTechoueyres:
I've made an attempt with the informations from (https://stackoverflow.com/a/76285404/5093149) on branch [PierreTechoueyres:fix-userBinPath-on-windows-1](https://github.com/PierreTechoueyres/terraform-switcher/commit/f11f75a14c3f25429b1fb4ea1f7dd6bdc4f32846).
With it and an unwritable path in my `~/.tfswitch.toml` I obtain :

> INFO Reading configuration from "C:\\Users\\MyName\\.tfswitch.toml"
> ERROR Could not create file "tmpfile" on "C:\\Program Files"
> INFO Installing Terraform at "C:\\Users\\MyName\\bin"
> INFO Switched Terraform to version "1.8.4"

Whereas with the actual PR it shows

> INFO Reading configuration from "C:\\Users\\MyName\\.tfswitch.toml"
> INFO Found binary path "C:/Program Files/terraform"
> INFO Installing Terraform at "C:/Program Files/terraform"
> FATAL Could not create target binary: C:\Program Files\terraform.exe
> exit status 1
> 

I guess it's a better solution. Should I use it ?

MatthewJohn:
> I've made an attempt with the informations from (https://stackoverflow.com/a/76285404/5093149) on branch [PierreTechoueyres:fix-userBinPath-on-windows-1](https://github.com/warrensbox/terraform-switcher/compare/master...PierreTechoueyres:terraform-switcher:fix-userBinPath-on-windows-1). With it and an unwritable path in my `~/.tfswitch.toml` I obtain :
> 
> > INFO Reading configuration from "C:\Users\MyName\.tfswitch.toml"
> > ERROR Could not create file "tmpfile" on "C:\Program Files"
> > INFO Installing Terraform at "C:\Users\MyName\bin"
> > INFO Switched Terraform to version "1.8.4"
> 
> Whereas with the actual PR it shows
> 
> > INFO Reading configuration from "C:\Users\MyName\.tfswitch.toml"
> > INFO Found binary path "C:/Program Files/terraform"
> > INFO Installing Terraform at "C:/Program Files/terraform"
> > FATAL Could not create target binary: C:\Program Files\terraform.exe
> > exit status 1
> 
> I guess it's a better solution. Should I use it ?

@PierreTechoueyres I do prefer the idea of still checking permissions.
I seems weird that the permissions _can't_ be checked without writing a file.
Do you know what was going wrong (I assume https://github.com/warrensbox/terraform-switcher/compare/master...PierreTechoueyres:terraform-switcher:fix-userBinPath-on-windows-1#diff-a0b72d66e4d7d196755924401e35084604ef682057492feed9cb18b8199e2c5fL22 wasn't working correctly?)

PierreTechoueyres:
I think we don't understand each other.
The branch [fix-userBinPath-on-windows-1](https://github.com/PierreTechoueyres/terraform-switcher/tree/fix-userBinPath-on-windows-1) has an write access check before using the custom path.
Hence the output:
> INFO Reading configuration from "C:\Users\MyName.tfswitch.toml"
> ERROR Could not create file "tmpfile" on "C:\Program Files"
> INFO Installing Terraform at "C:\Users\MyName\bin"
> INFO Switched Terraform to version "1.8.4"

Whereas the branch [fix-userBinPath-on-windows](https://github.com/PierreTechoueyres/terraform-switcher/tree/fix-userBinPath-on-windows) hasn't this check and fail if the directory isn't writable:
> INFO Reading configuration from "C:\Users\MyName.tfswitch.toml"
> INFO Found binary path "C:/Program Files/terraform"
> INFO Installing Terraform at "C:/Program Files/terraform"
> FATAL Could not create target binary: C:\Program Files\terraform.exe
> exit status 1

Both are compromises.
[fix-userBinPath-on-windows-1](https://github.com/PierreTechoueyres/terraform-switcher/tree/fix-userBinPath-on-windows-1) does the check with an attempt to write in the directory, [fix-userBinPath-on-windows](https://github.com/PierreTechoueyres/terraform-switcher/tree/fix-userBinPath-on-windows) doesn't this check and fail.

For your comment 
> I seems weird that the permissions can't be checked without writing a file.
I agree with you **but** checking the permission on a directory is really tricky especially on Windows with the ACLs. This sole point could be the subject for an whole lib at all.

For now I think I prefer the [fix-userBinPath-on-windows-1](https://github.com/PierreTechoueyres/terraform-switcher/tree/fix-userBinPath-on-windows-1)  branch.

> The original code looks very similar to https://stackoverflow.com/a/49148866 - is there any debugging that can be done to check why it's not working as expected?
The following points pop immediately on my mind but I'm certain there are others:
- the ACLs,
-  the fact that's owner access that's checked and not the user access.

MatthewJohn:
> I think we don't understand each other.

I assume this is referring to my:

> @PierreTechoueyres I do prefer the idea of still checking permissions.
I seems weird that the permissions can't be checked without writing a file.

?

If so, I think we do (or at least, I'm think understanding what you're saying).. what I mean is: I'm not too keen on this PR, removing the check entirely for windows. I agree with you that I prefer the other (writing the temp dir to check for access).
My comment was just that it seems strange that there _isn't_ a way to reliable check write access without actually writing a file.

I do wonder if the following could be done instead (obviously relatively pseudocode):
```
func ChangeProductSymlink(product Product, binVersionPath string, userBinPath string) error {
  var possibleInstallLocations := {userBinPath, filePath.join(binDir, "bin")}
  var err error
  for pathItx := range possibleInstallLocations {
    err = CreateSymlink(binVersionPath, pathItx)
    if err == nil {
      break
    }
  }
  return err
}
``` 
So the "test" of whether we can write to the directory is the act of adding the symlink itself, rather than trying to evaluate the paths beforehand. Then we don't have to write test files on every execution.

and then update `install` to use the `ChangeProductSymlink` method, rather than re-inventing the wheel.

yermulnik:
> what I mean is: I'm not too keen on this PR, removing the check entirely for windows.

+1

PierreTechoueyres:
@MatthewJohn: New version with your idea.
Work on Linux, will check tomorrow on Windows.

lib/symlink.go
```diff
@@ -93,21 +98,32 @@ func CheckSymlink(symlinkPath string) bool {
 // Deprecated: This function has been deprecated in favor of ChangeProductSymlink and will be removed in v2.0.0
 func ChangeSymlink(binVersionPath string, binPath string) {
 	product := getLegacyProduct()
-	ChangeProductSymlink(product, binVersionPath, binPath)
+	_ = ChangeProductSymlink(product, binVersionPath, binPath)
```
MatthewJohn, line 101: Since it historically used log.Fatal in this method, since it was never capable of returning an error, in this version of the method, could we take the error and do a log.Fatal?

lib/symlink.go
```diff
@@ -32,45 +35,47 @@ func CreateSymlink(cwd string, dir string) {
...
 }
 
 // RemoveSymlink : remove symlink
-func RemoveSymlink(symlinkPath string) {
+func RemoveSymlink(symlinkPath string) error {
 
 	_, err := os.Lstat(symlinkPath)
 	if err != nil {
-		logger.Fatalf(`
+		logger.Tracef(`
```
MatthewJohn, line 56: I think Debug has been preferred for these types of messages. Unless there's a specific reason that Trace is better for this type of message?

lib/install.go
```diff
@@ -51,17 +50,31 @@ func GetInstallLocation(installPath string) string {
...
+		return nil
+	}
+
+	// If the requested version had not been downloaded before
+	// Set list all true - all versions including beta and rc will be displayed
+	tflist, _ := getTFList(mirrorURL, true)  // Get list of versions
+	exist := versionExist(tfversion, tflist) // Check if version exists before downloading it
+
+	if !exist {
+		return fmt.Errorf("the provided terraform version does not exist: %q.\n Try `tfswitch -l` to see all available versions", tfversion)
```
MatthewJohn, line 76:
```suggestion
		return fmt.Errorf("the provided %s version does not exist: %q.\n Try `tfswitch -l` to see all available versions", product.GetId(), tfversion)
```

lib/install.go
```diff
@@ -51,17 +50,31 @@ func GetInstallLocation(installPath string) string {
...
-	binPath = installableBinLocation(product, binPath)
 
-	initialize(binPath)                               //initialize path
-	installLocation = GetInstallLocation(installPath) //get installation location -  this is where we will put our terraform binary file
+	//check to see if the requested version has been downloaded before
+	installLocation := GetInstallLocation(installPath)
+	installFileVersionPath := ConvertExecutableExt(filepath.Join(installLocation, product.GetVersionPrefix()+tfversion))
+	recentDownloadFile := CheckFileExist(installFileVersionPath)
+	if recentDownloadFile {
+		err := ChangeProductSymlink(product, installFileVersionPath, binPath)
```
MatthewJohn, line 61: This logic appears identical to L113, I wonder if it worth removing the duplication, e.g.
```
func install() {
    installFileVersionPath := ConvertExecutableExt(filepath.Join(installLocation, product.GetVersionPrefix()+tfversion))
    recentDownloadFile := CheckFileExist(installFileVersionPath)
    if !recentDownloadFile {
        tfList := getTfList()
        ... remaining download logic
    }
    err := ChangeProductSymlink()
    ...
}
```

PierreTechoueyres: I've added `func switchToVersion(product Product, tfversion string, binPath string, installPath string, installFileVersionPath string) error`

PierreTechoueyres:
This work as expected on Windows too.

warrensbox (approved):

lib/symlink.go
```diff
@@ -1,27 +1,31 @@
...
+func CreateSymlink(cwd string, dir string) error {
 	// If we are on windows the symlink is not working correctly.
 	// Copy the desired terraform binary to the path environment.
 	if runtime.GOOS == "windows" {
 		r, err := os.Open(cwd)
 		if err != nil {
-			logger.Fatalf("Unable to open source binary: %s", cwd)
-			os.Exit(1)
+			logger.Debugf("Unable to open source binary: %s", cwd)
+			return errors.New(fmt.Sprintf("Unable to open source binary: %s", cwd))
```
yermulnik, line 21: Would it make sense to log and/or return `err` also?

lib/symlink.go
```diff
@@ -1,27 +1,31 @@
...
 		}
 		defer r.Close()
 
 		w, err := os.Create(dir + ".exe")
 		if err != nil {
-			logger.Fatalf("Could not create target binary: %s", dir+".exe")
-			os.Exit(1)
+			logger.Debugf("Could not create target binary: %s.exe", dir)
+			return errors.New(fmt.Sprintf("Could not create target binary: %s.exe", dir))
```
yermulnik, line 28: Would it make sense to log and/or return `err` also?

lib/symlink.go
yermulnik: Also why `errors.New(fmt.Sprintf(…))` rather than just `fmt.Errorf()`?
Ref: https://pkg.go.dev/errors#New

lib/symlink.go
yermulnik: Also why `errors.New(fmt.Sprintf(…))` rather than just `fmt.Errorf()`?
Ref: https://pkg.go.dev/errors#New

lib/symlink.go
```diff
@@ -32,45 +36,47 @@ func CreateSymlink(cwd string, dir string) {
...
-			logger.Fatalf(`
+			logger.Debugf(`
 		Unable to create new symlink.
 		Maybe symlink already exist. Try removing existing symlink manually.
 		Try running "unlink %s" to remove existing symlink.
 		If error persist, you may not have the permission to create a symlink at %s.
 		Error: %s
 		`, dir, dir, err)
-			os.Exit(1)
+			return errors.New(fmt.Sprintf("Unable to create new symlink %s : %s", dir, err))
```
yermulnik, line 46: Why `errors.New(fmt.Sprintf(…))` rather than just `fmt.Errorf()`?
Ref: https://pkg.go.dev/errors#New

lib/symlink.go
```diff
@@ -32,45 +36,47 @@ func CreateSymlink(cwd string, dir string) {
...
-		logger.Fatalf(`
+		logger.Debugf(`
 		Unable to stat symlink.
 		Maybe symlink already exist. Try removing existing symlink manually.
 		Try running "unlink %s" to remove existing symlink.
 		If error persist, you may not have the permission to create a symlink at %s.
 		Error: %s
 		`, symlinkPath, symlinkPath, err)
-		os.Exit(1)
+		return errors.New(fmt.Sprintf("Unable to stat symlink %s : %s", symlinkPath, err))
```
yermulnik, line 64: Why `errors.New(fmt.Sprintf(…))` rather than just `fmt.Errorf()`?
Ref: https://pkg.go.dev/errors#New

lib/symlink.go
yermulnik: Also please use `%q` for strings and `%v` for errors.

lib/symlink.go
yermulnik: Also please use `%q` for strings and `%v` for errors.

lib/symlink.go
```diff
@@ -32,45 +36,47 @@ func CreateSymlink(cwd string, dir string) {
...
-			logger.Fatalf(`
+			logger.Debugf(`
 			Unable to remove symlink.
 			Maybe symlink already exist. Try removing existing symlink manually.
 			Try running "unlink %s" to remove existing symlink.
 			If error persist, you may not have the permission to create a symlink at %s.
 			Error: %s
 			`, symlinkPath, symlinkPath, errRemove)
-			os.Exit(1)
+			return errors.New(fmt.Sprintf("Unable to remove symlink %s : %s", symlinkPath, err))
```
yermulnik, line 76: Why `errors.New(fmt.Sprintf(…))` rather than just `fmt.Errorf()`?
Also please use `%q` for strings and `%v` for errors.

lib/symlink.go
```diff
@@ -93,21 +99,43 @@ func CheckSymlink(symlinkPath string) bool {
...
+			}
 
-	/* remove current symlink if exist*/
-	symlinkExist := CheckSymlink(binPath)
-	if symlinkExist {
-		RemoveSymlink(binPath)
+			/* set symlink to desired version */
+			err = CreateSymlink(binVersionPath, location)
+			if err == nil {
+				logger.Debugf("Symlink created at %s", location)
```
yermulnik, line 127:
```suggestion
				logger.Debugf("Symlink created at %q", location)
```

yermulnik:
@PierreTechoueyres Also please merge latest from `master`:
image

PierreTechoueyres:
Is there something blocking the merge ?

warrensbox (approved):

warrensbox:
@PierreTechoueyres 
Can you help us test your changes and let us know if you see any error?

Please test v1.2.0-alpha2 on your windows machine.

https://github.com/warrensbox/terraform-switcher/releases/tag/v1.2.0-alpha2

lib/symlink.go
```diff
@@ -93,21 +93,42 @@ func CheckSymlink(symlinkPath string) bool {
...
+				return nil
+			}
+		}
 	}
 
-	/* set symlink to desired version */
-	CreateSymlink(binVersionPath, binPath)
+	if err == nil {
+		return fmt.Errorf("Unable to find existing directory in %q. %s",
+			strings. Join(possibleInstallLocations, " or "),
```
yermulnik, line 129: Oops

yermulnik: https://github.com/warrensbox/terraform-switcher/pull/478

PierreTechoueyres:
> @PierreTechoueyres Can you help us test your changes and let us know if you see any error?
> 
> Please test v1.2.0-alpha2 on your windows machine.
> 
> https://github.com/warrensbox/terraform-switcher/releases/tag/v1.2.0-alpha2

Sorry, didn't have time to check until now. v1.2.1 seem to work as expected.

Issue #480, opened 2024-07-06
file extension (.exe) on Windows doubled in version 1.2.0


mj790:
Version 1.2.0alpha2 seems ok but in the released version 1.2.0, the terraform.exe that's put into the bin directory on Windows ends up with a duplicate file extension (terraform.exe.exe)

Last couple lines from debug output show it should have done the right thing (ignoring the slashes in the wrong direction):

```
DEBUG [files.go:19,RenameFile] Renaming file "C:\\Users\\mj790\\.terraform.versions\\terraform.exe" to "C:\\Users\\mj790\\.terraform.versions\\terraform_1.3.2.exe"
DEBUG [symlink.go:118,ChangeProductSymlink] Symlink created at "C:\\Users\\mj790/bin/terraform.exe"
```

Based on the debug output and what I see in symlink.go, looks like the problem might be on line 23 with an .exe being added that is no longer required? (and associated error message on line 25 adding the same .exe)

PR #481, opened 2024-07-06
fix: Remove duplicate .exe extension added to paths for windows inside symlink, as this is already handled by ConvertExecutableExt

MatthewJohn:
Issue #480


MatthewJohn:
Screenshot from 2024-07-06 09-37-44
seems to have fixed it

warrensbox (approved):

Issue #505, opened 2024-09-18
tfswitch -U logs instead of returning version


brainsiq:
There appears to have been a change in the behaviour of `tfswitch -U`. It now returns a result via logs, which I understand from other issues appears to be a fairly recent change.

This means if you set the log level to error then you get no output at all.

```
❯ tfswitch -U --log-level=ERROR

```

```
❯ tfswitch -U --log-level=INFO
14:42:24.303 INFO 1.9.6
```

I was using `-U` to bootstrap new projects with the latest as `required_version`, which has stopped working, and needs to be updated to instead parse the output. This isn't a big problem but wanted to check if it is intended behaviour?

This also applies to other flags like `-S` and `-P` but isn't consistent with `-v` which returns raw output without logs.

MatthewJohn:
Hmm, I agree.

I think that if the desired output of an action is purely the output itself, it almost definitely shouldn’t be logged and should just be printed. I wonder whether the logging goes to stdout or stderr! 🤔

Matt

brainsiq:
I tested that and it's definitely stdout.

```
tfswitch -U > program.stdout 2> program.stderr
```

The output is in the stdout file

yermulnik:
Yep, agree too. Need to reconfigure logging to send to stderr and output to stdout only what's requested for `--show-*` flags.
I wish we've got more contributions as we seem to be stagnating on development again 😢

MatthewJohn:
@yermulnik @brainsiq I've had an initial go at this in https://github.com/warrensbox/terraform-switcher/pull/506 :)

brainsiq:
Sorry for the lack of feedback, but I d/l the latest version of tswitch and the functionality now looks good to me. Not familiar with golang but was able to understand the code change too 👍

yermulnik:
@brainsiq Thank you for the feedback 👍🏻

Issue #462, opened 2024-06-13
[Feature request] Disable color logs when not a terminal


yermulnik:
**Is your feature request related to a problem? Please describe.**
At the moment `tfswitch` outputs logs with colors no matter whether it is run interactively or from within automation. Along with this when running within CI/CD or when `tfswitch` output is piped to another command or is written to a log-file it may be desired to log to console w/o colors to avoid vomiting console with escape sequences.

**Describe the solution you'd like**
* Default behavior: color output (same as it is at the moment)
* Non-terminal (non-interactive): force no color (and add `--force-color` cmdline arg) to switch off automatically **OR** add `--no-color` cmdline arg to let user decide explicitly per use-case
  * Add respective TOML config option for whichever option is considered

💡 Forcing color in non-interactive mode might be preferred e.g. when running within CI/CD pipeline that supports rendering ANSI escape sequences (e.g. https://plugins.jenkins.io/ansicolor/).

**Additional context**
Refs:
* https://github.com/warrensbox/terraform-switcher/blob/master/lib/logging.go#L22
* https://github.com/mattn/go-isatty#usage

MatthewJohn:
I guess this could be done at the same time as #451

yermulnik:
> I guess this could be done at the same time as #451

"at the same time" as in you could be working on both? 🙏🏻 
I think we should try and split different features into different PRs.

yermulnik:
Another use case: https://github.com/warrensbox/terraform-switcher/discussions/474

PR #506, opened 2024-09-18
Split output and logs

MatthewJohn:
Issue #505

### DIFF
--- a/lib/logging.go
+++ b/lib/logging.go
@@ -1,9 +1,11 @@
 package lib
 
 import (
+	"os"
+
+	"github.com/gookit/color"
 	"github.com/gookit/slog"
 	"github.com/gookit/slog/handler"
-	"os"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


lib/logging.go
```diff
@@ -17,6 +19,22 @@ var (
 	TraceLogging         = slog.Levels{slog.PanicLevel, slog.FatalLevel, slog.ErrorLevel, slog.WarnLevel, slog.InfoLevel, slog.NoticeLevel, slog.DebugLevel, slog.TraceLevel}
 )
 
+func NewStderrConsoleWithLF(lf slog.LevelFormattable) *handler.ConsoleHandler {
+	h := handler.NewIOWriterWithLF(os.Stderr, lf)
+
+	// default use text formatter
+	f := slog.NewTextFormatter()
+	// default enable color on console
+	f.WithEnableColor(color.SupportColor())
```
yermulnik, line 28: Not sure how this works as it still produces color output for me when redirecting to a file which in general should not happen with color output.
Btw we have a feature request to enable/disable color based on condition: https://github.com/warrensbox/terraform-switcher/issues/462 =)

yermulnik (commented):
Perfect. LGTM. Works for me. Though since it's sort of essential change in existing behavior, I'd summon @warrensbox to have him approve it.

lib/logging.go
MatthewJohn: Were you hoping anything to be done in this PR or wait for #462  - as suggested, this is copied from upstream for now, to just change the output :)

yermulnik: Nope =) PR looks (and works) good to me. Just need Warren to review and approve I guess.

warrensbox (approved):

PR #527, opened 2024-12-17
feat: Override default build format to Zip when building for Windows

felipebraga:
Override default build format to Zip when building for Windows

The idea behind it is to enable winget installation due the inability of winget decompress tar files.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -4,6 +4,8 @@ version: 2
 env:
   - CGO_ENABLED=0 # Build statically linked binaries
 
+
+
 builds:
   -
     ldflags:
@@ -37,6 +39,9 @@ archives:
       {{- .Os }}_
       {{- .Arch }}
       {{- if .Arm }}v{{ .Arm }}{{ end -}}
+    format_overrides:
+      - goos: windows
+        format: zip
 
 brews:
   -


.goreleaser.yml
```diff
@@ -4,6 +4,8 @@ version: 2
 env:
   - CGO_ENABLED=0 # Build statically linked binaries
 
+
+
```
yermulnik, line 8: Out of interest: why these two new blank lines?

yermulnik: other than than PR looks good to me

yermulnik (commented):
```shell
> git br
* pr/527
  master

> goreleaser check
  • checking                                 path=.goreleaser.yml
  • 1 configuration file(s) validated
  • thanks for using goreleaser!
```

yermulnik:
@felipebraga While you're here, could you please trim trailing whitespaces [in this file on line 29](https://github.com/felipebraga/terraform-switcher/blob/master/.goreleaser.yml#L29)? Thanks in advance.
image

.goreleaser.yml
yermulnik: @felipebraga ?

MatthewJohn:
Assuming that windows is in a state that can run shell scripts (I guess cygwin or such), then we need to consider the install script: https://github.com/warrensbox/terraform-switcher/blob/master/install.sh#L345 as it looks for the tar.gz

yermulnik:
> Assuming that windows is in a state that can run shell scripts (I guess cygwin or such), then we need to consider the install script: [`master`/install.sh#L345](https://github.com/warrensbox/terraform-switcher/blob/master/install.sh?rgh-link-date=2024-12-18T20%3A23%3A50Z#L345) as it looks for the tar.gz

Ah, good pointer.
Given the script already supports `zip`, we should be good to adjust `FORMAT` conditionally via https://github.com/warrensbox/terraform-switcher/blob/master/install.sh#L89-L92 like it's already done somewhere else, e.g. https://github.com/LilithGames/golangci-lint/blob/master/install.sh#L120-L126

The change looks quite simple, though I don't have Windows laptop to test 🤷🏻 

@felipebraga Could you please implement this change and test `install.sh` script to work as expected on Windows? 🤔

yermulnik (changes requested):
Requesting changes as install script needs to be adjusted accordingly.

MatthewJohn:
> > Assuming that windows is in a state that can run shell scripts (I guess cygwin or such), then we need to consider the install script: [`master`/install.sh#L345](https://github.com/warrensbox/terraform-switcher/blob/master/install.sh?rgh-link-date=2024-12-18T20%3A23%3A50Z#L345) as it looks for the tar.gz
> 
> Ah, good pointer. Given the script already supports `zip`, we should be good to adjust `FORMAT` conditionally via https://github.com/warrensbox/terraform-switcher/blob/master/install.sh#L89-L92

I did see that, the comment seems weird though: `# change format (tar.gz or zip) based on ARCH`

Seems to suggest that format can be changed based on architecture, but archive name can be changed based on OS 🤔

yermulnik:
> I did see that, the comment seems weird though: `# change format (tar.gz or zip) based on ARCH`
> 
> Seems to suggest that format can be changed based on architecture, but archive name can be changed based on OS 🤔
> 
> Edit: Ah yeh, that second example implies it should work fine 🥳

Yeah, that seems to be a copy-pasta typo that comes from someone else's original script I guess that is used across multiple repos and is based on portable POSIX libs collection at https://github.com/client9/shlib

.goreleaser.yml
yermulnik: Resolved via 990f20b7cad3a42c2dc9703022cd879b34e61949

yermulnik (approved):
LGTM:
```shell
> goreleaser check
  • checking                                 path=.goreleaser.yml
  • 1 configuration file(s) validated
  • thanks for using GoReleaser!

> RELEASE_VERSION=testing goreleaser release --snapshot --clean 2>&1 | fgrep .zip
    • archiving                                      name=dist/terraform-switcher_testing_windows_armv6.zip
    • archiving                                      name=dist/terraform-switcher_testing_windows_armv7.zip
    • archiving                                      name=dist/terraform-switcher_testing_windows_amd64.zip
    • archiving                                      name=dist/terraform-switcher_testing_windows_386.zip
```

ℹ️ Multi-labeler failure isn't critical and I can't figure why (maybe because it's me pushing to PR author's repo 🤷🏻)

PR #532, opened 2025-01-14
feat: Allow to download binary with custom CPU arch

### LINKED ISSUE #318: Ability to specify the terraform architecture to install on darwin platform
The terraform configuration used by my organisation relies on a deprecated provider that only provides binary in `darwin_amd64` architecture.  
During the upgrade process to make our code ARM compatible, I was testing the code on my local machine to ensure the existing code is known-good.  
To do that I fetched the code, used tfswitch to switch to the intended version and ran Terraform init.  

I then discovered while tfswitch is capable of identifying the platform that I'm on (`darwin_arm64`) and fetches the corresponding binary, I was not able to override this behaviour to force it to fetch `darwin_amd64`.  

Being unable to fetch `darwin_amd64` means I have to manually download the binary and move it into place before I can resume testing. 

tfswitch should accept an additional optional argument along with version, when supplied tfswitch should override the default architecture decision and install the specified architecture instead. 

For example:
```
tfswitch 1.5.6 darwin_amd64
```

I have manually downloaded and linked the binary on my local machine.  
This is the method I'd use without this feature in place, and there is no alternatives right now. 

**Additional context**
Mac OS on ARM is capable of running `darwin_amd64` binaries (with Rosetta 2 translation).  
Terraform fetches provider based on it's own architecture, which means Terraform code that uses providers with only `darwin_amd64` binary can still be executed on Mac, as long as Terraform binary is also running on `darwin_amd64` architecture.

yermulnik:
Fixes #318

Note: code in changed files is auto-formatted

Caveats:
- Supports overriding arch via env var and TOML config
- Doesn't override if binary had already been downloaded before
  - Remove existing binary from `tfswitch` download dir manually (see `--install` option)
- Spits out Warn log message if requested arch doesn't match actual
- Doesn't check arch value validity and just passes it to download function directly, which will fail if download file doesn't exist on remote

Usage example: `tfswitch --arch arm64 1.10.0`
image
OpenTofu (for completeness):
image

### DIFF
--- a/lib/install.go
+++ b/lib/install.go
@@ -13,9 +13,7 @@ import (
 	"github.com/hashicorp/go-version"
 )
 
-var (
-	installLocation = "/tmp"
-)
+var installLocation = "/tmp"
 
 // initialize : removes existing symlink to terraform binary based on provided binPath
 func initialize(binPath string) {
@@ -50,10 +48,10 @@ func GetInstallLocation(installPath string) string {
 }
 
 // install : install the provided version in the argument
-func install(product Product, tfversion string, binPath string, installPath string, mirrorURL string) error {
+func install(product Product, tfversion, binPath, installPath, mirrorURL, goarch string) error {
 	var wg sync.WaitGroup
 
-	//check to see if the requested version has been downloaded before
+	// check to see if the requested version has been downloaded before
 	installLocation := GetInstallLocation(installPath)
 	installFileVersionPath := ConvertExecutableExt(filepath.Join(installLocation, product.GetVersionPrefix()+tfversion))
 	recentDownloadFile := CheckFileExist(installFileVersionPath)
@@ -70,7 +68,10 @@ func install(product Product, tfversion string, binPath string, installPath stri
 		return fmt.Errorf("the provided %s version does not exist: %q.\n Try `tfswitch -l` to see all available versions", product.GetId(), tfversion)
 	}
 
-	goarch := runtime.GOARCH
+	if goarch != runtime.GOARCH {
+		logger.Warnf("Installing for %q architecture on %q!", goarch, runtime.GOARCH)
+	}
+
 	goos := runtime.GOOS
 
 	// Terraform darwin arm64 comes with 1.0.2 and next version
@@ -114,7 +115,7 @@ func switchToVersion(product Product, tfversion string, binPath string, installP
 	}
 
 	logger.Infof("Switched %s to version %q", product.GetName(), tfversion)
-	addRecent(tfversion, installPath, product) //add to recent file for faster lookup
+	addRecent(tfversion, installPath, product) // add to recent file for faster lookup
 	return nil
 }
 
@@ -135,24 +136,24 @@ func ConvertExecutableExt(fpath string) string {
 // If not, create $HOME/bin. Ask users to add  $HOME/bin to $PATH and return $HOME/bin as install location
 // Deprecated: This function has been deprecated and will be removed in v2.0.0
 func installableBinLocation(product Product, userBinPath string) string {
-	homedir := GetHomeDirectory()         //get user's home directory
-	binDir := Path(userBinPath)           //get path directory from binary path
-	binPathExist := CheckDirExist(binDir) //the default is /usr/local/bin but users can provide custom bin locations
+	homedir := GetHomeDirectory()         // get user's home directory
+	binDir := Path(userBinPath)           // get path directory from binary path
+	binPathExist := CheckDirExist(binDir) // the default is /usr/local/bin but users can provide custom bin locations
 
-	if binPathExist { //if bin path exist - check if we can write to it
+	if binPathExist { // if bin path exist - check if we can write to it
 
-		binPathWritable := false //assume bin path is not writable
+		binPathWritable := false // assume bin path is not writable
 		if runtime.GOOS != "windows" {
-			binPathWritable = CheckDirWritable(binDir) //check if is writable on ( only works on LINUX)
+			binPathWritable = CheckDirWritable(binDir) // check if is writable on ( only works on LINUX)
 		}
 
 		// IF: "/usr/local/bin" or `custom bin path` provided by user is non-writable, (binPathWritable == false), we will attempt to install terraform at the ~/bin location. See ELSE
 		if !binPathWritable {
 			homeBinDir := filepath.Join(homedir, "bin")
-			if !CheckDirExist(homeBinDir) { //if ~/bin exist, install at ~/bin/terraform
+			if !CheckDirExist(homeBinDir) { // if ~/bin exist, install at ~/bin/terraform
 				logger.Noticef("Unable to write to %q", userBinPath)
 				logger.Infof("Creating bin directory at %q", homeBinDir)
-				createDirIfNotExist(homeBinDir) //create ~/bin
+				createDirIfNotExist(homeBinDir) // create ~/bin
 				logger.Warnf("Run `export PATH=\"$PATH:%s\"` to append bin to $PATH", homeBinDir)
 			}
 			logger.Infof("Installing %s at %q", product.GetName(), homeBinDir)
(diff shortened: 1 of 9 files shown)


yermulnik:
@MatthewJohn While you're here, could you please review/approve?

lib/install.go
```diff
@@ -50,10 +48,10 @@ func GetInstallLocation(installPath string) string {
 }
 
 // install : install the provided version in the argument
-func install(product Product, tfversion string, binPath string, installPath string, mirrorURL string) error {
+func install(product Product, tfversion, binPath, installPath, mirrorURL, goarch string) error {
```
MatthewJohn, line 51: Maybe I've missed something (I don't follow golang), why has `string` type been removed?

lib/install.go
```diff
@@ -50,10 +48,10 @@ func GetInstallLocation(installPath string) string {
...
 	var wg sync.WaitGroup
 
-	//check to see if the requested version has been downloaded before
+	// check to see if the requested version has been downloaded before
```
MatthewJohn, line 54: 🎉

lib/install.go
```diff
@@ -114,7 +115,7 @@ func switchToVersion(product Product, tfversion string, binPath string, installP
 	}
 
 	logger.Infof("Switched %s to version %q", product.GetName(), tfversion)
-	addRecent(tfversion, installPath, product) //add to recent file for faster lookup
+	addRecent(tfversion, installPath, product) // add to recent file for faster lookup
```
MatthewJohn, line 118: Not entirely sure why it's inline
```suggestion

	// add to recent file for faster lookup
	addRecent(tfversion, installPath, product)
```

lib/install.go
```diff
@@ -171,38 +172,38 @@ func installableBinLocation(product Product, userBinPath string) string {
...
-func InstallLatestVersion(dryRun bool, customBinaryPath, installPath string, mirrorURL string) {
+func InstallLatestVersion(dryRun bool, customBinaryPath, installPath, mirrorURL, arch string) {
 	product := getLegacyProduct()
-	InstallLatestProductVersion(product, dryRun, customBinaryPath, installPath, mirrorURL)
+	InstallLatestProductVersion(product, dryRun, customBinaryPath, installPath, mirrorURL, arch)
 }
 
 // InstallLatestProductVersion install latest stable tf version
-func InstallLatestProductVersion(product Product, dryRun bool, customBinaryPath, installPath string, mirrorURL string) error {
+func InstallLatestProductVersion(product Product, dryRun bool, customBinaryPath, installPath, mirrorURL, arch string) error {
```
MatthewJohn, line 181: This is a public method change, so might need to be a major release?
I can't remember where we stood last time of public method signature changes? :/

MatthewJohn: If we're doing a major release, perhaps we can remove all of the non-`*Product*` method calls?

lib/param_parsing/environment_test.go
```diff
@@ -7,6 +7,19 @@ import (
...
 )
 
+func TestGetParamsFromEnvironment_arch_from_env(t *testing.T) {
+	logger = lib.InitLogger("DEBUG")
+	var params Params
+	expected := "amd64_from_env"
+	_ = os.Setenv("TF_ARCH", expected)
+	params = initParams(params)
+	params = GetParamsFromEnvironment(params)
+	_ = os.Unsetenv("TF_ARCH")
```
MatthewJohn, line 17: Maybe one for another time, but should this unset be deferred after the initial setting? This might avoid tests interferring with each other if this one fails before it is unset

lib/param_parsing/parameters_test.go
```diff
@@ -10,6 +10,19 @@ import (
 	"github.com/warrensbox/terraform-switcher/lib"
 )
 
+func TestGetParameters_arch_from_args(t *testing.T) {
+	expected := "amd64"
```
MatthewJohn, line 14:
```suggestion
	expected := "amd64_from_arg"
```
Could this be an arbitrary value to avoid it matching a default value from a real system?

MatthewJohn: Wait how does this test even work? The argument `amd64` is being supplied directly? Won't this be a positional argument and be used for the `version` argument and so the test is just validating that default argument?

lib/param_parsing/parameters_test.go
```diff
@@ -64,7 +83,7 @@ func TestGetParameters_params_are_overridden_by_toml_file(t *testing.T) {
 func TestGetParameters_toml_params_are_overridden_by_cli(t *testing.T) {
 	logger = lib.InitLogger("DEBUG")
 	expected := "../../test-data/integration-tests/test_tfswitchtoml"
-	os.Args = []string{"cmd", "--chdir=" + expected, "--bin=/usr/test/bin", "--product=terraform", "1.6.0"}
+	os.Args = []string{"cmd", "--chdir=" + expected, "--bin=/usr/test/bin", "--product=terraform", "--arch=amd64", "1.6.0"}
```
MatthewJohn, line 86: Could this be a unique argument value that won't match a real system?
```suggestion
	os.Args = []string{"cmd", "--chdir=" + expected, "--bin=/usr/test/bin", "--product=terraform", "--arch=amd64-from-args", "1.6.0"}
```

MatthewJohn (changes requested):

lib/install.go
yermulnik: `string` type is not removed. it's "optimized". in Go when you provide list of parameters for a function and if they are of the same type, you may specify type at the end of the list. so in this example the `product` is of type `Product` and the rest are of the type `string` since `string` is specified at the end of the string parameters list.

yermulnik: The guy explained it in details here: https://stackoverflow.com/a/76593363/5093149

lib/install.go
yermulnik: autformatted =)

lib/install.go
yermulnik: Please let me know if you're ok with this feature.

lib/install.go
yermulnik: This is part of autoformatter and is not actually part of my changeset. Do you really want me to re-format it to multiline?

lib/install.go
yermulnik: > This is a public method change

Ah my 😞 

> I can't remember where we stood last time of public method signature changes? :/

I don't remember too =(

> If we're doing a major release, perhaps we can remove all of the non-*Product* method calls?

As my Go skills are pretty much zero to nothing, I'd prefer a hand in such a task, which would be scope of a dedicated PR.

ps: as of recent development activity fade out, I'd really prefer to add this feature to make users happy, rather than postpone it till no one knows when to wait until we'll get together and do the clean up. but the point on major version bump is worthwhile. @warrensbox WDYT?

lib/param_parsing/environment_test.go
yermulnik: This test is copy-pasta from the one right down the code. So I'd leave this for another time please =)

lib/param_parsing/parameters_test.go
yermulnik: That's another copy-pasta test from the code right below it. With arbitrary value it was failing for me, so I set it to a correct value.
I'm a way far from being familiar with this testing stuff unfortunately. Would you be up to help reviewing and streamlining these tests please? (or I can remove this specific one and we can add it back another time)

lib/param_parsing/parameters_test.go
yermulnik: Let me adjust it.
Here it is: ad2fcee

lib/install.go
MatthewJohn: Yep, if it works, seems fine

lib/param_parsing/parameters_test.go
MatthewJohn: Yep, but the one below works because it's testing the position arg for version. I think you just need:
```
os.Args = []string{"cmd", "--arch=" + expected}
```

yermulnik: Ah, I see now. Thanks for the pointer 👍🏻 Addressed via a12fc57

lib/install.go
yermulnik: Addressed this too: 73af0e2

lib/install.go
yermulnik: @MatthewJohn https://github.com/warrensbox/terraform-switcher/discussions/452 — it looks like we haven't made it to anything definitive. Let's not block a feature at this moment because we couldn't make up a roadmap for the Public methods. WDYT?

yermulnik: Not sure whether this feature is worth of major bump. Maybe minor would be sufficient?
But can do major version bump as well 🤷🏻

yermulnik: If we decide to go with major version bump, maybe it's worth of adding also feature flag to customize OS for the download (windows, linux, darwin, and so forth) 🤔 I can't envision a real life use case for `tfswitch` to download binary for another OS from practical point of view, but maybe for completeness in addition to custom arch flag...

MatthewJohn: Technically, It's definitely a major bump - but only from a library point-of-view. So completely depends on what @warrensbox, I think. I know we saw a handful of repos using it as a library, it's just whether we (or Warren) want to support that in this :)

yermulnik: GH Copilot says we should be relatively safe updating this function:
image

I also checked this directly: https://github.com/search?q=github.com%2Fwarrensbox%2Fterraform-switcher+language%3AGo+&type=code — there seem to be no meaningful use cases across GitHub (and respectively I'm assuming similar with other VCS PaaSes) as most references are older forks of either `tfswitch` or `atlantis` (the latter doesn't use `tfswitcher` as library at the moment as far as I can see).

So we should be good to do a minor version bump at this time I guess.

ps: @warrensbox is hard to get hold of these days 😞 @MatthewJohn It's two of us here to decide I guess whether to add new feature or not.

MatthewJohn: Okay, let's go for it :)

MatthewJohn (approved):

lib/install.go
yermulnik: Appreciate it.

PR #534, opened 2025-01-20
docs: TOML file belongs in Home dir only

### LINKED ISSUE #533: tfswitch version 1.2.4 doesn't read .tfswitch.toml from current directory
Newest version of tfswitch doesn't read .tfswitch.toml from current working directory. It will only read if exist in home directory.

```
➜  /tmp ll tfswitch
-rwxr-xr-x  1 vinayak.krishna  wheel    11M Nov 23 04:19 tfswitch*

➜  /tmp cat .tfswitch.toml 
bin = "./terraform"
version = "1.10.4"

➜  /tmp rm ~/.tfswitch.toml
  
➜  /tmp ./tfswitch                                 
Select Terraform version
  ▸ 1.10.4 *recent
    1.10.3
    1.10.2
    1.10.1
↓   1.10.0

➜  /tmp cp .tfswitch.toml ~   

➜  /tmp ./tfswitch               
22:59:34.416 INFO Reading configuration from "/Users/vinayak.krishna/.tfswitch.toml"  
22:59:34.417 INFO Switched Terraform to version "1.10.4"  

➜  /tmp ll tfswitch terraform
lrwxr-xr-x  1 vinayak.krishna  wheel    59B Jan 20 22:59 terraform@ -> /Users/vinayak.krishna/.terraform.versions/terraform_1.10.4
-rwxr-xr-x  1 vinayak.krishna  wheel    11M Nov 23 04:19 tfswitch*

```

We use .tfswitch.toml in different directories to chose version per dir. This feature was working until version 1.1.1.

yermulnik:
Update documentation by dropping TOML mentions where it is irrelevant: TOML belongs in user Home directory only.

Closes #533

### DIFF
--- a/www/docs/usage/general.md
+++ b/www/docs/usage/general.md
@@ -3,12 +3,11 @@
 | Order | Method |
 | --- | ----------- |
 | 1 | `$HOME/.tfswitch.toml` (`version` parameter) |
-| 2 | `.tfswitch.toml` (`version` parameter) |
-| 3 | `.tfswitchrc` (version as a string) |
-| 4 | `.terraform-version` (version as a string) |
-| 5 | Terraform root module (`required_version` constraint) |
-| 6 | `terragrunt.hcl` (`terraform_version_constraint` parameter) |
-| 7 | Environment variable (`TF_VERSION`) |
+| 2 | `.tfswitchrc` (version as a string) |
+| 3 | `.terraform-version` (version as a string) |
+| 4 | Terraform root module (`required_version` constraint) |
+| 5 | `terragrunt.hcl` (`terraform_version_constraint` parameter) |
+| 6 | Environment variable (`TF_VERSION`) |
 
 With 1 being the **lowest** precedence and 7 — the **highest**  
 *(If you disagree with this order of precedence, please open an issue)*
--- a/www/docs/usage/ci-cd.md
+++ b/www/docs/usage/ci-cd.md
@@ -7,7 +7,7 @@
 ## Automatically switch with bash
 
 Add the following to the end of your `~/.bashrc` file:
-(Use either `.tfswitchrc` or `.tfswitch.toml` or `.terraform-version`)
+(Use either `.tfswitchrc` or `.terraform-version`)
 
 ```sh
 cdtfswitch(){
--- a/www/docs/usage/config-files.md
+++ b/www/docs/usage/config-files.md
@@ -7,12 +7,12 @@ tfswitch --chdir terraform_dir
 tfswitch -c terraform_dir
 ```
 
-## Use version.tf file
+## Use `version.tf` file
 
 If a .tf file with the terraform constraints is included in the current directory, it should automatically download or switch to that terraform version.  
 For example, the following should automatically switch terraform to the lastest version:  
 
-```
+```hcl
 terraform {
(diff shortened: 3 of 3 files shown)


www/docs/usage/general.md
```diff
@@ -3,12 +3,11 @@
...
-| 3 | `.tfswitchrc` (version as a string) |
-| 4 | `.terraform-version` (version as a string) |
-| 5 | Terraform root module (`required_version` constraint) |
-| 6 | `terragrunt.hcl` (`terraform_version_constraint` parameter) |
-| 7 | Environment variable (`TF_VERSION`) |
+| 2 | `.tfswitchrc` (version as a string) |
+| 3 | `.terraform-version` (version as a string) |
+| 4 | Terraform root module (`required_version` constraint) |
+| 5 | `terragrunt.hcl` (`terraform_version_constraint` parameter) |
+| 6 | Environment variable (`TF_VERSION`) |
```
MatthewJohn, line 10: This isn't valid anymore, Environment has highest precendence.
Could you change current `6` to `TF_DEFAULT_VERSION` and put a new item before (1) for `TF_VERSION`? :)

MatthewJohn (commented):
Looks good, but a minor thing, if that's okay :)

www/docs/usage/general.md
yermulnik: In the code I can see this order of precedence (the higher the number the higher precedence — this order of precedence is what is mentioned under the list):
1. TOML in HOME
2. `.tfswitchrc`
3. `.terraform-version`
4. Terraform root module (`required_version` constraint)
5. `terragrunt.hcl` (`terraform_version_constraint` parameter)
6. `TF_VERSION` env var

I'd suggest to not add `TF_DEFAULT_VERSION` to this list as it has its very own specific use case to provide fall-back for when TF version can't be determined by any other mean.

Screenshot to illustrate current state of this page:
image

MatthewJohn: Argh, the bottom is the highest priority :( Gets me every time! 😭

MatthewJohn: I would say it's still usefull to add.. for example in Ci/CD pipelines, it shows that it can be used as a fallback, which is still valid to document, no? But I think it's documented _somewhere_

MatthewJohn (approved):

www/docs/usage/general.md
yermulnik: > Argh, the bottom is the highest priority :( Gets me every time! 😭

Do you think it is worth of "reversing" the order to make `1` become highest precedence item? 🤔 

> I would say it's still usefull to add.. for example in Ci/CD pipelines, it shows that it can be used as a fallback

I don't think it's worth of adding this to **General** section. It belongs in CI/CD and has quite limited scope of use.
It also is documented at command line usage page. The documentation is pain and we need to revisit it for sure. Maybe when we have more active people joined the project to develop, review and approve as I also have quite limited time to dedicate to OSS 😢

yermulnik: > > Argh, the bottom is the highest priority :( Gets me every time! 😭
> 
> Do you think it is worth of "reversing" the order to make `1` become highest precedence item? 🤔

Let me know and I'll push another PR for this (I'm alright with both directions of the list though). Thanks.

Issue #543, opened 2025-02-17
Fresh install is failing to create new symlink. Creating `~/bin` is not being attempted.


n8felton:
This appears to be broken since v1.2.0. v1.1.1 was the last version that is creating the `bin` dir inside the user home directory 

Before I troubleshoot this further and attempt a PR, I want to ask the other contributors if this was an intended change or not. I see that `installableBinLocation` is deprecated, but I wanted to check if the logic of determining where to install is also deprecated, or if this is truly a bug. Is the goal to continue to install to `~/bin` if `/usr/local/bin` or the custom path is non-writable? 

https://github.com/warrensbox/terraform-switcher/blob/8fb671dde1259d0aa25bab890b45c983c8b64a87/lib/install.go#L137-L139

Possibly related: #493, #494 

v1.1.1
```
terraform-switcher_v1.1.1_darwin_arm64 % ./tfswitch -u --log-level TRACE
22:32:53.632 NOTICE [install.go:260,installableBinLocation] Unable to write to "/usr/local/bin/terraform"  
22:32:53.632 INFO [install.go:261,installableBinLocation] Creating bin directory at "/Users/n8felton/bin"  
22:32:53.632 INFO [files.go:78,createDirIfNotExist] Creating directory for terraform binary at "/Users/n8felton/bin"  
22:32:53.633 WARN [install.go:263,installableBinLocation] Run `export PATH="$PATH:/Users/n8felton/bin"` to append bin to $PATH  
22:32:53.681 INFO [install.go:89,install] Switched terraform to version "1.10.5"  
```

v1.3.0 
```
% tfswitch -u --log-level TRACE
22:23:18.275 DEBUG [versiontf.go:73,isTerraformModule] No required version constraints defined by Terraform module at "/Users/n8felton"  
22:23:18.355 INFO [files.go:103,createDirIfNotExist] Creating "/Users/n8felton/.terraform.versions" directory  
22:23:18.355 DEBUG [list_versions.go:41,getTFList] Getting list of versions  
22:23:18.388 DEBUG [download.go:134,downloadPublicKey] Looking up public key file at "/Users/n8felton/.terraform.versions/terraform_72D7468F.asc"  
22:23:18.388 INFO [download.go:99,downloadFromURL] Downloading to "/Users/n8felton/.terraform.versions/pgp-key.txt"  
22:23:18.484 INFO [download.go:128,downloadFromURL] 7717 bytes downloaded  
[...]

PR #544, opened 2025-02-22
fix: Create home bin directory if it does not already exist

MatthewJohn:
Issue #543

### DIFF
--- a/lib/symlink.go
+++ b/lib/symlink.go
@@ -101,12 +101,24 @@ func ChangeSymlink(binVersionPath string, binPath string) {
 func ChangeProductSymlink(product Product, binVersionPath string, userBinPath string) error {
 	homedir := GetHomeDirectory() // get user's home directory
 	homeBinPath := filepath.Join(homedir, "bin", product.GetExecutableName())
-	possibleInstallLocations := []string{userBinPath, homeBinPath}
+	// List of possible directories with boolean property as to whether to attempt to create
+	possibleInstallLocations := map[string]bool{
+		userBinPath: false,
+		homeBinPath: true,
+	}
 	possibleInstallDirs := []string{}
 	var err error
 
-	for _, location := range possibleInstallLocations {
+	for location, shouldCreate := range possibleInstallLocations {
 		possibleInstallDirs = append(possibleInstallDirs, Path(location))
+		// If directory does not exist, check if we should create it, otherwise skip
+		if !CheckDirExist(Path(location)) {
+			if shouldCreate {
+				os.MkdirAll(Path(location), 0755)
+			} else {
+				continue
+			}
+		}
 		if CheckDirExist(Path(location)) {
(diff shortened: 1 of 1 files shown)


lib/symlink.go
```diff
@@ -101,12 +101,24 @@ func ChangeSymlink(binVersionPath string, binPath string) {
...
 		possibleInstallDirs = append(possibleInstallDirs, Path(location))
+		// If directory does not exist, check if we should create it, otherwise skip
+		if !CheckDirExist(Path(location)) {
+			if shouldCreate {
+				os.MkdirAll(Path(location), 0755)
+			} else {
+				continue
+			}
+		}
 		if CheckDirExist(Path(location)) {
```
yermulnik, line 122: This `if CheckDirExist(Path(location)) {` conditional seems to make no sense with addition of `if !CheckDirExist(Path(location)) {` block above and we can use its inner code unconditionally I guess. WDYT?

lib/symlink.go
```diff
@@ -101,12 +101,24 @@ func ChangeSymlink(binVersionPath string, binPath string) {
...
 	possibleInstallDirs := []string{}
 	var err error
 
-	for _, location := range possibleInstallLocations {
+	for location, shouldCreate := range possibleInstallLocations {
 		possibleInstallDirs = append(possibleInstallDirs, Path(location))
+		// If directory does not exist, check if we should create it, otherwise skip
+		if !CheckDirExist(Path(location)) {
+			if shouldCreate {
+				os.MkdirAll(Path(location), 0755)
```
yermulnik, line 117: I think it's worth of adding diagnostic messages, same way that it exists in `createDirIfNotExist()` (could've used that function instead if it wouldn't be `panic`ing on error (I'm assuming `logger.Panicf()` prints error and exists immediately — right?) 🤷🏻):
```suggestion
                logger.Infof("Creating %q directory", dir)
				err = os.MkdirAll(Path(location), 0755)
				if err != nil {
					logger.Infof("Unable to create %q directory: %v", dir, err)
					continue
				}
```
And additionally a historical message with suggestion to update `PATH` env var right under `logger.Debugf("Symlink created at %q", location)`:
```go
			err = CreateSymlink(binVersionPath, location)
			if err == nil {
				logger.Warnf("Run `export PATH=\"$PATH:%s\"` to append %q to $PATH", location, location)
				return nil
			}
```
WDYT?

lib/symlink.go
MatthewJohn: 🤦 my plan was to remove it.. but got distracted verifying whether I had homeBinPath vs userBinPath was right.. that I immediately committed after and forgot :D

yermulnik (approved):
LGTM

yermulnik:
@MatthewJohn @warrensbox Please review.

warrensbox (approved):
I am okay with this change

PR #546, opened 2025-02-25
feat: Improve messaging

yermulnik:
* Drop trailing full stops in messages for consistency
* Capitalize messages
* Use tab instead of space to indent multiline messages
* `InstallProductVersion()`:
  * Drop pointless usage message
  * Improve error message by expanding word `args` to `Argument`
  * Use `product.GetName()` instead of hardcoded `terraform`
* Quit early when `--help` cmdline flag is supplied
* `PrintInvalidTFVersion()` print at Warn log level instead of Info
* `PrintInvalidMinorTFVersion()` print at Warn log level instead of Info
* `UsageMessage()` drop pointless blank lines and extraneous howto
* Code autoformatting in changed files (`golint`)

### DIFF
--- a/lib/install.go
+++ b/lib/install.go
@@ -65,7 +65,7 @@ func install(product Product, tfversion, binPath, installPath, mirrorURL, goarch
 	exist := versionExist(tfversion, tflist) // Check if version exists before downloading it
 
 	if !exist {
-		return fmt.Errorf("the provided %s version does not exist: %q.\n Try `tfswitch -l` to see all available versions", product.GetId(), tfversion)
+		return fmt.Errorf("Provided %s version does not exist: %q.\n\tTry `tfswitch -l` to see all available versions", product.GetId(), tfversion)
 	}
 
 	if goarch != runtime.GOARCH {
@@ -96,7 +96,7 @@ func install(product Product, tfversion, binPath, installPath, mirrorURL, goarch
 		logger.Fatalf("Unable to unzip %q file: %v", zipFile, errUnzip)
 	}
 
-	logger.Debug("Waiting for deferred functions.")
+	logger.Debug("Waiting for deferred functions")
 	wg.Wait()
 	/* rename unzipped file to terraform version name - terraform_x.x.x */
 	installFilePath := ConvertExecutableExt(filepath.Join(installLocation, product.GetExecutableName()))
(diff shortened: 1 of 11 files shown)


warrensbox (approved):
Looks good!

lib/install.go
```diff
@@ -65,7 +65,7 @@ func install(product Product, tfversion, binPath, installPath, mirrorURL, goarch
 	exist := versionExist(tfversion, tflist) // Check if version exists before downloading it
 
 	if !exist {
-		return fmt.Errorf("the provided %s version does not exist: %q.\n Try `tfswitch -l` to see all available versions", product.GetId(), tfversion)
+		return fmt.Errorf("Provided %s version does not exist: %q.\n\tTry `tfswitch -l` to see all available versions", product.GetId(), tfversion)
```
MatthewJohn, line 68: I think this would read better as "provided version %s does not exist" ?

lib/semver.go
```diff
@@ -47,15 +47,15 @@ func SemVerParser(tfconstraint *string, tflist []string) (string, error) {
...
 	}
 
-	return "", fmt.Errorf("did not find version matching constraint: %s", *tfconstraint)
+	return "", fmt.Errorf("Did not find version matching constraint: %s", *tfconstraint)
 }
 
 // PrintInvalidTFVersion Print invalid TF version
 func PrintInvalidTFVersion() {
-	logger.Info("Version does not exist or invalid terraform version format.\n Format should be #.#.# or #.#.#-@# where # are numbers and @ are word characters.\n For example, 0.11.7 and 0.11.9-beta1 are valid versions")
+	logger.Warn("Version does not exist or invalid terraform version format.\n\tFormat should be #.#.# or #.#.#-@# where # are numbers and @ are word characters.\n\tFor example, 0.11.7 and 0.11.9-beta1 are valid versions")
```
MatthewJohn, line 55: Should these not be errors?

lib/install.go
yermulnik: Yeah, good catch. Changed to `Requested version %s does not exist` via f03eb52 via https://github.com/warrensbox/terraform-switcher/pull/549

lib/semver.go
yermulnik: Hmm, not sure I'm getting the point.
I wanted to make these look differently instead of INFO, and WARN looked good.
Or do you mean these should better be ERR?

MatthewJohn: Forgot this PR was merged.. maybe for another time :)

yermulnik: Ah, I was reading "Should these not be errors?" as you meant INFO was okay 🤣 
Pushed to https://github.com/warrensbox/terraform-switcher/pull/549
Thanks

lib/install.go
yermulnik: hahaha, just noticed that `%s` in this place is for product name =))) reverted in #549

PR #549, opened 2025-02-27
feat: recommend updating `$PATH` only when necessary

yermulnik:
Do not print helper message to add bin directory to `$PATH` when it is already on the `$PATH`.

### DIFF
--- a/lib/symlink.go
+++ b/lib/symlink.go
@@ -139,7 +139,7 @@ func ChangeProductSymlink(product Product, binVersionPath string, userBinPath st
 			var checkDirInPath bool = false
 			checkDirPath := strings.TrimRight(dirPath, "/")
 
-			for _, path := range strings.Split(os.Getenv("ZZZ"), ":") {
+			for _, path := range strings.Split(os.Getenv("PATH"), ":") {
 				checkPath := strings.TrimRight(strings.Replace(path, "~", homedir, 1), "/")
 
 				if checkPath == checkDirPath {
--- a/lib/symlink.go
+++ b/lib/symlink.go
@@ -137,12 +137,11 @@ func ChangeProductSymlink(product Product, binVersionPath string, userBinPath st
 			logger.Debugf("Symlink created at %q", location)
 
 			var isDirInPath bool = false
-			checkDirPath := strings.TrimRight(dirPath, "/")
 
-			for _, path := range strings.Split(os.Getenv("PATH"), ":") {
-				checkPath := strings.TrimRight(strings.Replace(path, "~", homedir, 1), "/")
+			for _, envPathElement := range strings.Split(os.Getenv("PATH"), ":") {
+				expandedEnvPathElement := strings.TrimRight(strings.Replace(envPathElement, "~", homedir, 1), "/")
 
-				if checkPath == checkDirPath {
+				if expandedEnvPathElement == strings.TrimRight(dirPath, "/") {
 					isDirInPath = true
 					break
 				}
--- a/lib/symlink.go
+++ b/lib/symlink.go
@@ -135,7 +135,23 @@ func ChangeProductSymlink(product Product, binVersionPath string, userBinPath st
 		err = CreateSymlink(binVersionPath, location)
 		if err == nil {
 			logger.Debugf("Symlink created at %q", location)
-			logger.Warnf("Run `export PATH=\"$PATH:%s\"` to append %q to $PATH", dirPath, location)
+
+			var checkDirInPath bool = false
+			checkDirPath := strings.TrimRight(dirPath, "/")
+
+			for _, path := range strings.Split(os.Getenv("ZZZ"), ":") {
+				checkPath := strings.TrimRight(strings.Replace(path, "~", homedir, 1), "/")
+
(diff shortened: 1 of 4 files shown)


lib/symlink.go
```diff
@@ -135,7 +135,23 @@ func ChangeProductSymlink(product Product, binVersionPath string, userBinPath st
 		err = CreateSymlink(binVersionPath, location)
 		if err == nil {
 			logger.Debugf("Symlink created at %q", location)
-			logger.Warnf("Run `export PATH=\"$PATH:%s\"` to append %q to $PATH", dirPath, location)
+
+			var isDirInPath bool = false
+			checkDirPath := strings.TrimRight(dirPath, "/")
+
+			for _, path := range strings.Split(os.Getenv("PATH"), ":") {
+				checkPath := strings.TrimRight(strings.Replace(path, "~", homedir, 1), "/")
```
MatthewJohn, line 143: From what I can see/read, tilde expansion would only work during setting environment variables, meaning the resulting environment variable would contain the full path.
If an envionment variable does contain a tilde, then it won't be expanded and won't work as expected. I tested this a little locally and couldn't get it to expand.
So not sure if this replacement is required.
```suggestion
				checkPath := strings.TrimRight(path, "/")
```

lib/symlink.go
```diff
@@ -135,7 +135,23 @@ func ChangeProductSymlink(product Product, binVersionPath string, userBinPath st
 		err = CreateSymlink(binVersionPath, location)
 		if err == nil {
 			logger.Debugf("Symlink created at %q", location)
-			logger.Warnf("Run `export PATH=\"$PATH:%s\"` to append %q to $PATH", dirPath, location)
+
+			var isDirInPath bool = false
+			checkDirPath := strings.TrimRight(dirPath, "/")
+
+			for _, path := range strings.Split(os.Getenv("PATH"), ":") {
```
MatthewJohn, line 142: We have lots of location, path etc. variables floating around - is it worth denoting this as envVariablePath or similar?

lib/symlink.go
```diff
@@ -135,7 +135,23 @@ func ChangeProductSymlink(product Product, binVersionPath string, userBinPath st
...
+				checkPath := strings.TrimRight(strings.Replace(path, "~", homedir, 1), "/")
+
+				if checkPath == checkDirPath {
+					isDirInPath = true
+					break
+				}
+			}
+
+			if !isDirInPath {
+				logger.Warnf("Run `export PATH=\"$PATH:%s\"` to append %q to $PATH", dirPath, location)
```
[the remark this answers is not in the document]
yermulnik: Might be worth I guess. Though I don't have any Windows instance at the hand.
If you have a hint on what we should suggest for Windows users, please let me know. Maybe even add a commit suggestion so that I don't misinterpret Windows syntax 😺 
Thanks.

lib/symlink.go
yermulnik: This is what I've got to test tilde in `PATH`:
```shell
> export PATH="~/.local/bin:~/bin:/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin:/usr/local/go/bin:/home/linuxbrew/.linuxbrew/bin:"

> echo $PATH
~/.local/bin:~/bin:/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin:/usr/local/go/bin:/home/linuxbrew/.linuxbrew/bin:

> hash -r

> which -a kitty

> type -a kitty
kitty is /home/georgii/.local/bin/kitty

> kitten --version
kitten 0.39.1 created by Kovid Goyal
```
So it seems to be expanded by Bash correctly.

MatthewJohn: > `which` expectedly fails, because it's strictly POSIX-compliant and «It does not canonicalize path names»

Aha, this is why I went wrong :D 
Fair enough!

lib/symlink.go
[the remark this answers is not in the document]
MatthewJohn: I would be googling it :D
Perhaps:
```
set PATH="%PATH%:%s"
```
But, I suppose, to add it perminently, you'd be adding via `My Computer -> Something -> Profile Settings -> Environment` , probably with some bits inbetween :P
Though apparently windows 7+ has `setx`, which will set it perminently.
And of course `set` should be different if it's in powershell (though I guess the CMD-version should work fine in it?)

Ergh.. Windows

lib/symlink.go
yermulnik: Yeah, agree, it makes sense. Done via aeabba5
Please take a look again. Thanks.

lib/symlink.go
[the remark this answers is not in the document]
yermulnik: Well... let's then not guesstimate and expect Windows users either to be smart enough, or to raise a task for us and provide something that works for them under Windows =)

MatthewJohn (approved):

lib/symlink.go
yermulnik: Thanks for valuable suggestions. Will be looking into improving #551 now 👍🏻

PR #536, opened 2025-01-24
Add show latest command that prints latest version that conform to contraints

MatthewJohn:
Issue #301

PR #551, opened 2025-02-28
feat: Implement exclusive locking during download and install

yermulnik:
Ref: https://github.com/warrensbox/terraform-switcher/discussions/423

Inspired by: https://github.com/tofuutils/tenv/issues/116

### DIFF
--- a/lib/install.go
+++ b/lib/install.go
@@ -7,6 +7,7 @@ import (
 	"runtime"
 	"strconv"
 	"sync"
+	"time"
 
 	"github.com/manifoldco/promptui"
 
@@ -51,20 +52,30 @@ func GetInstallLocation(installPath string) string {
 func install(product Product, tfversion, binPath, installPath, mirrorURL, goarch string) error {
 	var wg sync.WaitGroup
 
-	// check to see if the requested version has been downloaded before
 	installLocation := GetInstallLocation(installPath)
 	installFileVersionPath := ConvertExecutableExt(filepath.Join(installLocation, product.GetVersionPrefix()+tfversion))
-	recentDownloadFile := CheckFileExist(installFileVersionPath)
-	if recentDownloadFile {
+
+	// Create exclusive lock to prevent multiple concurrent installations
+	lockFile := filepath.Join(installLocation, product.GetId()) + ".lock"
+	logger.Debugf("Attempting to acquire lock %q", lockFile)
+	// 90 attempts * 2 seconds = 3 minutes to acquire lock, otherwise bail out
+	if err := acquireLock(lockFile, 90, 2*time.Second); err != nil {
+		logger.Fatal(err)
+	}
+	// Release lock when done
+	defer releaseLock(lockFile)
+
+	// check to see if the requested version has been downloaded before
+	if recentDownloadFile := CheckFileExist(installFileVersionPath); recentDownloadFile {
 		return switchToVersion(product, tfversion, binPath, installPath, installFileVersionPath)
 	}
 
 	// If the requested version had not been downloaded before
 	// Set list all true - all versions including beta and rc will be displayed
-	tflist, _ := getTFList(mirrorURL, true)  // Get list of versions
-	exist := versionExist(tfversion, tflist) // Check if version exists before downloading it
+	tflist, _ := getTFList(mirrorURL, true) // Get list of versions
 
-	if !exist {
+	// Check if version exists before downloading it
+	if exist := versionExist(tfversion, tflist); !exist {
 		return fmt.Errorf("Provided %s version does not exist: %q.\n\tTry `tfswitch -l` to see all available versions", product.GetId(), tfversion)
 	}
 
@@ -87,12 +98,14 @@ func install(product Product, tfversion, binPath, installPath, mirrorURL, goarch
 
 	/* If unable to download file from url, exit(1) immediately */
 	if errDownload != nil {
+		releaseLock(lockFile)
 		logger.Fatalf("Error downloading: %s", errDownload)
 	}
 
 	/* unzip the downloaded zipfile */
 	_, errUnzip := Unzip(zipFile, installLocation, product.GetExecutableName())
 	if errUnzip != nil {
+		releaseLock(lockFile)
 		logger.Fatalf("Unable to unzip %q file: %v", zipFile, errUnzip)
 	}
 
--- a/lib/lockfile.go
+++ b/lib/lockfile.go
@@ -7,7 +7,7 @@ import (
 )
 
 // Create exclusive lock
-func acquireLock(lockFile string, LockWaitMaxAttempts int, LockWaitSleep time.Duration) error {
+func acquireLock(lockFile string, LockWaitMaxAttempts int, LockWaitInterval time.Duration) error {
 	var waitForLock bool = true
 	var LockWaitCount int = 0
 
@@ -27,7 +27,7 @@ func acquireLock(lockFile string, LockWaitMaxAttempts int, LockWaitSleep time.Du
 				logger.Warnf("Unable to get lock %q last modification time: %v", lockFile, err)
 			}
 
-			time.Sleep(LockWaitSleep)
+			time.Sleep(LockWaitInterval)
 		} else {
 			waitForLock = false
 			logger.Debugf("Acquired lock %q", lockFile)
--- a/lib/lockfile.go
+++ b/lib/lockfile.go
@@ -1,0 +1,52 @@
+package lib
+
+import (
+	"fmt"
+	"os"
+	"time"
+)
+
+// Create exclusive lock
+func acquireLock(lockFile string, LockWaitMaxAttempts int, LockWaitSleep time.Duration) error {
+	var waitForLock bool = true
+	var LockWaitCount int = 0
+
+	logger.Debugf("Attempting to acquire lock %q", lockFile)
+	for waitForLock {
+		LockWaitCount++
+		if _, err := os.OpenFile(lockFile, os.O_CREATE|os.O_EXCL, 0o644); err != nil {
+			if LockWaitCount > LockWaitMaxAttempts {
+				return fmt.Errorf("Unable to acquire lock %q", lockFile)
+			}
+
+			logger.Infof("Waiting for lock %q to be released (attempt %d out of %d)", lockFile, LockWaitCount, LockWaitMaxAttempts)
+
+			if lockFileInfo, err := os.Stat(lockFile); err == nil {
+				logger.Debugf("Lock %q last modification time: %s", lockFile, lockFileInfo.ModTime())
+			} else {
+				logger.Warnf("Unable to get lock %q last modification time: %v", lockFile, err)
+			}
+
+			time.Sleep(LockWaitSleep)
+		} else {
+			waitForLock = false
+			logger.Debugf("Acquired lock %q", lockFile)
+			break
+		}
+	}
+
+	return nil
+}
+
+// Release lock file
+func releaseLock(lockFile string) {
+	if exist := CheckFileExist(lockFile); exist {
+		if err := os.Remove(lockFile); err != nil {
+			logger.Debugf("Error releasing lock %q: %v", lockFile, err)
+		} else {
+			logger.Debugf("Releasing lock %q", lockFile)
+		}
+	} else {
+		logger.Debugf("Lock %q not found", lockFile)
+	}
+}


MatthewJohn:
I wonder if this should only occur during download, or whether there's value in doing this during the entire proces (including updating symlink?)

lib/lockfile.go
```diff
@@ -0,0 +1,52 @@
+package lib
```
MatthewJohn, line 1: Could we add some tests for this?

lib/install.go
```diff
@@ -51,20 +52,30 @@ func GetInstallLocation(installPath string) string {
...
+
+	// Create exclusive lock to prevent multiple concurrent installations
+	lockFile := filepath.Join(installLocation, product.GetId()) + ".lock"
+	logger.Debugf("Attempting to acquire lock %q", lockFile)
+	// 90 attempts * 2 seconds = 3 minutes to acquire lock, otherwise bail out
+	if err := acquireLock(lockFile, 90, 2*time.Second); err != nil {
+		logger.Fatal(err)
+	}
+	// Release lock when done
+	defer releaseLock(lockFile)
```
MatthewJohn, line 66: We would need to start being _SUPER_ careful with this...
Since `log.Fatal` does _not_ call the defer method (as you've alluded to with the below additional calls to `releaseLock` before the `log.fatal`s).. we need to ensure that any method that we call in this function never calls `log.Fatal` and the use of log.Fatal is quite broadly used across the code base.

Since `defer`s _do_ get called during panics, I think maybe we should convert all `log.Fatal`s to `log.Panic`s - what do you think?

lib/install.go
```diff
@@ -87,12 +98,14 @@ func install(product Product, tfversion, binPath, installPath, mirrorURL, goarch
 
 	/* If unable to download file from url, exit(1) immediately */
 	if errDownload != nil {
+		releaseLock(lockFile)
 		logger.Fatalf("Error downloading: %s", errDownload)
```
MatthewJohn, line 102: If we switched to log.Panic, we could remove the `releaseLock` call and let the defer handle it

lib/install.go
```diff
@@ -87,12 +98,14 @@ func install(product Product, tfversion, binPath, installPath, mirrorURL, goarch
...
 	}
 
 	/* unzip the downloaded zipfile */
 	_, errUnzip := Unzip(zipFile, installLocation, product.GetExecutableName())
 	if errUnzip != nil {
+		releaseLock(lockFile)
```
MatthewJohn, line 108: Same above

lib/lockfile.go
```diff
@@ -0,0 +1,52 @@
...
+
+			if lockFileInfo, err := os.Stat(lockFile); err == nil {
+				logger.Debugf("Lock %q last modification time: %s", lockFile, lockFileInfo.ModTime())
+			} else {
+				logger.Warnf("Unable to get lock %q last modification time: %v", lockFile, err)
+			}
+
+			time.Sleep(LockWaitInterval)
+		} else {
+			waitForLock = false
```
MatthewJohn, line 32: Not sure about the mixture of setting the conditional to stop the for loop and also breaking below.
Could the for loop be a `for _ := range LockWaitMaxAttempts`, which handles the increment and stops on reaching the max value.
If the lock is successful, we could `return nil` early. Otherwise, an error is returned outside of the loop that we've reached maximum attempts. This means we're checking the "success path" and then returning an error on all bad paths, rather than trying to check all errors?
What do you think?

Psuedo-code:
```
for _ := range maxAttempts {
  _, createLockError := os.OpenFile()
  if ! createLockError {
    // Happy path
    return nil
  }
  // Logic to handle check error cases
  
}
return fmt.Errorf("Reached timeout trying to acquire lock %q", lockFile) 
```

lib/lockfile.go
```diff
@@ -0,0 +1,52 @@
...
+func acquireLock(lockFile string, LockWaitMaxAttempts int, LockWaitInterval time.Duration) error {
+	var waitForLock bool = true
+	var LockWaitCount int = 0
+
+	logger.Debugf("Attempting to acquire lock %q", lockFile)
+	for waitForLock {
+		LockWaitCount++
+		if _, err := os.OpenFile(lockFile, os.O_CREATE|os.O_EXCL, 0o644); err != nil {
+			if LockWaitCount > LockWaitMaxAttempts {
+				return fmt.Errorf("Unable to acquire lock %q", lockFile)
```
MatthewJohn, line 19: In these logs, do we wish to refer to it as a lock file?
```suggestion
				return fmt.Errorf("Unable to acquire lock file %q", lockFile)
```

yermulnik:
> I wonder if this should only occur during download, or whether there's value in doing this during the entire proces (including updating symlink)?

I was thinking on this item too as initially I went with locking only requested version, which looked like kinda quick and targeted locking.
But then I realized that during entire process we also should secure against concurrent writes to a GPG key (if it didn't exist yet), to an archive being downloaded, to a content being unarchived (the binary in archive in unversioned), and also to a symlnik.
So I thought we'd better lock entire install process for whatever files are involved to not get written to concurrently.
The overall install process, given there's no network blips and hiccups, takes around 3 secs for me, so it should not block other `tfswitch` runs for too long. Especially given that such a use case of concurrent `tfswitch` runs looks to be more of a rare case and since ever has been [reported only once](https://github.com/warrensbox/terraform-switcher/discussions/423#discussioncomment-12347489) as of yet.

lib/lockfile.go
yermulnik: Let me try to figure it out 👍🏻

lib/install.go
yermulnik: The thing about diff on handling deferred functions certainly gives preference to `log.Panic`, though when I see the exception it throws to console output, I look into `log.Fatal` as a more user-friendly looking thing 🤷🏻 
So my preference is for what you suggested at https://github.com/warrensbox/terraform-switcher/pull/536 — replace `log.Fatal`s with `return`'s with error and have `main.go` handle the rest.
When I'm assuming that `return` handles deferred functions, am I not mistaken though? 🤔

lib/lockfile.go
yermulnik: > Not sure about the mixture of setting the conditional to stop the for loop and also breaking below.

Exactly! I was to remove it but go distracted by something else. Thanks for catching it 👍🏻 

Looking into other suggestions as they indeed a great. Thanks! 🤝

lib/lockfile.go
yermulnik: I'd prefer it to be referred as "lock" rather than a "file" as its main idea is locking concurrent access, rather than handling a file. Does it sound reasonable?

lib/lockfile.go
yermulnik: 5950f28

lib/lockfile.go
yermulnik: 80256a3

lib/install.go
yermulnik: https://github.com/warrensbox/terraform-switcher/pull/551#discussion_r1976474724

lib/install.go
yermulnik: https://github.com/warrensbox/terraform-switcher/pull/551#discussion_r1976474724

yermulnik:
@MatthewJohn I'm done updating it base on your suggestions. Please give it another round of review. Thanks.

lib/lockfile_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+// Test Locking
+func TestLocking(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Log("Skipping test on Windows: cannot figure it out 01-Mar-2025")
+	} else {
+		var lockFile string = "../test-data/lockfile.lock"
+
+		// Acquire lock
+		if lockAcquireErr := acquireLock(lockFile, 1, 1); lockAcquireErr == nil {
```
MatthewJohn, line 16: I think the main tests I think would be needed are:
 * Test running aquireLock
   *  Ensure lock file is created
 * Test create the lock file and then releaseLock:
   *  ensure the file is removed
 * Test running aquireLock and then running aquireLock
   * ensuring that an error is returned

lib/lockfile.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	return fmt.Errorf("Failed to acquire lock %q", lockFile)
+}
+
+// Release lock file
+func releaseLock(lockFile string) {
+	if exist := CheckFileExist(lockFile); exist {
+		if err := os.Remove(lockFile); err != nil {
+			logger.Debugf("Error releasing lock %q: %v", lockFile, err)
```
MatthewJohn, line 41: I'm not sure if releasing the lock file shouldn't be a more urgent problem.
It could mean that the lock file is left in place and, if only debug, won't (by default) tell the user. And future runs will fail and won't be too obvious where the lock file was left from (and making GH issues around it very difficult, because they'll be no trace of the source of the lock file)

lib/lockfile_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"runtime"
+	"testing"
+)
+
+// Test Locking
+func TestLocking(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Log("Skipping test on Windows: cannot figure it out 01-Mar-2025")
```
MatthewJohn, line 11: Errr!!
I'm not _as_ worried about the issues with perhaps the test coverage being missing.. does this not mean that the main locking function doesn't work on Windows?

lib/lockfile.go
yermulnik: > I'm not sure if releasing the lock file shouldn't be a more urgent problem.

Makes sense. Switched to Warn level.

lib/lockfile_test.go
yermulnik: I guess it was my mistake in unlocking logic. Looking into it now.

lib/lockfile_test.go
yermulnik: Working on it. Thanks for great review comments! Appreciate it 🤝

yermulnik: Hope I got this working as expected 🤞🏻 c3ef2e3

yermulnik:
@MatthewJohn Please give it another round.

lib/lockfile.go
```diff
@@ -0,0 +1,66 @@
...
+		if lockAttemptCounter < lockWaitMaxAttempts {
+			time.Sleep(lockWaitInterval)
+		}
+	}
+
+	return nil, fmt.Errorf("Failed to acquire lock %q", lockFile)
+}
+
+// Release and remove lock
+func releaseLock(lockFile string, lockedFile *os.File) {
```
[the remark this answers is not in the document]
MatthewJohn: Though, I guess it expands to "lock file file handle", which maybe is a weird

lib/lockfile.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+	"os"
+	"time"
+)
+
+// Acquire exclusive lock
+func acquireLock(lockFile string, lockWaitMaxAttempts int, lockWaitInterval time.Duration) (*os.File, error) {
+	logger.Debugf("Attempting to acquire lock %q", lockFile)
+
+	for lockAttempt := range lockWaitMaxAttempts {
```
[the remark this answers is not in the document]
yermulnik: 1b32dd9

lib/lockfile.go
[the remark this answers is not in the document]
yermulnik: `lockedFH`?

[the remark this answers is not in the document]
yermulnik: Let me push it as it looks better I guess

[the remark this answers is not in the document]
yermulnik: fc1caa5

lib/install.go
```diff
@@ -51,21 +52,32 @@ func GetInstallLocation(installPath string) string {
...
 
-	// check to see if the requested version has been downloaded before
 	installLocation := GetInstallLocation(installPath)
 	installFileVersionPath := ConvertExecutableExt(filepath.Join(installLocation, product.GetVersionPrefix()+tfversion))
-	recentDownloadFile := CheckFileExist(installFileVersionPath)
-	if recentDownloadFile {
+
+	// Create exclusive lock to prevent multiple concurrent installations
+	// Put lockfile in temp directory to get it cleaned up on reboot
+	lockFile := filepath.Join(os.TempDir(), ".tfswitch."+product.GetId()+".lock")
```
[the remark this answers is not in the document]
yermulnik: I believe that any sane system, where `tfswitch` might be a matter, can't be that much borked to have no system-managed temp dir. The scenario for the disclaimer, I guess, is for when tmp dir is re-declared by means of env var, which is then means the operator intentionally shot themselves in the foot and it's out of our scope and responsibility then 🤷🏻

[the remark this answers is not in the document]
MatthewJohn: True - but still makes me nervous :D

lib/lockfile_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"time"
+)
+
+// Test Locking
+func TestLocking(t *testing.T) {
+	var lockFile string = ".tfswitch.lock"
+	lockFilePath := filepath.Join(t.TempDir(), lockFile)
+
+	t.Logf("Testing lock acquirement: %s", lockFilePath)
+
```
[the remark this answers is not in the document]
MatthewJohn: Maybe just `Testing lock aquire` 🧀

lib/lockfile_test.go
```diff
@@ -0,0 +1,38 @@
...
+	// Acquire lock
+	if lockedFile, err := acquireLock(lockFilePath, 1, 1*time.Second); err == nil {
+		t.Logf("Lock acquired successfully: %s", lockFilePath)
+
+		// Concurrent lock
+		t.Logf("Testing concurrent lock acquirement: %s", lockFilePath)
```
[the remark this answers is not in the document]
yermulnik: Nah, thanks =)

lib/lockfile_test.go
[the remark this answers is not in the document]
yermulnik: Any objection to leave it as is? =)

lib/install.go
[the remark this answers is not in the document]
yermulnik: What makes you nervous less: absence of locking or the hypothetical broken station which has no tmp dir or has it not writable? 🤣

lib/lockfile_test.go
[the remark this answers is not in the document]
yermulnik: Nah, thanks.

MatthewJohn (approved):

PR #552, opened 2025-03-01
chore: [`CODEOWNERS`] Take code ownership for safekeeping

yermulnik:
Take code ownership for safekeeping after more than half a year of inactivity (@jukie @MatrixCrawler @crablab)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,4 +5,5 @@
 
 # The entries are placed in a historical order
 # 20240408: initial CODEOWNERS commit (added repo owner and all current collaborators)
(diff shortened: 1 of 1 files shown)


MatthewJohn (approved):
One for @warrensbox I think as well

yermulnik:
> One for @warrensbox I think as well

Yep, that's why I also changed line ordering to match how the CODEOWNERS is parsed 👍🏻

warrensbox:
Looks good. Go for it! Thanks @yermulnik @MatthewJohn

warrensbox (approved):

PR #566, opened 2025-03-14
feat: Add env vars for `install` and `bin` args

### LINKED ISSUE #562: [Feature request] Environment variable for non-standard install location
### Discussed in https://github.com/warrensbox/terraform-switcher/discussions/561

<div type='discussions-op-text'>

<sup>Originally posted by **dbopus2** March 10, 2025</sup>
Would you consider adding an environment variable for non-standard install location? Same functionality as the -i and --install cli options.</div>

yermulnik:
Resolves #561 (discussion)
Resolves #562 (issue from discussion)

* Add Env vars for custom binary path and custom install path
  * Add Env var for log level for parity with other TOML keys
* Add TOML config key for custom install path for consistency
* Improve wording in resolved parameter values debugging output
* Add tests for new Env vars and TOML config key
* Re-factor env vars and TOML keys parsing for simplicity and consistency
* Documentation
  * Update https://tfswitch.warrensbox.com/usage/commandline/ accordingly
  * Update https://tfswitch.warrensbox.com/usage/config-files/ accordingly
  * Add `arch` and `log-level` TOML keys description to https://tfswitch.warrensbox.com/usage/config-files/
  * Improve wording in some places
* GitHub Actions:
  * Have workflows read Go version from `go.mod` instead of explicit hard-coded version
  * Reset generic workflow perms to none and adjust at job level as necessary


.github/workflows/build.yml
```diff
@@ -21,7 +22,8 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v5
         with:
-          go-version: '1.23'
```
MatthewJohn, line 24: Is go version not needed anymore?

lib/param_parsing/parameters.go
```diff
@@ -35,6 +35,24 @@ type Params struct {
...
+}
+
+// This is used to automatically instate Environment variables and TOML keys
+var paramMappings = []paramMapping{
+	{param: "Arch", env: "TF_ARCH", toml: "arch", description: "CPU architecture"},
+	{param: "CustomBinaryPath", env: "TF_BINARY_PATH", toml: "bin", description: "Custom binary path"},
+	{param: "DefaultVersion", env: "TF_DEFAULT_VERSION", toml: "default-version", description: "Default version"},
+	{param: "InstallPath", env: "TF_INSTALL_PATH", toml: "install", description: "Custom install path"},
+	{param: "LogLevel", env: "TF_LOG_LEVEL", toml: "log-level", description: "Log level"},
+	{param: "Product", env: "TF_PRODUCT", toml: "product", description: "Product"},
```
MatthewJohn, line 52: I feel like there's a bunch of overlap with https://github.com/warrensbox/terraform-switcher/pull/566/files#diff-6102d401ce5628c1df4636f2ca1d1d81e8d52346a74662f47d018ed8611b436bR73
Could the whole definitions not be used, so we don't duplicate name, and description and keep the definitions in a single place?

www/docs/usage/commandline.md
```diff
@@ -59,13 +59,32 @@ This can either be set to:
...
+export TF_PRODUCT="opentofu"
 tfswitch # Will install opentofu instead of terraform
 ```
 
+### `TF_LOG_LEVEL`
+
+`TF_LOG_LEVEL` environment variable can be set to override default log level.
+
+- Supported log levels:
+  - `INFO`:   includes `PANIC`, `FATAL`, `ERROR`
```
MatthewJohn, line 71: I guess, INFO debug level also includes `INFO` ;) 

Worth mentioning what the default log level is?

www/docs/usage/commandline.md
```diff
@@ -59,13 +59,32 @@ This can either be set to:
...
+  - `ERROR`:  includes `PANIC`, `FATAL`, `ERROR`, `WARN`, `INFO`
```
MatthewJohn, line 72: Wait, error won't include INFO?
Are the keys just switched up between 71/72?

www/docs/usage/commandline.md
```diff
@@ -76,8 +95,32 @@ tfswitch # Will install opentofu instead of terraform
...
+
+```bash
+export TF_BINARY_PATH="$HOME/bin/terraform" # Path to the file
+tfswitch # Will install binary as $HOME/bin/terraform
+```
+
+### `TF_INSTALL_PATH`
+
+`tfswitch` defaults to download binaries to the `$HOME/.terraform.versions/` directory.  
+`TF_INSTALL_PATH` environment variable can be set to specify a parent directory for `.terraform.versions` directory. Current user must have write permissions to the target directory. If the target directory does not exist, `tfswitch` will create it.
```
MatthewJohn, line 117: I think it should say "specify the parent directory for".. but also worth mentioning that .terraform.versions will be automatically created. A new user might wonder what .terraform.versions directory is and how they should create it

www/docs/usage/commandline.md
```diff
@@ -76,8 +95,32 @@ tfswitch # Will install opentofu instead of terraform
...
+
+For example:
+
+```bash
+export TF_INSTALL_PATH="/tmp" # Path to the directory where `.terraform.versions` directory resides
```
MatthewJohn, line 122: I don't know if /tmp is a good idea - it sort of implies that all these are temporary files. But the cached binaries are relatively important (since it's expensive to keep downloading versions).
Maybe a dir in /var/cache would be a good example?

MatthewJohn (commented):
Will finish reviewing after work today

.github/workflows/build.yml
yermulnik: I switched to `go-version-file: 'go.mod'` to let `setup-go` action read Go version dynamically from `go.mod` file to reduce manual toil on syncing Go versions between multiple files and have one single "source of truth" for Go version
image

lib/param_parsing/parameters.go
yermulnik: Yep, we probably can. Though I don't see this mapping to match the `getopts` thing as this new mapping is solely for parity between env vars and params in TOML.
We can probably expand this new mapping onto the `getops` 🤔 But defo not in scope of this PR which is about another feature. I also do not want to overload this PR with changes that worth of its own issue and PR. Does it sound meaningful?

www/docs/usage/commandline.md
yermulnik: Yeah, exactly =) Fixed via 625d9e1

www/docs/usage/commandline.md
yermulnik: > I think it should say "specify the parent directory for"..

c249c9e

> but also worth mentioning that .terraform.versions will be automatically created. A new user might wonder what .terraform.versions directory is and how they should create it

Isn't this covered by «_If the target directory does not exist, `tfswitch` will create it._»? 🤔

lib/param_parsing/parameters.go
MatthewJohn: It was mainly the duplicate descriptions that concerned me tbh

yermulnik: You talked me round =) Let me push update and see if I got your idea right.

www/docs/usage/config-files.md
```diff
@@ -53,28 +56,27 @@ bin = "$HOME/bin/terraform"
...
 
 ```toml
 bin = "C:\\Users\\<%USRNAME%>\\bin\\terraform.exe"
 ```
 
-## Setting the default version using `.tfswitch.toml` file
+## Setting the default (fallback) version using `.tfswitch.toml` file
 
-The `.tfswitch.toml` file can be configured with a `default-version` attribute to configure tfswitch a particular version, if no other sources of versions are found
+If `tfswsitch` is unable to determine the version to use, it errors out.
```
MatthewJohn, line 69:
```suggestion
By default, If `tfswsitch` is unable to determine the version to use, it errors out.  
```
Taken out of context, this probably isn't great and is the first thing someone will read.

www/docs/usage/config-files.md
```diff
@@ -53,28 +56,27 @@ bin = "$HOME/bin/terraform"
...
 ```toml
 bin = "C:\\Users\\<%USRNAME%>\\bin\\terraform.exe"
 ```
 
-## Setting the default version using `.tfswitch.toml` file
+## Setting the default (fallback) version using `.tfswitch.toml` file
 
-The `.tfswitch.toml` file can be configured with a `default-version` attribute to configure tfswitch a particular version, if no other sources of versions are found
+If `tfswsitch` is unable to determine the version to use, it errors out.  
+The `.tfswitch.toml` file can be configured with a `default-version` attribute for `tfswitch` to use a particular version, if no other sources of versions are found
```
MatthewJohn, line 70: This mentions the TOML, is the only way to find out about the env variable to see the seperate page about "Setting the default (fallback) version"?
I know the title emplies it's _just_ for using the .tfswitch.toml, but the fact there's an environment variable feels weird wording it in this way.

I almost feel like we should have a config guide.. and each of the configs say whether they can be set via cmd-switch, env variable or toml file. Rather than duplicates for each of the configuration methodologies

Then the title is just "## Setting the default (fallback) version" explaining the behavoir, the configuration available to overcome and then the various methods for setting the config

lib/param_parsing/environment.go
```diff
@@ -1,23 +1,43 @@
...
+
+		f := reflect.Indirect(reflectedParams).FieldByName(param)
+		if f.Kind() != reflect.String {
+			logger.Warnf("Parameter %q is not a string, skipping assignment from environment variable %q", param, env)
+			continue
+		}
+
+		if envVarValue := os.Getenv(env); envVarValue != "" {
+			logger.Debugf("%s (%q) from environment variable %q: %q", description, toml, env, envVarValue)
+			if !f.CanSet() {
```
MatthewJohn, line 36: In what situation can't this be set? 🤔 
Does CanSet give an error?

www/docs/usage/config-files.md
yermulnik: Thanks 👍🏻 
b11152a

www/docs/usage/config-files.md
yermulnik: Yep, I agree that we 're better off revisiting and reviewing the documentation to bring it consistent with what have changes over past year or so... Gonna be a task with its own issue/PR.

lib/param_parsing/environment.go
yermulnik: > In what situation can't this be set? 🤔

«_CanSet reports whether the value of v can be changed. A [Value](https://pkg.go.dev/reflect#Value) can be changed only if it is addressable and was not obtained by the use of unexported struct fields._»
I believe it's best to check, given corresponding func exists in the same package, rather than blindly rely that `SetString()` won't fail which results in panic: https://pkg.go.dev/reflect#Value.CanSet

The `f` is reflection of a key from `params`. Let me adjust naming for clarity.
Done: b8648ee

> Does CanSet give an error?

Nope. It returns bool as the only outcome: https://pkg.go.dev/reflect#Value.CanSet

MatthewJohn: I guess it's a situation where.. I (we?) think this _should_ always work, but reflect may think otherwise. We're only iterating through "known" env variables that _should_ always work.
So, in the case it doesn't and this warning is shown to users, there's several situations:
 1. They may or may not realise it (since it doesn't cause a failure). This will result in the configuration that they _wanted_ to apply to fail (with/without being noticed)
 2. If the user sees this warning, they've been told it can't be set and I'd imagine they wouldn't know why. So they may waste time trying to solve themselves. But really, it's probably more of an edge-case that we'd need to investigate.

Basically, I'm not sure if panicing isn't a _bad_ thing here, or at least erroring and saying it's _also_ and "internal error" and to raise an issue.

MatthewJohn (approved):

PR #570, opened 2025-03-20
fix: reinstate env var expansion in `bin` and add the same for `install` TOML params

yermulnik:
* Reinstate historical behavior on Env var expansion for `bin` TOML key, which was inadvertently dropped in #566
* Add the same feature for `install` TOML key
* Update corresponding tests

### DIFF
--- a/lib/param_parsing/toml_test.go
+++ b/lib/param_parsing/toml_test.go
@@ -1,6 +1,7 @@
 package param_parsing
 
 import (
+	"os"
 	"testing"
 
 	"github.com/warrensbox/terraform-switcher/lib"
@@ -15,6 +16,7 @@ func prepare() Params {
 
 func TestGetParamsTOML_BinaryPath(t *testing.T) {
 	expected := "/usr/local/bin/terraform_from_toml"
+	os.Setenv("BIN_DIR_FROM_TOML", "/usr/local/bin")
 	params := prepare()
 	params, err := getParamsTOML(params)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


lib/param_parsing/toml_test.go
```diff
@@ -15,6 +16,7 @@ func prepare() Params {
 
 func TestGetParamsTOML_BinaryPath(t *testing.T) {
 	expected := "/usr/local/bin/terraform_from_toml"
+	os.Setenv("BIN_DIR_FROM_TOML", "/usr/local/bin")
```
MatthewJohn, line 19: What is using these environment variables? I think I'm missing something 🤔

yermulnik: It is loaded by `prepare()` func at the line below the one that you commented on.

MatthewJohn: But I mean, what is reading _this_ env variable in particular, I can't see to see it mentioned/used anywhere?
Sorry if I'm being stupid :(

yermulnik: Nothing is reading this particular var. The env vars in `bin` and `install` value in TOML are expanded: 
https://github.com/warrensbox/terraform-switcher/pull/570/files#diff-70bb7ecbe02c1c0fb2f1ba6c3626d57287498e82043aca43fa54cb8649c12392R57-R62
изображение

This was the feature for `bin` value in TOML that I dropped on accident in previous PR: https://github.com/warrensbox/terraform-switcher/pull/566/files#diff-70bb7ecbe02c1c0fb2f1ba6c3626d57287498e82043aca43fa54cb8649c12392L35
And hence I've put it back and implemented also for `install` key.

yermulnik: The var is used in test file and thus is set inside test function.

Issue #537, opened 2025-01-24
Add shell command completion autogeneration


yermulnik:
`fswitch` already has around a dozen+ of command line arguments which is complicate to remember and operate by humans when necessary.
Would be great to look up and employ a Golang library to autogenerate shell command completions from code on the fly (e.g. `tswitch --completion bash`) that can be soaked into shell dynamically (e.g. `. <(tfswitch --completion bash)`).

Common shells that shell command completion is shipped for usually are: Bash, Zsh, and Fish.

Once implemented, update [Homebrew formulae](https://github.com/warrensbox/homebrew-tap/blob/master/Formula/tfswitch.rb) to autogenerate and install shell command completions into common location.
Caveat: only @warrensbox has write access to https://github.com/warrensbox/homebrew-tap

yermulnik:
FWIW: My local quick workaround for Bash until this is implemented kinda natively
```bash
_tfswitch()
{
    local cur
    cur=${COMP_WORDS[COMP_CWORD]}
    COMPREPLY=( $( compgen -W "$( tfswitch --help 2>&1 | awk '$1 ~ "-.+,"' | awk -F= '{print $1}' | awk '{print $1,$2}' | tr -d ',' )" -- "$cur" ) )

} &&
complete -F _tfswitch tfswitch
```

yermulnik:
JFYI: #586

PR #586, opened 2025-05-23
feat(shell-completion): Add Bash completion script

yermulnik:
* Remove it once #537 is implemented
* Additionally bundle it into release artifacts

### DIFF
--- a/tfswitch-completion.bash
+++ b/tfswitch-completion.bash
@@ -1,5 +1,7 @@
 # Remove once https://github.com/warrensbox/terraform-switcher/issues/537 is implemented
-# shellcheck disable=SC2207 # W: Prefer mapfile or read -a to split command output (or quote to avoid splitting).
+# - SC2015 (info): Note that A && B || C is not if-then-else. C may run when A is true.
+# - SC2207 W: Prefer mapfile or read -a to split command output (or quote to avoid splitting).
+# shellcheck disable=SC2015,SC2207
 _tfswitch() {
 	local cur prev
 	cur=${COMP_WORDS[COMP_CWORD]}


.goreleaser.yml
```diff
@@ -41,6 +41,7 @@ archives:
       - CHANGELOG.md
       - LICENSE
       - README.md
+      - tfswitch-completion.*
```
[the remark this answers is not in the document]
yermulnik: It's intentional (for when/if we add shell completions for other shells).

tfswitch-completion.bash
```diff
@@ -0,0 +1,38 @@
...
+		return 0
+	fi
+
+	case "${prev}" in
+	-A | --arch)
+		COMPREPLY=($(compgen -W "386 amd64 arm arm64" -- "$cur"))
+		return 0
+		;;
+	-b | --bin)
+		_filedir
```
MatthewJohn, line 19: I think this is deprecated? https://github.com/scop/bash-completion/blob/721fa87c9c4ed5f2be8e890948f351f07731dcbb/bash_completion.d/000_bash_completion_compat.bash#L294

tfswitch-completion.bash
```diff
@@ -0,0 +1,38 @@
...
+		return 0
+		;;
+	-c | --chdir | -i | --install)
+		_filedir -d
+		return 0
+		;;
+	-g | --log-level)
+		COMPREPLY=($(compgen -W "DEBUG ERROR INFO NOTICE TRACE" -- "$cur"))
```
MatthewJohn, line 27: Could these be in order, perhaps?
TRACE
DEBUG
INFO
NOTICE
ERROR

What about warn?

tfswitch-completion.bash
yermulnik: Ah, great catch. Thanks. Though… it's complicated =)
Ubuntu 22.04 has v2.11 of scop/bash-completion and latest Homebrew have `bash-completion` v1.3 because of macOS's version of Bash (and folks have to install `bash-completion@2` if they have Bash upgraded).
So I implemented 524d1cd to support backward compatibility. WDYT?

tfswitch-completion.bash
yermulnik: > Could these be in order, perhaps? TRACE DEBUG INFO NOTICE ERROR

You mean in order of logging levels, while I prefer human friendly alphabetical sorting order =) Either way `compgen` sorts words alphabetically so putting these in completion file in an arbitrary order imho makes no point. Also sorted lists managed by human are less prone to e.g. duplicates (even despite `compgen` squeezes duplicates).
Thanks.

> What about warn?

Hmm, I took the list from output of `tfswitch --help` 🤔 Looking at `lib/logging.go` I also see no standalone `WARN  level 🤷🏻 It's probably something that we accidentally had not implemented (we also have `INFO` and `NOTICE` levels misordered in Normal logging which needs to be fixed by will change default logging verbosity).
I created #589 to look into this.

tfswitch-completion.bash
```diff
@@ -0,0 +1,38 @@
...
+	prev=${COMP_WORDS[COMP_CWORD - 1]}
+
+	if [[ ${cur} == -* ]]; then
+		COMPREPLY=($(compgen -W "$(tfswitch --help 2>&1 | grep -Eo '[[:space:]]+(-{1,2}[a-zA-Z0-9-]+)')" -- "$cur"))
+		return 0
+	fi
+
+	case "${prev}" in
+	-A | --arch)
+		COMPREPLY=($(compgen -W "386 amd64 arm arm64" -- "$cur"))
```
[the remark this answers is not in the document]
yermulnik: Bash arrays are more common to Bash completion files, so I'd like to keep it consistent with common approach. Also `mapfile` doesn't exist in Bash3 (which is default Bash version in macOS). Also Bash arrays (`VAR=(…)`) are a more readable notation I reckon.

MatthewJohn (approved):
I don't know much about this - I'm happy with it, but I'm not a great person to approve (but I'll give it anyway). I assume for something like this.. if it works, then... ;) 
@warrensbox  Any experience with bash completion?

yermulnik:
> I don't know much about this - I'm happy with it, but I'm not a great person to approve (but I'll give it anyway). I assume for something like this.. if it works, then... ;)

I use `tfswitch` daily, incl Bash completion suggestions, so I'll resolve any hiccups that may surface 🤝 Merging it.

yermulnik:
Oops, I missed that Brew tap content is managed via this repo rather than via tap's repo, so here's the consequent PR to update Brew tap with Bash completion installation: https://github.com/warrensbox/terraform-switcher/pull/590

PR #594, opened 2025-05-27
feat: Togglable color logging

### LINKED ISSUE #462: [Feature request] Disable color logs when not a terminal

yermulnik:
Resolves #462

- [x] Enable color logging by default when terminal supports colors
  - [x] Disable color logging by default when terminal doesn't support colors or when no TTY is allocated (non-interactive session)
- [x] Add `NO_COLOR` and `FORCE_COLOR` env vars support
- [x] Add TOML `no-color` and `force-color` params
- [x] Add `--no-color` and `--force-color` cmdline flags
- [x] Update documentation:
  - [x] env vars
  - [x] cmdline flags
  - [x] TOML parameters
- [x] Add tests for:
  - [x] env vars
  - [x] cmdline flags
  - [x] TOML params


lib/param_parsing/parameters.go
```diff
@@ -72,31 +79,49 @@ func populateParams(params Params) Params {
...
-	isShortRun := !params.VersionFlag && !params.HelpFlag
+	isNotShortRun := !params.VersionFlag && !params.HelpFlag
+
+	if isNotShortRun {
+		if params.ForceColor && params.NoColor {
+			logger.Fatal("(init) Cannot force color and disable color at the same time. Please choose one of them.")
+		}
+
+		if params.ForceColor {
+			os.Setenv("FORCE_COLOR", "true")
```
MatthewJohn, line 117: Why are we setting environment variables? Is it to just easily pass variables from here to logger?

yermulnik: `NO_COLOR` is already supported natively by underlying library: [gookit/color#features](https://github.com/gookit/color?rgh-link-date=2024-06-13T13%3A39%3A33.000Z#features) — so I decided to use this approach to not reinvent the wheel and let underlying library deal with this.

MatthewJohn: Ahh, gotcha :)

www/docs/usage/commandline.md
```diff
@@ -53,45 +56,78 @@ The most recently selected versions are presented at the top of the dropdown.
...
+running `tfswitch`
 
 ## Install binary for non-default architecture
 
-By default `tfswitch` will download the binary for the architecture of the host machine.
+By default `tfswitch` will download the binary for the architecture of the host
+machine.
 
-If you want to download the binary for non-default CPU architecture then you can provide the `-A` or `--arch` command line argument to download binaries for custom CPU architecture. Useful if you need to override binary architecture for whatever reason.
+If you want to download the binary for non-default CPU architecture then you
```
MatthewJohn: Not sure what's being changed in these lines - but seeing this, perhaps `non-default` architecture doesn't sound quite right - maybe `a CPU architecture that doens't match the host`?

www/docs/usage/commandline.md
```diff
@@ -53,45 +56,78 @@ The most recently selected versions are presented at the top of the dropdown.
 ## Show latest implicit version for stable releases
 
 1. Show the latest implicit stable version.
-2. Ex: `tfswitch -S 0.13` or `tfswitch --show-latest-stable 0.13` shows 0.13.6 (latest) version.
+2. Ex: `tfswitch -S 0.13` or `tfswitch --show-latest-stable 0.13` shows 0.13.6
```
MatthewJohn, line 59: Is there a reason these are being split into new lines? Just linting, because it feels.. odd

www/docs/usage/commandline.md
yermulnik: Agree, makes sense. Addressed: 3e67620

www/docs/usage/commandline.md
yermulnik: > Is there a reason these are being split into new lines?

Yes, the reason is to maintain sane line lengths.
E.g. https://google.github.io/styleguide/docguide/style.html#character-line-limit: «_Markdown content follows the residual convention of an 80-character line limit. Why? Because it’s what most of us do for code._»
Also:
- https://dsccommunity.org/styleguidelines/markdown-files/#line-length-not-longer-than-80-characters
- https://learn.microsoft.com/en-us/powershell/scripting/community/contributing/general-markdown?view=powershell-7.5#limit-line-length-to-100-characters
- 

> Just linting, because it feels.. odd

Well, for me the overly long lines feels odd 😺

lib/param_parsing/toml.go
```diff
@@ -48,25 +49,46 @@ func getParamsTOML(params Params) (Params, error) {
...
 				continue
 			}
+
 			if viperParser.Get(toml) != nil {
-				configKeyValue := viperParser.GetString(toml)
+				configKeyValue := viperParser.Get(toml)
+
+				if reflect.TypeOf(configKeyValue).Kind() != ptype {
+					logger.Warnf(
+						"TOML key %q is not a %s but a %s, skipping assignment of %q patrameter from TOML",
```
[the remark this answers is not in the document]
yermulnik: Resolved via 5cca58b 👍🏻

lib/param_parsing/toml.go
```diff
@@ -23,16 +23,17 @@ func getParamsTOML(params Params) (Params, error) {
 		viperParser.SetConfigName(configfileName)
 		viperParser.AddConfigPath(params.TomlDir)
 
-		errs := viperParser.ReadInConfig() // Find and read the config file
-		if errs != nil {
-			logger.Errorf("Could not to read %q: %v", tomlPath, errs)
-			return params, errs
+		// Find and read the config file
+		if err := viperParser.ReadInConfig(); err != nil {
+			logger.Errorf("Could not to read %q: %v", tomlPath, err)
```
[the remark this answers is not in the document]
yermulnik: Resolved via 4d5f389

lib/param_parsing/parameters.go
```diff
@@ -72,31 +77,51 @@ func populateParams(params Params) Params {
...
+			logger = lib.InitLogger(params.LogLevel)
+			logger.Fatal("(init) Cannot force color and disable color at the same time. Please choose either of them.")
+		}
+
+		if params.ForceColor {
+			os.Setenv("FORCE_COLOR", "true")
+		} else if params.NoColor {
+			os.Setenv("NO_COLOR", "true")
+		}
+
```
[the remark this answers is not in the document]
yermulnik: 👎🏻

lib/param_parsing/environment_test.go
```diff
@@ -97,3 +101,83 @@ func TestGetParamsFromEnvironment_log_level_from_env(t *testing.T) {
...
+		goCommandArgs := []string{"run", "../../main.go", "--dry-run", "1.10.5"}
+
+		t.Logf("Testing %q env var", envVarName)
+
+		out, err := exec.Command("go", goCommandArgs...).CombinedOutput()
+		if err != nil {
+			t.Fatalf("Unexpected failure: \"%v\", output: %q", err, string(out))
+		}
+
+		_ = os.Unsetenv(envVarName)
```
MatthewJohn: Does t.Fatal stop all further tests?
Should this not be unset even in L140 branch?

lib/param_parsing/environment_test.go
```diff
@@ -97,3 +101,83 @@ func TestGetParamsFromEnvironment_log_level_from_env(t *testing.T) {
...
+	} else {
+		t.Log("Success: no ANSI color codes in output")
+	}
+}
+
+func TestForceColorEnvVar(t *testing.T) {
+	envVarName := "FORCE_COLOR"
+	if color.SupportColor() {
+		_ = os.Setenv(envVarName, "true")
+		goCommandArgs := []string{"run", "../../main.go", "--dry-run", "1.10.5"}
```
MatthewJohn: Is there anything more simple that could be run? It'll be one of those things that whenever there's a test failure for a chunk of the "main" functionality, the color logging tests will also fail

lib/param_parsing/environment_test.go
```diff
@@ -97,3 +101,83 @@ func TestGetParamsFromEnvironment_log_level_from_env(t *testing.T) {
...
+			t.Fatalf("Unexpected failure: \"%v\", output: %q", err, string(out))
+		}
+
+		if !matched {
+			t.Errorf("Expected ANSI color codes in output, but found none: %q", string(out))
+		} else {
+			t.Log("Success: found ANSI color codes in output")
+		}
+	} else {
+		t.Skipf("Skipping test for %q env var as terminal doesn't support colors", envVarName)
```
MatthewJohn: Does this get skipped in github actions? How does it behave with the SupportColor?

lib/param_parsing/environment_test.go
```diff
@@ -97,3 +101,83 @@ func TestGetParamsFromEnvironment_log_level_from_env(t *testing.T) {
...
+	if matched {
+		t.Errorf("Expected no ANSI color codes in output, but found some: %q", string(out))
+	} else {
+		t.Log("Success: no ANSI color codes in output")
+	}
+}
+
+func TestForceColorEnvVar(t *testing.T) {
+	envVarName := "FORCE_COLOR"
+	if color.SupportColor() {
```
MatthewJohn: Could this logic be inverted and just return early to avoid putting everyuthing into this scope?

lib/param_parsing/environment_test.go
```diff
@@ -97,3 +101,83 @@ func TestGetParamsFromEnvironment_log_level_from_env(t *testing.T) {
...
+	envVarNameNoColor := "NO_COLOR"
+	_ = os.Setenv(envVarNameNoColor, "true")
+
+	expectedOutput := "FATAL (env) Cannot force color and disable color at the same time. Please choose either of them."
+
+	goCommandArgs := []string{"run", "../../main.go", "--dry-run", "1.10.5"}
+
+	t.Logf("Testing %q and %q env vars both present", envVarNameForceColor, envVarNameNoColor)
+
+	out, _ := exec.Command("go", goCommandArgs...).CombinedOutput() // nolint:errcheck // We want to test the output even if it fails
```
MatthewJohn, line 171: Do we need to check error? If it's expected to error, can ensure this?

lib/param_parsing/parameters.go
```diff
@@ -72,31 +77,51 @@ func populateParams(params Params) Params {
...
+	getopt.BoolVarLong(&params.ListAllFlag, "list-all", 'l', "List all versions of terraform - including beta and rc")
+	getopt.BoolVarLong(&params.NoColor, "no-color", 'k', "Disable color output. Useful for piping output to a file or when the terminal does not support colors")
+	getopt.BoolVarLong(&params.ShowLatestFlag, "show-latest", 'U', "Show latest stable version")
 	getopt.BoolVarLong(&params.VersionFlag, "version", 'v', "Displays the version of tfswitch")
 
 	// Parse the command line parameters to fetch stuff like chdir
 	getopt.Parse()
 
-	isShortRun := !params.VersionFlag && !params.HelpFlag
+	isNotShortRun := !params.VersionFlag && !params.HelpFlag
```
MatthewJohn, line 107: What's changed here, am I missing soemthing?

MatthewJohn: Oh, the variable has changed name - so it was just wrongly named the whole time?

lib/param_parsing/parameters.go
```diff
@@ -108,11 +133,39 @@ func populateParams(params Params) Params {
 			if err != nil {
 				logger.Fatalf("Failed to obtain settings from TOML config in home directory: %v", err)
 			}
+
+			if params.ForceColor && params.NoColor {
```
MatthewJohn, line 137: Can this check of params.ForceColor, NoColor to 1) set env variables and 2) fail if both are set, not be done once after all of the params have been gathered?
If there is a reason for this (more obvious errors as to where the configs have come from, if so, could this just be moved into a function?

It would seem that just updating params like normal: env, toml etc. Then after all this, check params and exit/update env variables. If it's so that logging coloring is updated _as soon as_ it finds the param, but it's quite a bit of logic littered about for, what I'd say, relatively little improvement and just becomes confusing tech debt

MatthewJohn: Plus.. if you have an env variable that sets coloring, but TOML file that unsets it.. you'd end with:
 * tfswitch starts with coloring enabled
 * read env - Logger re-initiliased with color enabled
 * read TOML - Logger re-initiliased with color disabled
 * read env (second pass through) - Logger re-initiliased with color enabled
and debug logs would switch back and forth.

I'd imagine that just saying, during params init, that coloring is _always_ enabled (or even disabled)

www/docs/usage/config-files.md
```diff
@@ -90,10 +113,11 @@ or
 product = "terraform"
 ```
 
-## Setting log level using `.tfswitch.toml` file
+### Setting log level
```
MatthewJohn, line 116: I'm guessing you're updating these to _just_ be configs (not TOML specific) and removed the environment variables section for this.. but I can't see the environment variables being mentioned in these (which I'm guessing they were meant to be?)

www/docs/usage/config-files.md
```diff
@@ -36,49 +42,66 @@ terraform {
...
+The `.tfswitch.toml` file can be configured with a `default-version` parameter
+for `tfswitch` to use a particular version, if no other sources of versions are
+found
 
 ```toml
 default-version = "1.5.4"
 ```
 
-## Setting product using `.tfswitch.toml` file
+### Setting product (`terarform` or `opentofu`)
```
MatthewJohn: I'm not sure about having the valid options in the title, as if we support a third product, it would break all of the links.
Also, typo in `terraform` :)

www/docs/usage/config-files.md
MatthewJohn: Oh, I stand corrected, environment variables just got moved.
Think I suggested it a while back, but to me, would make sense to _just_ have environment variables mixed in with the configs.
I.e. 
```
## Setting log level

Log level can be adjusted blah blah.
Defaults to X.
Can be set in .tfswitch.toml by setting log-level
Environment variable: `LOG_LEVEL`
```

MatthewJohn (changes requested):

lib/param_parsing/environment_test.go
yermulnik: > Does t.Fatal stop all further tests?

Not really. It fails current test and continues at the next test.

> Should this not be unset even in L140 branch?

Yeah, that makes sense. Thanks. Addressing this and the same by moving `os.Unsetenv()` to right before `if`: 9f7fc72

lib/param_parsing/environment_test.go
yermulnik: > Is there anything more simple that could be run?

Not sure I can imagine anything that would be simpler 🤷🏻 

> It'll be one of those things that whenever there's a test failure for a chunk of the "main" functionality, the color logging tests will also fail

Hopefully the error logged by this test in such a case, would give a hint that "some other" test is the issue 🤞🏻

lib/param_parsing/environment_test.go
yermulnik: > Does this get skipped in github actions?

Yes
```
2025-06-29T01:10:56.6664831Z === RUN   TestForceColorEnvVar
2025-06-29T01:10:56.6665348Z     environment_test.go:157: Skipping test for "FORCE_COLOR" env var as terminal doesn't support colors
2025-06-29T01:10:56.6665779Z --- SKIP: TestForceColorEnvVar (0.00s)
```

> How does it behave with the SupportColor?

Not sure I got the question. Would you mind rewording maybe?
At a sneak peak, if underlying library detects that terminal doesn't support color (escape) sequences, color output is not enabled.

lib/param_parsing/environment_test.go
yermulnik: Yep, good point. Addressing: c783ac3

lib/param_parsing/environment_test.go
yermulnik: If the error isn't happening, the `expectedOutput` check below will trigger test to error out. And the expected output value also includes the `FATAL` prefix from `log.Fatal`.
Do you think it is best to check `if err != nil` anyways? My idea was to avoid checking the same technically twice.
Let me know.

lib/param_parsing/parameters.go
yermulnik: Well, technically yes. I guess at some point in time I reverted the condition to streamline it, but missed to rename the var. Technically this has no effect on the underlying functionality.

lib/param_parsing/parameters.go
yermulnik: > Can this check of params.ForceColor, NoColor to 1) set env variables and 2) fail if both are set, not be done once after all of the params have been gathered?

I meant to fail as early as possible when mutually exclusive parameters are requested. The tool throws lots of diags (especially with verbose logging levels) and I see no reason to continue with e.g. color output enabled when user wanted no color but mistakenly also request color output. Does this make sense to you?

> if so, could this just be moved into a function?

You mean the `if params.ForceColor && params.NoColor { logger.Fatal(…) }` thing? 😕 While I agree it looks as a candidate for a function as it is used three times, but each time it has its own context: `init`, `env`, and `toml`. And hence I'm not really sure it is worth of a dedicated function. I could think of moving repeated error test to a variable (which still looks a bit of overengineering to me though) — do you think it is worth of it? 🤔 

> It would seem that just updating params like normal: env, toml etc. Then after all this, check params and exit/update env variables.

I don't think so. If the user wants color to be disabled, I see it not really straightforward to disable it at an arbitrary place from user perspective.

> If it's so that logging coloring is updated _as soon as_ it finds the param, but it's quite a bit of logic littered about for, what I'd say, relatively little improvement and just becomes confusing tech debt

Well... that's not the "_little improvement_" that causes so much "_quite a bit of logic littered about for_" to get it implemented, but the existing codebase that I needed to shove this feature into to get things done so that the feature works as expected.
I'm not really keen starting yet another round of refactoring just to implement pretty much standard feature 🤷🏻 I'm just trying to keep this thing rolling while I'm a way not a Golang programmer and hence I **have** to learn on the go. The so called "development on best effort basis", you know. Hopefully, together we'll be able to streamline all the kludgy bits later on 🤞🏻

www/docs/usage/config-files.md
yermulnik: > Think I suggested it a while back, but to me, would make sense to _just_ have environment variables mixed in with the configs.

Yeah, that may make sense.
I was trying to follow the existing format and split out Env var as they already have grown enough for their own page.
If you see a direction for improvement on how the docs are organized historically (when neither of us two were part of this project yet), please file an issue and we can re-organize it later on. Deal?

www/docs/usage/config-files.md
yermulnik: Ah, good catch 👍🏻 Also found a bit more of typos in that file. Fixing them and renaming section. Thanks.
Addressed: cb5901b — please see if it looks better now.

lib/param_parsing/parameters.go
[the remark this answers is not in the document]
yermulnik: Those check are repeatedly a bit **different** 🤷🏻

lib/param_parsing/parameters.go
```diff
@@ -72,31 +77,51 @@ func populateParams(params Params) Params {
...
 
-	isShortRun := !params.VersionFlag && !params.HelpFlag
+	isNotShortRun := !params.VersionFlag && !params.HelpFlag
+
+	if isNotShortRun {
+		if params.ForceColor && params.NoColor {
+			os.Setenv("NO_COLOR", "true") // Prefer no color in this case
+			logger = lib.InitLogger(params.LogLevel)
+			logger.Fatal("(init) Cannot force color and disable color at the same time. Please choose either of them.")
+		}
```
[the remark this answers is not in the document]
yermulnik: They are slightly different and I see not good reason to try and create a function to cover each different check, even despite they have part of the error message in common.

lib/param_parsing/environment_test.go
MatthewJohn: 👍

lib/param_parsing/environment_test.go
MatthewJohn: It's probably fine :)

MatthewJohn (approved):

PR #597, opened 2025-07-04
chore(`Makefile`): Improve `super-linter` target for a plain repo and a worktree

yermulnik:


Makefile
```diff
@@ -92,7 +92,7 @@ else
 		--env GITHUB_WORKSPACE="$(shell git rev-parse --show-toplevel)" \
 		--env RUN_LOCAL=true \
 		--env VALIDATE_ALL_CODEBASE=false \
-		--env FILTER_REGEX_EXCLUDE='^$$PWD/test-data/' \
+		--env FILTER_REGEX_EXCLUDE="^$(CURDIR)/test-data/" \
```
MatthewJohn: I assume this will break if the user has some regex-matching characters in their path.
Probably okay, but maybe a bit brittle

MatthewJohn: Wait, shouldn't this just be:
```suggestion
		--env FILTER_REGEX_EXCLUDE="^$(shell git rev-parse --show-toplevel)/test-data/" \
```
for consistency?
With this in mind, is it worth setting the paths (L88 and L89 and L90, L92) to variables to avoid the duplicate calls?

MatthewJohn (approved):

Makefile
yermulnik: > Wait, shouldn't this just be:
> 
> for consistency? With this in mind, is it worth setting the paths (L88 and L89 and L90, L92) to variables to avoid the duplicate calls?

Yeah, thanks for catching this!

> I assume this will break if the user has some regex-matching characters in their path.

It won't. But the `git rev-parse --show-toplevel` is what it was meant to be.

yermulnik: While I was addressing your suggestion, I found another issue.
Fixed in latest commit and it looks to be working as expected now including your worthwhile suggestion.
Please give it another look now. Thanks.

Makefile
```diff
@@ -10,6 +10,16 @@ GOOS             ?= $(shell $(GOBINARY) env GOOS)
 GOARCH           ?= $(shell $(GOBINARY) env GOARCH)
 CONTAINER_ENGINE ?= $(shell command -v podman || command -v docker || echo "NONE")
 
+# Git dirs for `super-linter` target
+GIT_COMMON_DIR   := $(shell git rev-parse --git-common-dir)
+GIT_DIR          := $(shell git rev-parse --git-dir)
+GIT_TOPLEVEL_DIR := $(shell git rev-parse --show-toplevel)
+# Assume that if GIT_COMMON_DIR difers from GIT_DIR, then we're in Git worktree
```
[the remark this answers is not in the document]
yermulnik: 226cd65

Makefile
```diff
@@ -86,13 +96,12 @@ else
 	# Keep `--env' vars below the `VALIDATE_ALL_CODEBASE' in sync with .github/workflows/super-linter.yml
 	$(CONTAINER_ENGINE) run \
 		--name super-linter \
-		--volume "$(shell git rev-parse --show-toplevel):$(shell git rev-parse --show-toplevel)" \
-		--volume "$(shell git rev-parse --git-common-dir):$(shell git rev-parse --git-common-dir)" \
-		--workdir "$(shell git rev-parse --show-toplevel)" \
-		--env GITHUB_WORKSPACE="$(shell git rev-parse --show-toplevel)" \
+		--volume "$(GIT_TOPLEVEL_DIR):$(GIT_TOPLEVEL_DIR)" $(GIT_COMMON_DIR_VOLUME)\
```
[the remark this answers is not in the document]
yermulnik: When it is placed on its own line, the output looks ugly in plain repo. So, assuming I'm one of not many using Git worktrees, I'd leave the ugly look for when it is run in Git worktree.

Makefile
yermulnik: @MatthewJohn

yermulnik: I'll merge it now. Please let me know if you find anything else to improve. Thanks.

PR #601, opened 2025-07-14
fix: Commad-line flags to show/install latest stable versions are broken

### LINKED ISSUE #585: Commad-line flags to show/install latest stable versions are broken
```bash
> tfswitch --version
Version: v1.4.5

> tfswitch --help
[…]
 -s, --latest-stable=value
                    Latest implicit version based on a constraint. Ex: tfswitch
                    --latest-stable 0.13.0 downloads 0.13.7 and 0.13 downloads
                    0.15.5 (latest)
 -S, --show-latest-stable=value
                    Show latest implicit version. Ex: tfswitch
                    --show-latest-stable 0.13 prints 0.13.7 (latest)
[…]

> tfswitch 1.10.5
01:37:29.351 INFO Reading configuration from "/home/georgii/.tfswitch.toml"
01:37:29.353 INFO Switched Terraform to version "1.10.5"

> tfswitch -s 1.9
01:38:57.361 INFO Reading configuration from "/home/georgii/.tfswitch.toml"
01:38:57.464 INFO Matched version: "1.12.1"
01:38:57.464 INFO Switched Terraform to version "1.12.1"

> tfswitch -S 1.9
01:39:26.223 INFO Reading configuration from "/home/georgii/.tfswitch.toml"
01:39:26.311 INFO Matched version: "1.12.1"
1.12.1
```

yermulnik:
Resolves #585

### DIFF
--- a/lib/list_versions.go
+++ b/lib/list_versions.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"reflect"
 	"regexp"
+	"slices"
 	"strings"
 )
 
@@ -16,20 +17,25 @@ type tfVersionList struct {
 
 // Semantic version regexes without `^` and `$` anchors
 // Follows https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
-var regexSemVer = map[string]string{
-	"full":             `(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`,
-	"minor":            `(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)`,
-	"patch":            `(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)`,
-	"preReleaseSuffix": `\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`,
+var regexSemVer = struct {
+	Full             *regexp.Regexp
+	Minor            *regexp.Regexp
+	Patch            *regexp.Regexp
+	PreReleaseSuffix *regexp.Regexp
+}{
+	Full:             regexp.MustCompile(`(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`),
+	Minor:            regexp.MustCompile(`(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)`),
+	Patch:            regexp.MustCompile(`(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)`),
+	PreReleaseSuffix: regexp.MustCompile(`\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`),
 }
 
 func getVersionsFromBody(body string, preRelease bool, tfVersionList *tfVersionList) {
 	var semver string
 	// Without the ending '"' pre-release folders would be tried and break.
 	if preRelease {
-		semver = `\/?` + regexSemVer["full"] + `/?"`
+		semver = `\/?` + regexSemVer.Full.String() + `/?"`
 	} else if !preRelease {
-		semver = `\/?` + regexSemVer["patch"] + `\/?"`
+		semver = `\/?` + regexSemVer.Patch.String() + `\/?"`
 	}
 	r, err := regexp.Compile(semver)
 	if err != nil {
@@ -70,7 +76,7 @@ func getTFLatest(mirrorURL string) (string, error) {
 		return "", err
 	}
 	// Getting versions from body; should return match /X.X.X/ where X is a number
-	semver := `\/?` + regexSemVer["patch"] + `\/?"`
+	semver := `\/?` + regexSemVer.Patch.String() + `\/?"`
 	r, errSemVer := regexp.Compile(semver)
 	if errSemVer != nil {
 		return "", fmt.Errorf("Error compiling %q regex: %v", semver, errSemVer)
@@ -95,7 +101,7 @@ func getTFLatestImplicit(mirrorURL string, preRelease bool, version string) (str
 			return "", err
 		}
 		// Getting versions from body; should return match /X.X.X-@/ where X is a number,@ is a word character between a-z or A-Z
-		semver := `\/?` + version + regexSemVer["preReleaseSuffix"] + `\/?"`
+		semver := `\/?` + version + regexSemVer.PreReleaseSuffix.String() + `\/?"`
 		r, errReSemVer := regexp.Compile(semver)
 		if errReSemVer != nil {
 			return "", errReSemVer
@@ -191,29 +197,25 @@ func removeDuplicateVersions(elements []string) []string {
 }
 
 // validVersionFormat : Return true if valid semantic version provided based on the type of version requested
-func validVersionFormat(version string, reqType ...string) bool {
-	var verType string
+func validVersionFormat(version string, validation ...*regexp.Regexp) bool {
 	var semverRegex *regexp.Regexp
 
-	switch len(reqType) {
+	switch len(validation) {
 	case 0:
-		verType = "full"
+		semverRegex = regexSemVer.Full
 	case 1:
-		verType = reqType[0]
+		semverRegex = validation[0]
+		// Limit the valid values in this switch case to only minor and patch versions,
+		// as currently it is used in `ShowLatestImplicitVersion()` only
+		validValues := []*regexp.Regexp{regexSemVer.Minor, regexSemVer.Patch}
+		if !slices.Contains(validValues, semverRegex) {
+			logger.Fatalf("Internal error: invalid \"validation\" argument value")
+		}
 	default:
-		logger.Fatalf("Internal error: invalid number of arguments (must be 2, got %d)", 1+len(reqType))
+		logger.Fatalf("Internal error: invalid number of arguments (must be 2, got %d)", 1+len(validation))
 	}
 
-	switch verType {
-	case "full":
-		semverRegex = regexp.MustCompile(`^` + regexSemVer["full"] + `$`)
-	case "minor":
-		semverRegex = regexp.MustCompile(`^` + regexSemVer["minor"] + `$`)
-	case "patch":
-		semverRegex = regexp.MustCompile(`^` + regexSemVer["patch"] + `$`)
-	default:
-		logger.Fatalf("Internal error: \"type\" argument must be one of \"full\", \"minor\", or \"patch\" (got %q)", reqType)
-	}
+	semverRegex = regexp.MustCompile(`^` + semverRegex.String() + `$`)
 
 	return semverRegex.MatchString(version)
 }
@@ -230,7 +232,7 @@ func ShowLatestVersion(mirrorURL string) {
 
 // ShowLatestImplicitVersion : show latest implicit version given the mirror URL
 func ShowLatestImplicitVersion(requestedVersion, mirrorURL string, preRelease bool) {
-	if validVersionFormat(requestedVersion, "minor") || (validVersionFormat(requestedVersion, "patch") && !preRelease) {
+	if validVersionFormat(requestedVersion, regexSemVer.Minor) || (validVersionFormat(requestedVersion, regexSemVer.Patch) && !preRelease) {
 		tfversion, err := getTFLatestImplicit(mirrorURL, preRelease, requestedVersion)
 		if err != nil {
 			logger.Fatalf("Error getting latest implicit version %q from %q: %v", requestedVersion, mirrorURL, err)
--- a/lib/list_versions.go
+++ b/lib/list_versions.go
@@ -212,7 +212,7 @@ func validVersionFormat(version string, validation ...*regexp.Regexp) bool {
 			logger.Fatalf("Internal error: invalid \"validation\" argument value")
 		}
 	default:
-		logger.Fatalf("Internal error: invalid number of arguments (must be 2, got %d)", 1+len(validation))
+		logger.Fatalf("Internal error: invalid number of arguments (must be 1 or 2, got %d)", 1+len(validation))
 	}
 
 	semverRegex = regexp.MustCompile(`^` + semverRegex.String() + `$`)
--- a/lib/list_versions.go
+++ b/lib/list_versions.go
@@ -7,7 +7,6 @@ import (
 	"net/http"
 	"reflect"
 	"regexp"
-	"slices"
 	"strings"
 )
 
@@ -197,6 +196,7 @@ func removeDuplicateVersions(elements []string) []string {
 }
 
 // validVersionFormat : Return true if valid semantic version provided based on the type of version requested
+// Caveat: Passing no validation argument validates the full semantic version format to provide backward compatibility
 func validVersionFormat(version string, validation ...*regexp.Regexp) bool {
 	var semverRegex *regexp.Regexp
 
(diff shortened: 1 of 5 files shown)


lib/list_versions.go
```diff
@@ -183,34 +190,35 @@ func removeDuplicateVersions(elements []string) []string {
...
+	var verType string
+	var semverRegex *regexp.Regexp
+
+	switch len(reqType) {
+	case 0:
+		verType = "full"
+	case 1:
+		verType = reqType[0]
+	default:
+		logger.Fatalf("Internal error: invalid number of arguments (must be 2, got %d)", 1+len(reqType))
```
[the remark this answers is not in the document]
yermulnik: > The fatal error message for `validVersionFormat` claims “must be 2” arguments, but the function actually accepts 0 or 1 optional `reqType` parameters.

Isn't true.

lib/list_versions.go
```diff
@@ -14,15 +14,22 @@ type tfVersionList struct {
...
-		// Getting versions from body; should return match /X.X.X-@/ where X is a number,@ is a word character between a-z or A-Z
-		semver = `\/?(\d+\.\d+\.\d+)(-[a-zA-z]+\d*)?/?"`
+		semver = `\/?` + regexSemVer["full"] + `/?"`
 	} else if !preRelease {
-		// Getting versions from body; should return match /X.X.X/ where X is a number
-		// without the ending '"' pre-release folders would be tried and break.
-		semver = `\/?(\d+\.\d+\.\d+)\/?"`
+		semver = `\/?` + regexSemVer["patch"] + `\/?"`
 	}
 	r, err := regexp.Compile(semver)
```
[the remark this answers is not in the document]
yermulnik: Good catch. Though this would require larger re-factoring of this lib and I'd prefer to do it sometime later 🤷🏻

lib/list_versions.go
```diff
@@ -14,15 +14,22 @@ type tfVersionList struct {
 	tflist []string
 }
 
+// Semantic version regexes without `^` and `$` anchors
+// Follows https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
+var regexSemVer = map[string]string{
```
MatthewJohn: I do love a good complex regex... and this wall of text is satisfies that

MatthewJohn: On another note, because this is just a map, it doesn't allow for any static checking.
e.g.:
```
var regexSemVer = map[string]string{
	"blahy": "yes",
}

func main() {
	println(regexSemVer["blahay"])
}

```
Runs fine and just outputs an empty string, which isn't great.

I'd suggest doing:
```
var regexSemVer = struct {
	Full             string
	Minor            string
	Patch            string
	PreReleaseSuffix string
}{
	Full:             `(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`,
	Minor:            `(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)`,
	Patch:            `(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)`,
	PreReleaseSuffix: `\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`,
}
```

The functions that pass what type of validation they want to do can still pass this.

e.g.:
```
func validVersionFormat(version string, validation ...string) bool {
	var validationString string
	var semverRegex *regexp.Regexp

	switch len(validation) {
	case 0:
		validationString = regexSemVer.Full
	case 1:
		validationString = validation[0]
	default:
		logger.Fatalf("Internal error: invalid number of arguments (must be 2, got %d)", 1+len(validation))
	}

	regexp.MustCompile(`^` + validationString + `$`)

	return semverRegex.MatchString(version)
}
```
This makes the logic of this function easier and more bullet-proof.
Also calling it isn't less verbose:
```
validVersionFormat(version, regexSemVer.Patch)
```

Hope it doesn't seem knit-picky, I'd just be conserned that a typo in the map indexing could easily go uncaught and that function becomes more verbose (as it needs to handle each index of the map for validation and there's some duplication as well, which is tidied up

MatthewJohn: Could even just pre-compile into the struct:
```
var regexSemVer = struct {
	Full             *regexp.Regexp
	Minor            *regexp.Regexp
	Patch            *regexp.Regexp
	PreReleaseSuffix *regexp.Regexp
}{
	Full:             regexp.MustCompile(`(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`),
	Minor:            regexp.MustCompile(`(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)`),
	Patch:            regexp.MustCompile(`(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)`),
	PreReleaseSuffix: regexp.MustCompile(`\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`),
}

func validVersionFormat(version string, validation ...*regexp.Regexp) bool {
	var semverRegex *regexp.Regexp

	switch len(validation) {
	case 0:
		semverRegex = regexSemVer.Full
	case 1:
		semverRegex = validation[0]
	default:
		logger.Fatalf("Internal error: invalid number of arguments (must be 2, got %d)", 1+len(validation))
	}

	return semverRegex.MatchString(version)
}
```

lib/list_versions.go
```diff
@@ -183,34 +190,35 @@ func removeDuplicateVersions(elements []string) []string {
...
 
-// ValidMinorVersionFormat : returns valid MINOR version format
-/* For example: 0.1 = valid
-// For example: a.1.2 = invalid
-// For example: 0.1.2 = invalid
-*/
-func validMinorVersionFormat(version string) bool {
-	// Getting versions from body; should return match /X.X./ where X is a number
-	semverRegex := regexp.MustCompile(`^(\d+\.\d+)$`)
+	switch verType {
```
MatthewJohn: As suggested, this seems quite overly verbose and relying on magic strings (and duplicates the names of the keys in the map).

yermulnik: Won't be able to get rid of magic strings at this moment as regexps are used with different anchors throughout this file. The rest of comments and ideas from the other thread are being worked on towards implementation.

lib/list_versions.go
yermulnik: Appreciate this much useful feedback 👍🏻 Working on implementation.

yermulnik: Hopefully I did as much as I could: 5c4ca9d
Please review.

lib/param_parsing/parameters.go
```diff
@@ -83,13 +83,13 @@ func populateParams(params Params) Params {
...
-	getopt.StringVarLong(&params.ShowLatestPre, "show-latest-pre", 'P', "Show latest pre-release implicit version. Ex: tfswitch --show-latest-pre 0.13 prints 0.13.0-rc1 (latest)")
-	getopt.StringVarLong(&params.ShowLatestStable, "show-latest-stable", 'S', "Show latest implicit version. Ex: tfswitch --show-latest-stable 0.13 prints 0.13.7 (latest)")
 	getopt.StringVarLong(&params.Product, "product", 't', fmt.Sprintf("Specifies which product to use. Ex: `tfswitch --product opentofu` will install OpenTofu. Options: (%s). Default: %s", strings.Join(productIds, ", "), lib.DefaultProductId))
+	getopt.StringVarLong(&params.ShowLatestPre, "show-latest-pre", 'P', "Show latest pre-release implicit version. Accepts #.# as argument where # are numbers. Ex: tfswitch --show-latest-pre 0.13 prints latest in 0.13.x branch up to pre-release (if any)")
+	getopt.StringVarLong(&params.ShowLatestStable, "show-latest-stable", 'S', "Show latest stable implicit version. Accepts either #.# or #.#.# as argument where # are numbers. Ex: tfswitch --show-latest-stable 0.13.0 prints latest in 0.13.x branch, 0.13 prints latest in 0.x branch, 1.1 prints latest in 1.x branch")
```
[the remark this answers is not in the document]
yermulnik: irrelevant comment

lib/list_versions.go
```diff
@@ -220,9 +230,9 @@ func ShowLatestVersion(mirrorURL string) {
 	fmt.Printf("%s\n", tfversion)
 }
 
-// ShowLatestImplicitVersion show latest - argument (version) must be provided
+// ShowLatestImplicitVersion : show latest implicit version given the mirror URL
 func ShowLatestImplicitVersion(requestedVersion, mirrorURL string, preRelease bool) {
-	if validMinorVersionFormat(requestedVersion) {
+	if validVersionFormat(requestedVersion, regexSemVer.Minor) || (validVersionFormat(requestedVersion, regexSemVer.Patch) && !preRelease) {
```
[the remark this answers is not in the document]
yermulnik: irrelevant

lib/list_versions_test.go
```diff
@@ -165,11 +165,8 @@ func TestGetTFLatest(t *testing.T) {
 func TestGetTFLatestImplicit(t *testing.T) {
 	logger = InitLogger("DEBUG")
 	tName := "version=%s_preRelease=%v"
-	t.Run(fmt.Sprintf(tName, "0.11.0", false), func(t *testing.T) { testGetTFLatestImplicit(t, "0.11.0", false, "0.11.13") })
-	// @TODO Couldn't get this test working
-	// t.Run(fmt.Sprintf(tName, "0.11", true), func(t *testing.T) { testGetTFLatestImplicit(t, "0.11", true, "0.11.13") })
+	t.Run(fmt.Sprintf(tName, "0.11.0", false), func(t *testing.T) { testGetTFLatestImplicit(t, "0.12.0", false, "0.12.2") })
```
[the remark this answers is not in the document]
yermulnik: Thanks. Addressed: b4d8123

lib/list_versions.go
```diff
@@ -183,34 +196,31 @@ func removeDuplicateVersions(elements []string) []string {
...
+	case 1:
+		semverRegex = validation[0]
+		// Limit the valid values in this switch case to only minor and patch versions,
+		// as currently it is used in `ShowLatestImplicitVersion()` only
+		validValues := []*regexp.Regexp{regexSemVer.Minor, regexSemVer.Patch}
+		if !slices.Contains(validValues, semverRegex) {
+			logger.Fatalf("Internal error: invalid \"validation\" argument value")
+		}
+	default:
+		logger.Fatalf("Internal error: invalid number of arguments (must be 2, got %d)", 1+len(validation))
```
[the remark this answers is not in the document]
yermulnik: Thanks. Addressed: 6a51376

lib/list_versions.go
```diff
@@ -183,34 +196,31 @@ func removeDuplicateVersions(elements []string) []string {
...
-*/
-func validVersionFormat(version string) bool {
-	// Getting versions from body; should return match /X.X.X-@/ where X is a number,@ is a word character between a-z or A-Z
-	// Follow https://semver.org/spec/v1.0.0-beta.html
-	// Check regular expression at https://rubular.com/r/ju3PxbaSBALpJB
-	semverRegex := regexp.MustCompile(`^(\d+\.\d+\.\d+)(-[a-zA-z]+\d*)?$`)
-	return semverRegex.MatchString(version)
-}
+// validVersionFormat : Return true if valid semantic version provided based on the type of version requested
+func validVersionFormat(version string, validation ...*regexp.Regexp) bool {
```
MatthewJohn, line 200: I guess noting that if you _don't_ pass validation, it uses Full, but you can't pass full 🤔

lib/list_versions.go
```diff
@@ -183,34 +196,31 @@ func removeDuplicateVersions(elements []string) []string {
...
 
-// ValidMinorVersionFormat : returns valid MINOR version format
-/* For example: 0.1 = valid
-// For example: a.1.2 = invalid
-// For example: 0.1.2 = invalid
-*/
-func validMinorVersionFormat(version string) bool {
-	// Getting versions from body; should return match /X.X./ where X is a number
-	semverRegex := regexp.MustCompile(`^(\d+\.\d+)$`)
+	semverRegex = regexp.MustCompile(`^` + semverRegex.String() + `$`)
```
MatthewJohn, line 216: It's a shame that the original versions can't include the match of start/end, and require Compile(Compile().String()), but oh well, I think it's definitely safer as is :)

lib/list_versions_test.go
```diff
@@ -215,96 +213,117 @@ func TestRemoveDuplicateVersions(t *testing.T) {
...
-		t.Logf("Valid version format : %s (expected)", version)
+	if valid == false {
+		t.Logf("Invalid version format : %s (expected)", version)
 	} else {
 		t.Errorf("Failed to verify version format: %s\n", version)
 	}
 
-	version = "1.11.4-boom"
-
+	version = "1.11.4-01"
```
MatthewJohn, line 319: Why is : `version = "0.12.0-rc2"` valid, but this isn't? 🤔

lib/list_versions.go
yermulnik: Yeah, that's a hiccup. We use those regexps to parse HTML body for valid versions from Mirror URL and those need to be not wrapped in `^` and `$` anchors. E.g. https://github.com/warrensbox/terraform-switcher/blob/issue-585/fix-show-and-install-latest-stable/lib/list_versions.go#L32-L39
```go

func getVersionsFromBody(body string, preRelease bool, tfVersionList *tfVersionList) {
        var semver string
        // Without the ending '"' pre-release folders would be tried and break.
        if preRelease {
                semver = `\/?` + regexSemVer.Full.String() + `/?"`
        } else if !preRelease {
                semver = `\/?` + regexSemVer.Patch.String() + `\/?"`
        }
```
So I'm leaving to a future re-factoring/improvement.

lib/list_versions_test.go
yermulnik: https://semver.org/#spec-item-9 (see the text highlighted with bold font)
> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]. Identifiers MUST NOT be empty. **Numeric identifiers MUST NOT include leading zeroes**. […]

lib/list_versions.go
yermulnik: Hmm... I was trying to make the check as specific as possible, but... but I guess you;re right that it's a bit of overengineering. I'm going to update it now to make it more general-purpose and less strict.

yermulnik: @MatthewJohn eb5e3f6

lib/semver.go
```diff
@@ -58,7 +58,12 @@ func PrintInvalidTFVersion() {
 	logger.Error("Version does not exist or invalid terraform version format.\n\tFormat should be #.#.# or #.#.#-@# where # are numbers and @ are word characters.\n\tFor example, 1.11.7 and 0.11.9-beta1 are valid versions")
 }
 
-// PrintInvalidMinorTFVersion Print invalid minor TF version
+// PrintInvalidMinorTFVersion Print invalid minor version format
```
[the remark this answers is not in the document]
yermulnik: No trailing period is consistent with the current file's convention.

lib/semver.go
```diff
@@ -58,7 +58,12 @@ func PrintInvalidTFVersion() {
...
 func PrintInvalidMinorTFVersion() {
 	logger.Error("Invalid minor terraform version format.\n\tFormat should be #.# where # are numbers. For example, 1.11 is valid version")
 }
+
+// printInvalidVersionFormat Print invalid version format
```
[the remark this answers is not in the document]
yermulnik: No trailing period is consistent with the current file's convention.

lib/list_versions.go
```diff
@@ -14,15 +14,27 @@ type tfVersionList struct {
...
 
+// Semantic version regexes without `^` and `$` anchors
+// Follows https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
+var regexSemVer = struct {
+	Full             *regexp.Regexp
+	Minor            *regexp.Regexp
+	Patch            *regexp.Regexp
+	PreReleaseSuffix *regexp.Regexp
+}{
+	Full:             regexp.MustCompile(`(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`),
```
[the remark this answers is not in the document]
yermulnik: An interesting suggestion... 🤔 Though at this stage I see no need for lazy init as I don't think their compilation at binary compilation time is anything expensive and hence it isn't worth of effort.

lib/list_versions.go
```diff
@@ -183,34 +195,30 @@ func removeDuplicateVersions(elements []string) []string {
...
 
-// ValidMinorVersionFormat : returns valid MINOR version format
-/* For example: 0.1 = valid
-// For example: a.1.2 = invalid
-// For example: 0.1.2 = invalid
-*/
-func validMinorVersionFormat(version string) bool {
-	// Getting versions from body; should return match /X.X./ where X is a number
-	semverRegex := regexp.MustCompile(`^(\d+\.\d+)$`)
+	semverRegex = regexp.MustCompile(`^` + semverRegex.String() + `$`)
```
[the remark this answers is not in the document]
yermulnik: Yeah, we'll need to give this a thought and maybe re-factor later. Good job catching this bit 👍🏻 There's an response to the similar comment: https://github.com/warrensbox/terraform-switcher/pull/601#discussion_r2226071385

lib/list_versions.go
MatthewJohn: I approved, was just a passing comment :)

MatthewJohn (approved):

lib/list_versions.go
yermulnik: It was a good comment which was worth of rethinking this bit and I did it. Appreciate your worthwhile and helpful reviews 👍🏻

PR #602, opened 2025-07-15
fix: Rectify incomplete/incorrect logging library logic

### LINKED ISSUE #589: Rectify incomplete/incorrect logging library logic
* The logging library (`lib/logging.go`) is missing a dedicated `WARN` logging level
* Also it has `NOTICE` and `INFO` levels swapped as opposite to the order provided by [parent lib](https://github.com/gookit/slog/blob/master/common.go#L54-L74): `TRACE/DEBUG/NOTICE/INFO/WARN/ERROR/FATAL/DEBUG/PANIC` (ours) vs `TRACE/DEBUG/INFO/NOTICE/WARN/ERROR/FATAL/DEBUG/PANIC` (parent)
  * Fixing this will affect Normal logging as `NOTICE` should replace `INFO` in this type of logging configuration — need to consider how to reduce impact to UX (if applicable at all)
  * Consider adding a "fake" log level that allows turning logging completely off to support common practices: `OFF`
* ⚠ Once logging lib bits are streamlined, **update Bash completion script accordingly**
* Relates to #584
* Indirectly relates to #462

yermulnik:
Resolves #589


www/docs/usage/environment-variables.md
```diff
@@ -96,12 +96,17 @@ tfswitch # Will download actual binary to /var/cache/.terraform.versions/
...
 - Supported log levels:
-  - `ERROR`: includes `PANIC`, `FATAL`, `ERROR`
-  - `INFO`: includes `PANIC`, `FATAL`, `ERROR`, `WARN`, `INFO` (default)
-  - `NOTICE`: includes `PANIC`, `FATAL`, `ERROR`, `WARN`, `NOTICE`, `INFO`
-  - `DEBUG`: includes `PANIC`, `FATAL`, `ERROR`, `WARN`, `NOTICE`, `INFO`, `DEBUG`
-  - `TRACE`: includes `PANIC`, `FATAL`, `ERROR`, `WARN`, `NOTICE`, `INFO`, `DEBUG`, `TRACE`
-  - Any other log level value falls under default logging level
+  - `OFF`: disables (suppresses) logging
+  - `PANIC`: High severity, unrecoverable errors
+  - `FATAL`: Fatal, unrecoverable errors + previous log level
```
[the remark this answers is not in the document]
yermulnik: > [nitpick] The phrase `+ previous log level` is inconsistent with the plural `+ previous log levels` used elsewhere.

That's because the `FATAL` has only single included previous log level.

lib/logging.go
```diff
@@ -47,34 +66,66 @@ func NewStderrConsoleHandler(levels []slog.Level) *handler.ConsoleHandler {
...
+	// See `initParams()` in lib/param_parsing/parameters.go for default log level
+	h = NewStderrConsoleHandler(loggingLevel[fallbackLogLevel])
+
+	isUnknownLogLevel := false
+
+	if slices.Contains(LogLevels(), logLevel) {
+		h = NewStderrConsoleHandler(loggingLevel[logLevel])
+		if slices.Contains(useDebugTemplateLogLevels, logLevel) {
+			formatter.SetTemplate(loggingTemplateDebug)
+		}
```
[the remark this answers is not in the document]
yermulnik: The current implementation is well enough I think 🤷🏻

www/docs/usage/config-files.md
```diff
@@ -124,12 +124,17 @@ log-level = "INFO"
...
 - Supported log levels:
-  - `ERROR`: includes `PANIC`, `FATAL`, `ERROR`
-  - `INFO`: includes `PANIC`, `FATAL`, `ERROR`, `WARN`, `INFO` (default)
-  - `NOTICE`: includes `PANIC`, `FATAL`, `ERROR`, `WARN`, `NOTICE`, `INFO`
-  - `DEBUG`: includes `PANIC`, `FATAL`, `ERROR`, `WARN`, `NOTICE`, `INFO`, `DEBUG`
-  - `TRACE`: includes `PANIC`, `FATAL`, `ERROR`, `WARN`, `NOTICE`, `INFO`, `DEBUG`, `TRACE`
-  - Any other log level value falls under default logging level
+  - `OFF`: disables (suppresses) logging
+  - `PANIC`: High severity, unrecoverable errors
+  - `FATAL`: Fatal, unrecoverable errors + previous log level
```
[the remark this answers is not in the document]
yermulnik: > [nitpick] This entry uses singular `previous log level` while the others use plural.

That's because the `FATAL` has only single included previous log level.

lib/logging.go
```diff
@@ -10,14 +12,31 @@ import (
...
 	loggingTemplateDebug = "{{datetime}} {{level}} [{{caller}}] {{message}} {{data}} {{extra}}\n"
 	loggingTemplate      = "{{datetime}} {{level}} {{message}} {{data}} {{extra}}\n"
-	logger               *slog.Logger
-	ErrorLogging         = slog.Levels{slog.PanicLevel, slog.FatalLevel, slog.ErrorLevel}
-	NormalLogging        = slog.Levels{slog.PanicLevel, slog.FatalLevel, slog.ErrorLevel, slog.WarnLevel, slog.InfoLevel}
-	NoticeLogging        = slog.Levels{slog.PanicLevel, slog.FatalLevel, slog.ErrorLevel, slog.WarnLevel, slog.InfoLevel, slog.NoticeLevel}
-	DebugLogging         = slog.Levels{slog.PanicLevel, slog.FatalLevel, slog.ErrorLevel, slog.WarnLevel, slog.InfoLevel, slog.NoticeLevel, slog.DebugLevel}
-	TraceLogging         = slog.Levels{slog.PanicLevel, slog.FatalLevel, slog.ErrorLevel, slog.WarnLevel, slog.InfoLevel, slog.NoticeLevel, slog.DebugLevel, slog.TraceLevel}
+	// Parent lib: https://github.com/gookit/slog/blob/f857defd050dd7fc3c3013134cf50ed51b917a1f/common.go#L69-L88
+	loggingLevel = map[string]slog.Levels{
```
MatthewJohn: Map of strings :D

yermulnik: I already attempted to justify this in Slack =)
- The keys from this map are also used in cmdline arg description (is there a way to extract the same from struct so that we can put a list of available/allowed/legit values in cmdline arg description so that users know what they can use?)
- There's a code to handle unknown logging values (so we're protected against unknown/mistaken map values)
- These map keys are mapped from/to string values that user provides by means of config files and/or env vars (I can't figure out how to replace the map of strings by struct for the use case (for things like `loggingLevel[logLevel]`))

Let me know if you see a better approach (I'm away not a Go guy, I'm just a blind kitten poking around and hence I may (and potentially do) do lots of mistakes in Go code).

MatthewJohn: > I already attempted to justify this in Slack =)

Sorry, missed the reply - this comment was originally here (before my initial slack message), but I realised it was pending :face-palm:

yermulnik: @MatthewJohn Okay, I guess I sort of vibecoded this with help from Copilot, Google search and stroke of luck: d3880ca 🤞🏻

lib/logging.go
```diff
@@ -47,34 +79,82 @@ func NewStderrConsoleHandler(levels []slog.Level) *handler.ConsoleHandler {
...
+			h = NewStderrConsoleHandler(loggingLevel)
+			if slices.Contains(useDebugTemplateLogLevels, logLevel) {
+				formatter.SetTemplate(loggingTemplateDebug)
+			}
+		} else {
+			// Handle unkonwn log levels down the code once the logger is initialized
+			isUnknownLogLevel = true
+		}
+	} else {
+		// Handle unkonwn log levels down the code once the logger is initialized
```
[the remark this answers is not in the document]
yermulnik: bf94546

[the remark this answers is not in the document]
yermulnik: bf94546

lib/logging.go
yermulnik: I'm resolving this thread. Please feel free to unresolve it if any 👍🏻

lib/logging.go
```diff
@@ -47,34 +79,82 @@ func NewStderrConsoleHandler(levels []slog.Level) *handler.ConsoleHandler {
 	return NewStderrConsoleWithLF(slog.NewLvsFormatter(levels))
 }
 
+func LogLevels() []string {
```
MatthewJohn, line 82: This could be private:
```suggestion
func logLevels() []string {
```

lib/logging.go
```diff
@@ -47,34 +79,82 @@ func NewStderrConsoleHandler(levels []slog.Level) *handler.ConsoleHandler {
...
+	logLevel = strings.ToUpper(logLevel)
+
 	formatter := slog.NewTextFormatter()
-	formatter.EnableColor = isColorLogging()
 	formatter.ColorTheme = slog.ColorTheme
+	formatter.EnableColor = isColorLogging()
+	formatter.SetTemplate(loggingTemplate)
 	formatter.TimeFormat = "15:04:05.000"
 
+	useDebugTemplateLogLevels := []string{"DEBUG", "ERROR", "FATAL", "PANIC", "TRACE"}
```
MatthewJohn, line 107: So this is just anything other than the default?
I imagine that if someone sets to a lower level (e.g. from INFO, to FATAL/PANIC/ERROR) they're looking for less verbosity, so changing the formatting to a more verbose template might not be appropriate?

lib/logging.go
```diff
@@ -47,34 +79,82 @@ func NewStderrConsoleHandler(levels []slog.Level) *handler.ConsoleHandler {
...
+
+	isUnknownLogLevel := false
+
+	loggingLevelProp := logLevels.FieldByName(logLevel)
+	if loggingLevelProp.IsValid() {
+		loggingLevel, ok = logLevels.FieldByName(logLevel).Interface().(slog.Levels)
+		if ok {
+			h = NewStderrConsoleHandler(loggingLevel)
+			if slices.Contains(useDebugTemplateLogLevels, logLevel) {
+				formatter.SetTemplate(loggingTemplateDebug)
```
MatthewJohn, line 127: I suspect the template could just be an attribute of the `loggingLevels` struct? Maybe something for future?

lib/logging.go
```diff
@@ -47,34 +79,82 @@ func NewStderrConsoleHandler(levels []slog.Level) *handler.ConsoleHandler {
...
+	if !ok {
+		panic("Internal error: invalid fallback log level - " + fallbackLogLevel)
+	}
+	h = NewStderrConsoleHandler(loggingLevel)
+
+	isUnknownLogLevel := false
+
+	loggingLevelProp := logLevels.FieldByName(logLevel)
+	if loggingLevelProp.IsValid() {
+		loggingLevel, ok = logLevels.FieldByName(logLevel).Interface().(slog.Levels)
```
MatthewJohn: Do you want to call logLevels.FieldByName twice?
```suggestion
		loggingLevel, ok = loggingLevelProp.Interface().(slog.Levels)
```

lib/logging.go
yermulnik: This function is called from the `lib/param_parsing/parameters.go` and fails to be called if is private 🤷🏻 Let me know if there's a way to call it as a private function from there. Thanks.

lib/logging.go
yermulnik: > So this is just anything other than the default?

Not really.
The full list of log levels is: `DEBUG ERROR FATAL INFO NOTICE OFF PANIC TRACE WARN`.

> I imagine that if someone sets to a lower level (e.g. from INFO, to FATAL/PANIC/ERROR) they're looking for less verbosity, so changing the formatting to a more verbose template might not be appropriate?

I can see an advantage in a more verbose template for error'ish log levels listed.
Apart from that, this is just a re-implementation of what existed before this PR.
If we'd like to change or re-factor this bit, I suggest we open an issue for that. Thanks.

lib/logging.go
yermulnik: Thanks for catching this 🤝 Updated in a75ce8c

lib/logging.go
yermulnik: - The logging template falls apart logging levels struct properties that define logging levels rather than logging templates. I'd prefer to refrain from mixing different types of props in this struct.
- This is actually how this existed prior to this PR and this var is part of package-level vars along with regular logging template var. I'd prefer to leave this as is for now. And if there's a rationale in re-factoring this bit, it's worth of creating an issue for this (to help us track tasks like this). Thanks.

lib/logging.go
MatthewJohn: Oh sorry, when I was looking for uses, I thought it was in the same namespace

MatthewJohn (approved):

PR #609, opened 2025-07-31
feat: Support falling back to Terraform for pre-1.6 releases

mbainter:
This adds an option to automatically fall back to using Terraform when the version required from a terraform module is older than 1.6.0. This will aid teams working across multiple repositories or monorepos that have modules that haven't yet been upgraded.

I only implemented the fallback for versiontf handling, as I expect that is the place it will most be needed.

### DIFF
--- a/lib/param_parsing/versiontf.go
+++ b/lib/param_parsing/versiontf.go
@@ -42,7 +42,7 @@ func GetVersionFromVersionsTF(params Params) (Params, error) {
 	requiredVersions := module.RequiredCore
 
 	for key := range requiredVersions {
-		// Check if the version contraint is valid
+		// Check if the version constraint is valid
 		constraint, constraintErr := semver.NewConstraint(requiredVersions[key])
 		if constraintErr != nil {
 			logger.Errorf("Invalid version constraint found: %q", requiredVersions[key])
@@ -73,5 +73,5 @@ func isTerraformModule(params Params) bool {
 	if len(module.RequiredCore) == 0 {
 		logger.Debugf("No required version constraints defined by Terraform module at %q", params.ChDirPath)
 	}
-	return err == nil && len(module.RequiredCore) > 0
+	return len(module.RequiredCore) > 0
 }
--- a/lib/param_parsing/versiontf.go
+++ b/lib/param_parsing/versiontf.go
@@ -56,11 +56,38 @@ func GetVersionFromVersionsTF(params Params) (Params, error) {
 
 	version, err2 := lib.GetSemver(tfConstraint, params.MirrorURL)
 	if err2 != nil {
-		logger.Errorf("No version found matching %q", tfConstraint)
-		return params, err2
+		logger.Warnf("No %s version found matching %s", params.ProductEntity.GetName(), tfConstraint)
+		if params.FossFallback && strings.EqualFold(params.Product, "opentofu") {
+			tfparams := params // capture original settings
+			logger.Info("Testing for possible fallback to a matching FOSS Terraform version")
+			params = TofuFossFallback(params)
+
+			fossVersion, fossErr := lib.GetSemver(tfConstraint, params.MirrorURL)
+			if fossErr != nil {
+				logger.Errorf("No %s version found matching %s", params.Product, tfConstraint)
+				return params, fossErr
+			}
+			
+			fossLicensed, fossErr := lib.SemVerCheckFoss(fossVersion)
+			if fossErr != nil {
+				logger.Errorf("Terraform license check for %s failed: %v", fossVersion, err)
+				return params, fossErr
+			}
+
+			if !fossLicensed {
+				// If this is not a valid fallback then return the original error and settings
+				logger.Errorf("Matching Terraform version is not FOSS licensed: %s", fossVersion)
+				return tfparams, err2
+			}
+
+			version = fossVersion
+		} else {
+			return params, err2
+		}
 	}
+
 	params.Version = version
-	logger.Debugf("Using version from Terraform module at %q: %q", relPath, params.Version)
+	logger.Debugf("Using %q version from Terraform module at %q: %q", params.Product, relPath, params.Version)
 	return params, nil
 }
 
--- a/lib/param_parsing/parameters.go
+++ b/lib/param_parsing/parameters.go
@@ -32,6 +32,7 @@ type Params struct {
 	NoColor          bool
 	ProductEntity    lib.Product
 	Product          string
+	FossFallback     bool
 	ShowLatestFlag   bool
 	ShowLatestPre    string
 	ShowLatestStable string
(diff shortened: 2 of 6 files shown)


lib/param_parsing/versiontf.go
```diff
@@ -42,7 +42,7 @@ func GetVersionFromVersionsTF(params Params) (Params, error) {
 	requiredVersions := module.RequiredCore
 
 	for key := range requiredVersions {
-		// Check if the version contraint is valid
+		// Check if the version constraint is valid
```
MatthewJohn, line 45: This absolutely looks like a typo I could have made...

MatthewJohn (commented):
I'm not fully sure I understand the value of this (ro necessarily agree with the functionality)...
I'm assuming:
 * For a given number of teams and projects (where this number is relatively high) and presumably has a centralised CI/CD configuration (or templated)
 * Historically projects using Terraform <1.6
 * At the license change, it was decided to move to OpenTofu

In this centralised CI/CD pipeline, it is now (or will be) using tfswitch specifying that opentofu is the desired product.
Some teams have Terraform that still uses <1.6, so there is not compatible version of opentofu.
The `--foss-fallback` flag is passed to ensure that it always works.

However, in TofuFossFallback, you are updating the binary path to use the Terraform binary name.
This means that tfswitch, if specified to install opentofu with this flag may:
 * Install/update opentofu binary; or
 * Install/update terraform binary; or
 * Do nothing

Which feels very ambiguous. If this is meant to be a compatibility layer, I would almost expect that it installs Terraform *as* the `opentofu` binary... so it _just works_.

Having said this, this presumably means that you'd have a pipeline that had some logic that understands that the opentofu binary may not exist (_or_ if it does already exist, won't have been switched and instead terraform binary would have been installed/updated!).
If this is the case, can it not simly run tfswitch to install opentofu - if it fails, falls back to using tfswitch with `terraform` as the product?

Some other notes:
 * TofuFossFallback is quite confusing since opentofu is (in all versions) FOSS.
 * The same with the argument `--foss-fallback` - since this is being run against the `opentofu` product, to me, this doesn't make sense - you'd never need a FOSS fallback to OpenTofu. I would perhaps say `--terraform-fallback`, meaning it's a compatibiity layer for OpenTofu where versions don't exist. OR EVEN... When calling with the Terraform product and using passing an argument that says `--always-foss`, meaning it falls.. forward to opentofu (but I'm not sure about this idea at all).. just playing it the other way round in my mind.
 * I guess summing this up, it doesn't seem to me that this is really anything to do with FOSS at all - it's about _wanting_ to use OpenTofu, but it not having an available version <1.6, so falling back to Terraform.. it just so happens (well, not exactly by coincidence) that these versions are FOSS. Perhaps if it's just `--terraform-fallback` it would make more sense.

Maybe @warrensbox / @yermulnik

lib/param_parsing/versiontf.go
MatthewJohn: Nvm, it wasn't

yermulnik:
> Having said this, this presumably means that you'd have a pipeline that had some logic that understands that the opentofu binary may not exist (_or_ if it does already exist, won't have been switched and instead terraform binary would have been installed/updated!).
> If this is the case, can it not simly run tfswitch to install opentofu - if it fails, falls back to using tfswitch with `terraform` as the product?

I absolutely agree (including the ambiguity of parameter naming).
While I can relate the use case, I'm a way far from thinking this is something that `tfswitch` should be handling.

What I'd probably suggest (additionally to @MatthewJohn's comment) is to utilize any kind of automation like https://direnv.net/ (or similar https://direnv.net/#related-projects) and set `TF_PRODUCT` env var to `terraform` if the required version in the module is lower than 1.6.0 (or whatever is your preference).
Once the https://github.com/warrensbox/terraform-switcher/pull/536 is implemented, you'll also be able to utilize `tfswitch` to detect the required version of the module directory and make a decision on e.g. exporting `TF_PRODUCT` env var based on that.

@warrensbox Please let me know if you think otherwise and I'll re-open this PR.

@mbainter We do appreciate the contribution and hope to see more from you. Apologies that this one doesn't match our view 🤷🏻

mbainter:
This does actually have to do with FOSS, as the goal isn't to just switch to any version of Terraform if the resolved version isn't a valid OpenTofu version regardless of license. The dividing line for when it falls back is for versions of Terraform that were FOSS licensed.  The intent is for the fallback to be safe if you are in a situation where using BSL licensed Terraform would be a violation of the license for you.

This definitely can affect CICD, and yes, you could write your own logic for that. Even given that use case though,most of us use tfswitch and leverage it across environments to *avoid* recreating that logic everywhere. 

Here I'm way more concerned about the person working with the repository and trying to navigate versions. TFSwitch is a godsend for that, but the current implementation of product support is entirely all or nothing. If you're in transition it's terribly frustrating.

direnv sounds great, but if you have hundreds or even thousands of modules, do you really want to create a .envrc in each and every one, and direnv allow each and every directory if you want to work in it? And for those of us already using envrc, you have to "source_up" in every one in order to keep everything working,

On a more personal note, I'm a little surprised that when your review is "we're not sure we understand the use case, and don't like the names you chose" you opted to straight up close the pull-request before we could even discuss it.

yermulnik:
> TFSwitch is a godsend for that, but the current implementation of product support is entirely all or nothing. If you're in transition it's terribly frustrating.

I'm really sorry about the negative experience at your end. We talked about this internally and we found this feature to be not quite transparent. And additionally, based on your explanation of use case, the gist of this feature is transient and limited to use cases where teams migrate from TF to Tofu only (which means the use case is transient) and hence implementing this as a general feature looks a bit one-off'ish.

> direnv sounds great, but if you have hundreds or even thousands of modules, do you really want to create a .envrc in each and every one, and direnv allow each and every directory if you want to work in it? And for those of us already using envrc, you have to "source_up" in every one in order to keep everything working,

- I guess you can use `~/.config/direnv/direnvrc` (or `~/.direnvrc` 🤔), which is loaded between the `direnv`'s stdlib and any `.envrc`, to define your own sort of global logic (like «_if `*.tf` files exist in current dir and required version is less than 1.6.0, then export `TF_PRODUCT=terraform`, else export `TF_PRODUCT=opentofu`_»
  - (until the https://github.com/warrensbox/terraform-switcher/pull/536 is implemented to ease this bit of logic) I personally use at my current job a pretty simple self-written Go utility that reads version constraint (similarly as `tfswitch` does and which can also be extracted by https://github.com/hashicorp/terraform-config-inspect like `terraform-config-inspect --json . | jq -r .required_core[0]`) and returns either success or failure if my current TF version matches the requirement from version constraint (which can be used to decide on which Product to use in your use case)
- You can configure `direnv` to allow any dir hierarchy prefixes w/o a need for explicit approval: https://github.com/direnv/direnv/blob/master/man/direnv.toml.1.md#whitelist

> On a more personal note, I'm a little surprised that when your review is "we're not sure we understand the use case, and don't like the names you chose" you opted to straight up close the pull-request before we could even discuss it.

We're open for discussion and that was why we did not limit the conversation in this PR to collaborators only.
So if you're up to discuss this further, please go on.
As I mentioned, I do relate the use case of this PR, thought we didn't find this PR to be the right solution for it.

Issue #621, opened 2025-09-05
tfswitch pgp key file error from www.hashicorp.com


tommyreilly:
Seeing 

```
 % tfswitch -s 1.13.0
16:13:23.692 INFO Matched version: "1.13.1"  
16:13:23.814 INFO Downloading to "/Users/tommyreilly/.terraform.versions/pgp-key.txt"  
16:13:23.945 ERROR Error fetching public key file from https://www.hashicorp.com/.well-known/pgp-key.txt  
16:13:23.945 ERROR Could not download public PGP key file  
16:13:23.946 FATAL Error downloading: Unable to download from https://www.hashicorp.com/.well-known/pgp-key.txt
```

tommyreilly:
Is there some form of rate limiting happening?

```
% curl -I https://www.hashicorp.com/.well-known/pgp-key.txt
HTTP/2 429 
cache-control: private, no-store, max-age=0
content-type: text/html; charset=utf-8
server: Vercel
x-vercel-challenge-token: 2.1757104679.60.MTY3NDUxYzM3NDdlYmVlYmZiODkyN2JiN2IwZjUyZTM7Yjk1NGJkMzI7MjgwZDk5NTk2ZmUzZWY4ZDRkN2E5MTY5ZDUxYmUwNjc1YjQ1MGEwZDszO4yH6atbPuFwrRO23uw6OSRifDwmuSOQPZNoyiqSabxjn5el4AkgC1HeezcQtyk=.cc118cc964333dad5cf55645efff7931
x-vercel-id: iad1::1757104679-7Xxm9HRatCA4bIINtFmWAxST3iCIItxW
x-vercel-mitigated: challenge
date: Fri, 05 Sep 2025 20:37:59 GMT
```

yermulnik:
Please see linked issue. Seems like there's not much we can do about this (and it's not the first time Hashicorp breaks the thing: https://github.com/warrensbox/terraform-switcher/issues/540).

yermulnik:
The temp workaround is to download PGP-signature manually (e.g. via browser) and put it where `tfswitch` is looking for it: `~/.terraform.versions/terraform_72D7468F.asc`

yermulnik:
Resolved for me:
```shell
> curl -SsL https://hashicorp.com/.well-known/pgp-key.txt | head -c 100
-----BEGIN PGP PUBLIC KEY BLOCK-----

[...]

Issue #540, opened 2025-02-03
ERROR Could not download public PGP key file


Maple-Shade:
### Description
I am using terraform-switcher in my Gitlab CI/CD pipeline. Starting today (2/3/2025) I am getting the error shown below. My runner configuration hasn't changed and this was working previously, so I'm not sure what the issue is. I don't believe this is an internet access issue as I can still run other Gitlab jobs on my runner that require internet access.

I am successfully able to run the command `curl https://www.hashicorp.com/.well-known/pgp-key.txt` before using the `tfswitch` command so I believe this is an issue with terraform-switcher.

Error:
```
$ curl https://www.hashicorp.com/.well-known/pgp-key.txt
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0Redirecting...
100    15    0    15    0     0     94      0 --:--:-- --:--:-- --:--:--    95
$ tfswitch -g DEBUG $TERRAFORM_VERSION
23:01:52.242 DEBUG [versiontf.go:73,isTerraformModule] No required version constraints defined by Terraform module at "{removed for privacy}"  
23:01:52.242 DEBUG [install.go:2[25](https://gitlab.com/{removed for privacy}/-/jobs/9030533140#L25),InstallProductVersion] Install version 1.9.8. Dry run: false  
23:01:52.242 INFO [files.go:103,createDirIfNotExist] Creating "/root/.terraform.versions" directory  
23:01:52.242 DEBUG [list_versions.go:41,getTFList] Getting list of versions  
23:01:52.325 DEBUG [download.go:134,downloadPublicKey] Looking up public key file at "/root/.terraform.versions/terraform_72D7468F.asc"  
23:01:52.325 INFO [download.go:99,downloadFromURL] Downloading to "/root/.terraform.versions/pgp-key.txt"  
23:01:52.601 ERROR [download.go:140,downloadPublicKey] Error fetching public key file from https://www.hashicorp.com/.well-known/pgp-key.txt  
23:01:52.601 ERROR [download.go:[27](https://gitlab.com/{removed for privacy}/-/jobs/9030533140#L27),DownloadProductFromURL] Could not download public PGP key file.  
23:01:52.601 FATAL [install.go:90,install] Error downloading: Unable to download from https://www.hashicorp.com/.well-known/pgp-key.txt
```

[...]

PR #624, opened 2025-09-06
feat: Add KeyBase as fallback source of Hashicorp's public PGP-key

yermulnik:
- Add https://keybase.io/hashicorp/pgp_keys.asc as fallback source of Hashicorp's public PGP-key
- Allow multiple sources for Product's public PGP-key
- (unrelated) Preserve ordering of `LogLevels` in output messages to make clear the order of increasing level of verbosity (try to improve UX)

Relates to:
- #540
- #621

### DIFF
--- a/lib/download.go
+++ b/lib/download.go
@@ -136,18 +136,37 @@ func downloadFromURL(installLocation string, url string, wg *sync.WaitGroup) (st
 
 func downloadPublicKey(product Product, installLocation string, wg *sync.WaitGroup) (string, error) {
 	pubKeyFilePath := filepath.Join(installLocation, "/", product.GetId()+"_"+product.GetPublicKeyId()+pubKeySuffix)
-	logger.Debugf("Looking up public key file at %q", pubKeyFilePath)
+	logger.Debugf("Looking up public PGP-key file at %q", pubKeyFilePath)
 	publicKeyFileExists := FileExistsAndIsNotDir(pubKeyFilePath)
 	if !publicKeyFileExists {
-		// Public key does not exist. Let's grab it from hashicorp
-		pubKeyFile, errDl := downloadFromURL(installLocation, product.GetPublicKeyUrl(), wg)
-		if errDl != nil {
-			logger.Errorf("Error fetching public key file from %s", product.GetPublicKeyUrl())
-			return "", errDl
+		// Public PGP-key does not exist. Let's grab it
+		publicKeyURLs := product.GetPublicKeyURLs()
+		var pubKeyFile string
+		var errDl error
+		var errsDl []string
+		for idx, publicKeyURL := range publicKeyURLs {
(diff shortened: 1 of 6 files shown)


lib/download.go
```diff
@@ -136,18 +136,37 @@ func downloadFromURL(installLocation string, url string, wg *sync.WaitGroup) (st
...
+			logger.Debugf("Attempting to download public PGP-key from %q", publicKeyURL)
+			pubKeyFile, errDl = downloadFromURL(installLocation, publicKeyURL, wg)
+			if errDl != nil {
+				errsDl = append(errsDl, errDl.Error())
+				logger.Errorf("Failed to fetch public PGP-key from %q", publicKeyURL)
+
+				// Return all failures if all URLs have been tried so far
+				if idx+1 == len(publicKeyURLs) {
+					return "", errors.New(strings.Join(errsDl, "; "))
+				}
```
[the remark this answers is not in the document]
yermulnik: Logger messages above already have this context.

warrensbox (approved):
Looks good

lib/download.go
```diff
@@ -136,18 +136,37 @@ func downloadFromURL(installLocation string, url string, wg *sync.WaitGroup) (st
...
-		// Public key does not exist. Let's grab it from hashicorp
-		pubKeyFile, errDl := downloadFromURL(installLocation, product.GetPublicKeyUrl(), wg)
-		if errDl != nil {
-			logger.Errorf("Error fetching public key file from %s", product.GetPublicKeyUrl())
-			return "", errDl
+		// Public PGP-key does not exist. Let's grab it
+		publicKeyURLs := product.GetPublicKeyURLs()
+		var pubKeyFile string
+		var errDl error
+		var errsDl []string
```
MatthewJohn, line 146: I imagine this shouldl be a list of errors - I think you can do some clever things with them: to combine them: https://pkg.go.dev/errors

lib/download_test.go
```diff
@@ -366,7 +366,7 @@ func TestDownloadProductFromURL_invalid_public_key(t *testing.T) {
 			ExecutableName: "myprod",
 			ArchivePrefix:  "my_product_download_",
 			PublicKeyId:    downloadProductTestConfig.GpgFingerprint,
-			PublicKeyUrl:   mockServer.URL + "/testproduct/gpg-key.txt",
+			PublicKeyURLs:  []string{mockServer.URL + "/testproduct/gpg-key.txt"},
```
MatthewJohn, line 369: Could porbably do witgh a dedicated test to ensure that an invalid URL falls back to the secondary one

MatthewJohn (commented):
Looks good :) Sorry for the delay, am on holiday

lib/download.go
yermulnik: Nah, I wanted it to be a list of strings, as I join them later with `;` to keep all errors on single line in output (the `errors.Join` joins with newline and cannot use custom separator).

PR #640, opened 2025-10-13
feat: Check if version matches version requirement

yermulnik:
```bash
 -n, --match-version-requirement=value
                    Check if the requested version matches the requirement
                    mandated by the configuration (env var, module version
                    constraint, config files). Exit successfully if it does (or
                    if there's no requirement found), otherwise exit with a code
                    of `2` (code of `1` denotes a general error)
```

### DIFF
--- a/lib/param_parsing/match_version_requirement.go
+++ b/lib/param_parsing/match_version_requirement.go
@@ -14,7 +14,7 @@ func MatchVersionRequirement(parameters Params) (bool, error) {
 	// Sanity check MatchVersionRequirement parameter value
 	if !lib.IsValidVersionFormat(parameters.MatchVersionRequirement) {
 		lib.PrintInvalidTFVersion()
-		//nolint:staticcheck // We use capiltalized error messages
+		//nolint:staticcheck // We use capitalized error messages
 		return false, fmt.Errorf("Failed to validate version format: %q", parameters.MatchVersionRequirement)
 	}
 


www/docs/usage/commandline.md
```diff
@@ -138,3 +138,45 @@ running scripts in CI/CD pipeline or when piping output to other commands.
   (non-interactive session), you can use the `--force-color` (`-K`) flag.
 
 **NOTE**: `--no-color` and `--force-color` flags are mutually exclusive.
+
+## Check if a specific Product version matches a version requirement
+
+`tfswitch` can be used to match a specific Product version against a version
+mandated by the configuration (version provided on command line as argument to
+`tfswtich`, `TF_VERSION` environment variable, root module version constraint,
```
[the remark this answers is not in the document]
yermulnik: Fixed

lib/param_parsing/match_version_requirement.go
```diff
@@ -0,0 +1,42 @@
+//nolint:revive // FIXME: don't use an underscore in package name
+package param_parsing
+
+import "github.com/warrensbox/terraform-switcher/lib"
+
+// MatchVersionRequirement : Check if a given version meets a version requirement mandated by the configuration.
+// - Return `0` if the version matches the requirement, otherwise return `2`.
+// - Return `1` for any kind of general error (like version format parsing error).
+func MatchVersionRequirement(parameters Params) int {
```
MatthewJohn: It's probably more normal to return a bool (matches) and an error with the error.
I'm not returning numbers here makes sense (as I don't see more than 2 options, other than error).
But, if you do stick to returning numbers, you definitely need to return some sort of Enum - numbers (or strings) are just too magic and don't provide a logical way to validate calling behaoir (other than reading the comment)

lib/param_parsing/versiontf.go
```diff
@@ -12,7 +12,11 @@ import (
 	"github.com/warrensbox/terraform-switcher/lib"
 )
 
-func GetVersionFromVersionsTF(params Params) (Params, error) {
```
MatthewJohn, line 15: Was this not used before?

MatthewJohn:
Might be worth taking a look at sysexits (man sysexits) for list of common exit codes which may be applicable

lib/param_parsing/match_version_requirement.go
yermulnik: Given that both you and @warrensbox made a notice on this behavior, let me work on improving this bit. Thanks.

lib/param_parsing/versiontf.go
yermulnik: It was and it still is. I split a portion of this function off into its own function (consumed by `GetVersionFromVersionsTF()` func) and thus the `getConstraintFromVersionsTF()` function definition has moved to line 64 instead. So it is still there. Almost untouched in its essence.

lib/param_parsing/match_version_requirement.go
yermulnik: This is done now via e6a6346
Please give this PR another round of review.

lib/param_parsing/match_version_requirement.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+// MatchVersionRequirement : Check if a given version meets a version requirement mandated by the configuration.
+// - Return `true` if the version matches the requirement, otherwise return `false`.
+// - Return `error` for any kind of general error (like version format parsing error).
+func MatchVersionRequirement(parameters Params) (bool, error) {
+	// Sanity check MatchVersionRequirement parameter value
+	if !lib.IsValidVersionFormat(parameters.MatchVersionRequirement) {
+		lib.PrintInvalidTFVersion()
+		//nolint:staticcheck // We use capiltalized error messages
```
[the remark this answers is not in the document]
yermulnik: Fixed

Issue #508, opened 2024-09-27
Consider .tofu files


braun1928:
Starting with opentofu 1.8.0, it can read `<file>.tofu` as well as .tf files, preferring the former if the file names are the same. tfswitch should follow that when issuing `-t opentofu` to consider the `.tofu` files to define the version.

yermulnik:
`tfswitch` doesn't rely on file extension directly, but instead relies upon `tfconfig` library: https://github.com/hashicorp/terraform-config-inspect

While your feature request makes sense to me, do you by any chance know does OpenTofu provide the counterpart for `tfconfig` lib?

ps: obviously it would be great for the underlying library to add support for OpenTofu-specific file extension, though I'm a bit hesitant on whether Hashicorp would be really keen to implement that any time soon I guess 🤷🏻

braun1928:
I didn't see any fork of that repo; I found a commit in 1.8.0 tree "OpenTofu Specific Code Override: Add support to .tofu files" that does [exactly](https://github.com/opentofu/opentofu/commit/ab289fc07c4dbbb75eecb2b526e8e8abbf23c5fd#diff-25ade80935f9d3d9172f35b7e6ce6ccfeb5286ea6e2e8f468c73fe9192e034a3R246) what I described.

I doubt Hashi will do that change on their repo, though, and I haven't seen if that repo being forked to add this change (or similar) is on opentofu's roadmap as well.

yermulnik:
> to add this change (or similar) is on opentofu's roadmap as well.

Well, given OpenTofu has started to diverge from Terraform, I guess we should then hope for them to start providing common public libs that include compatibility for diverged features to allow easier updates/changes to existing tooling and libraries 🤞🏻 

Let's re-open this issue when that starts to become reality (as I don't think it's good idea to re-factor `tfswitch` to re-implement what has to be provided by the parent).
Thanks for raising this for our awareness.

PR #696, opened 2026-02-04
feat(versiontf): handle tofu files

Intuinewin:
Hello,

As stated in #508, `.tofu` files should be considered to define the version tfswitch will use.

It seems that OpenTofu is not going to reimplement the lib terraform-config-inspect. For their registry they used directly a custom opentofu binary instead of providing a lib (cf https://github.com/opentofu/registry-ui/blob/ac212136d86e13ded7d9796f2f7df39f5b13b152/documentation/documentation-sources.md#the-challenge-with-terraform-config-inspect).

My solution here is to use the hcl parser already used for terragrunt.

> as I don't think it's good idea to re-factor tfswitch to re-implement what has to be provided by the parent

I saw your response but as the parser is already used in this project and the code is not *that* complicated, could this PR be merged ?

Let me know what you think.
Thanks

Related https://github.com/warrensbox/terraform-switcher/issues/508

### DIFF
--- a/lib/param_parsing/versiontf.go
+++ b/lib/param_parsing/versiontf.go
@@ -61,7 +61,7 @@ func getVersionConstraintsFromHCLFile(fileName string, hclFile *hcl.File) ([]str
 					continue
 				}
 				if !val.IsKnown() || !val.Type().Equals(cty.String) {
-					logger.Debugf("Skipping not known or non-string value of %q at %q: %q", requiredVersionAttrName, fileName, val)
+					logger.Debugf("Skipping not known or non-string value of %q in %q: %#v", requiredVersionAttrName, fileName, val)
 					continue
 				}
 				versionStr := val.AsString()
@@ -71,7 +71,7 @@ func getVersionConstraintsFromHCLFile(fileName string, hclFile *hcl.File) ([]str
 				}
 				constraint, constraintErr := semver.NewConstraint(versionStr)
 				if constraintErr != nil {
-					logger.Errorf("Invalid version constraint found: %q", versionStr)
+					logger.Errorf("Invalid version constraint found in %q: %q", fileName, versionStr)
 					return nil, constraintErr
 				}
 				logger.Debugf("Found %q %q in %q", requiredVersionAttrName, constraint.String(), fileName)
@@ -88,7 +88,6 @@ func getVersionConstraintsFromFiles(filesPath []string) (string, error) {
 	for _, filePath := range filesPath {
 		_, diagnostics := parser.ParseHCLFile(filePath)
 		if diagnostics.HasErrors() {
-			logger.Errorf("Unable to parse HCL file %q: %v", filePath, diagnostics.Error())
 			return "", fmt.Errorf("Could not parse HCL file %q: %v", filePath, diagnostics.Error())
 		}
 	}
@@ -115,23 +114,30 @@ func getConstraintFromVersionsTF(params Params) (Params, error) {
 
 	extensionsPerProduct := lib.GetProductById(params.Product).GetFileExtensions()
 	var hclFiles []string
+	var fileGlobs []string
 	for _, ext := range extensionsPerProduct {
-		files, globErr := filepath.Glob(filepath.Join(relPath, fmt.Sprintf("*.%s", ext)))
+		globPattern := fmt.Sprintf("*.%s", ext)
+		fileGlobs = append(fileGlobs, globPattern)
+		files, globErr := filepath.Glob(filepath.Join(relPath, globPattern))
 		if globErr != nil {
-			return params, fmt.Errorf("Could not list %s files in %q: %v", ext, relPath, globErr)
+			return params, fmt.Errorf("Could not list %s files in %q: %v", globPattern, relPath, globErr)
 		}
 		hclFiles = append(hclFiles, files...)
 	}
 
 	if len(hclFiles) == 0 {
-		logger.Debugf("No %s files found in %q", strings.Join(extensionsPerProduct, ", "), relPath)
+		logger.Debugf("No %s files found in %q", strings.Join(fileGlobs, ", "), relPath)
 		return params, nil
 	}
 
 	versionRequirements, err := getVersionConstraintsFromFiles(hclFiles)
 	if err != nil {
-		logger.Errorf("Error parsing %s files in %q: %v", paramTypeVersionTF, relPath, err)
 		return params, err
+	}
+
+	if versionRequirements == "" {
+		logger.Debugf("No version requirements found in %s files in %q", strings.Join(fileGlobs, ", "), relPath)
+		return params, nil
 	}
 
 	params.VersionRequirement = versionRequirements


lib/param_parsing/versiontf.go
```diff
@@ -7,21 +7,83 @@ package param_parsing
...
+	// Define the schema for the terraform block
+	terraformBlockSchema := &hcl.BodySchema{
+		Blocks: []hcl.BlockHeaderSchema{
+			{Type: terraformBlockType},
+		},
+	}
+
+	content, _, diags := hclFile.Body.PartialContent(terraformBlockSchema)
+	if diags.HasErrors() {
+		logger.Debugf("No terraform blocks found in %q", filePath)
```
yermulnik:
```suggestion
		logger.Debugf("No %q blocks found in %q", terraformBlockType, filePath)
```

lib/param_parsing/versiontf.go
```diff
@@ -7,21 +7,83 @@ package param_parsing
...
+		if block.Type == terraformBlockType {
+			// Extract required_version from the terraform block
+			terraformAttributesSchema := &hcl.BodySchema{
+				Attributes: []hcl.AttributeSchema{
+					{Name: requiredVersionAttrName},
+				},
+			}
+			blockContent, _, attrDiags := block.Body.PartialContent(terraformAttributesSchema)
+			if attrDiags.HasErrors() {
+				logger.Debugf("Error getting attributes from terraform block in %q: %s", filePath, attrDiags.Error())
```
yermulnik:
```suggestion
				logger.Debugf("Error getting attributes from %q block in %q: %s", terraformBlockType, filePath, attrDiags.Error())
```

lib/param_parsing/versiontf.go
```diff
@@ -7,21 +7,83 @@ package param_parsing
...
+				continue
+			}
+
+			if attr, exists := blockContent.Attributes[requiredVersionAttrName]; exists {
+				val, valDiags := attr.Expr.Value(nil)
+				if valDiags.HasErrors() {
+					logger.Debugf("Error evaluating required_version in %q: %s", filePath, valDiags.Error())
```
yermulnik:
```suggestion
					logger.Debugf("Error evaluating %q in %q: %s", requiredVersionAttrName, filePath, valDiags.Error())
```

lib/param_parsing/versiontf.go
```diff
@@ -7,21 +7,83 @@ package param_parsing
...
+					continue
+				}
+				if !val.IsKnown() || !val.Type().Equals(cty.String) {
+					continue
```
yermulnik, line 65:
```suggestion
					logger.Debugf("Skipping not known or non-string value of %q at %q: %q", requiredVersionAttrName, filePath, val)
					continue
```

lib/param_parsing/versiontf.go
```diff
@@ -7,21 +7,83 @@ package param_parsing
...
+				}
+				versionStr := val.AsString()
+				if versionStr != "" {
+					logger.Debugf("Found required_version %q in %q", versionStr, filePath)
```
yermulnik:
```suggestion
					logger.Debugf("Found %q in %q: %q", requiredVersionAttrName, filePath, versionStr)
```

lib/param_parsing/versiontf.go
```diff
@@ -7,21 +7,83 @@ package param_parsing
...
-		return params, nil
+// getRequiredVersionsFromFile parses a single .tf or .tofu file and extracts all required_version constraints
+func getRequiredVersionsFromFile(filePath string) ([]string, error) {
+	var versions []string
+
+	parser := hclparse.NewParser()
+	hclFile, diagnostics := parser.ParseHCLFile(filePath)
+	if diagnostics.HasErrors() {
+		logger.Errorf("Unable to parse HCL file %q: %s", filePath, diagnostics.Error())
+		return nil, fmt.Errorf("Could not parse HCL file %q: %s", filePath, diagnostics.Error())
```
yermulnik: Is there a reason to print errors as `%s` rather than as `%v`?
We mostly use `%v` for errors.
If there's no particular reason for `%s`, please update the PR to use `%v` for errors for consistency.
Thanks.

lib/param_parsing/versiontf.go
```diff
@@ -30,22 +92,48 @@ func getConstraintFromVersionsTF(params Params) (Params, error) {
...
+				return params, constraintErr
+			}
+			tfConstraints = append(tfConstraints, constraint.String())
 		}
-		// It's valid. Add to list
-		tfConstraints = append(tfConstraints, constraint.String())
+	}
+
+	if len(tfConstraints) == 0 {
+		logger.Debugf("No terraform version constraint found in %s configuration", paramTypeVersionTF)
```
yermulnik:
```suggestion
		logger.Debugf("No version constraint found in %s configuration", paramTypeVersionTF)
```

lib/param_parsing/versiontf_test.go
```diff
@@ -33,6 +33,30 @@ func TestGetVersionFromVersionsTF_matches_version(t *testing.T) {
...
+func TestGetVersionFromVersionsTF_matches_version_opentofu(t *testing.T) {
+	logger = lib.InitLogger("DEBUG")
+	var params Params
+	var getVerErr error
+	params = initParams(params)
+	params.ChDirPath = "../../test-data/integration-tests/test_versiontf_opentofu"
+	params.MirrorURL = lib.GetProductById("terraform").GetDefaultMirrorUrl()
+	params, getVerErr = GetVersionFromVersionsTF(params)
+	if getVerErr != nil {
+		t.Errorf("Error getting version from Terraform module: %v", getVerErr)
```
yermulnik:
```suggestion
		t.Errorf("Error getting version from OpenTofu module: %v", getVerErr)
```

lib/param_parsing/versiontf_test.go
```diff
@@ -33,6 +33,30 @@ func TestGetVersionFromVersionsTF_matches_version(t *testing.T) {
 	}
 }
 
+func TestGetVersionFromVersionsTF_matches_version_opentofu(t *testing.T) {
+	logger = lib.InitLogger("DEBUG")
+	var params Params
+	var getVerErr error
+	params = initParams(params)
+	params.ChDirPath = "../../test-data/integration-tests/test_versiontf_opentofu"
+	params.MirrorURL = lib.GetProductById("terraform").GetDefaultMirrorUrl()
```
yermulnik: I remember that OpenTofu support isn't complete, though should here `opentofu` be used instead of `terraform` as Product id? 🤔

lib/param_parsing/versiontf.go
```diff
@@ -59,6 +147,10 @@ func GetVersionFromVersionsTF(params Params) (Params, error) {
 		return params, err
 	}
 
+	if params.VersionRequirement == "" {
+		return params, nil
+	}
```
yermulnik, line 157: I'm not really sure this if-conditional is necessary here.
Could you please elaborate on the reason for it to be added? Thanks.

lib/param_parsing/versiontf_test.go
Intuinewin: Yep nice catch

lib/param_parsing/versiontf.go
Intuinewin: Before this PR, there was a condition around `GetVersionFromVersionsTF` to prevent `GetVersionFromVersionsTF` from exiting without version constraint.

Now to avoid parsing the code twice, `GetVersionFromVersionsTF` is always called and therefore need to handle the case where the parsing is successful but no `required_version` is found. In this case the params object is returned "as is" without trying to parse the version. it's very similar to terragrunt code.

I will add a comment

yermulnik: Ah, I see, missed that bit, thanks.

lib/param_parsing/versiontf.go
```diff
@@ -30,22 +93,48 @@ func getConstraintFromVersionsTF(params Params) (Params, error) {
 	}
 
 	logger.Infof("Reading version constraint from %s at %q", paramTypeVersionTF, relPath)
-	module, _ := tfconfig.LoadModule(relPath) // nolint:errcheck // covered by conditional below
-	if module.Diagnostics.HasErrors() {
-		return params, fmt.Errorf("Could not load %s at %q", paramTypeVersionTF, relPath)
+
+	// Find all .tf and .tofu files in the directory
+	var hclFiles []string
+	for _, ext := range []string{"*.tf", "*.tofu"} {
```
MatthewJohn: I imagine that as opentofu find it's feet, they're going to make more opentofu-specificites.
I wonder if the list of file extensions can be passed into getConstraintFromVersionsTF (only to keep this method quite self-contianed) and this can be determined based on a per-product configuration?

MatthewJohn: The reason is:
If tfswitch is using `terraform` product and tfswitch interprets .tofu files, it would seem _very_ odd behavoir for tfswitch to switch terraform versions based on this.

lib/param_parsing/versiontf.go
```diff
@@ -7,21 +7,84 @@ package param_parsing
...
+	paramTypeVersionTF      = "Terraform/OpenTofu module"
+	terraformBlockType      = "terraform"
+	requiredVersionAttrName = "required_version"
+)
 
-func getConstraintFromVersionsTF(params Params) (Params, error) {
-	if !isTerraformModule(params) {
-		return params, nil
+// getRequiredVersionsFromFile parses a single .tf or .tofu file and extracts all required_version constraints
+func getRequiredVersionsFromFile(filePath string) ([]string, error) {
```
MatthewJohn: To make this a little more similar, perhaps, to the original version (and I'm wondering if it's more efficient), would it make sense to:
 * pass a list of fiels to this function
 * iterate over files and call parser.ParseHCLFile
 * Once all files are parsed, perform the logic looking for the terraform block, etc.

MatthewJohn: If anything, it would make the functionality more similar to the previous version, at least in a failure case:
If there is one terraform file that is fails to be parsed, then parsing of Terraform files won't be used to obtain a version

lib/param_parsing/versiontf.go
Intuinewin: Indeed this makes more sense
https://github.com/warrensbox/terraform-switcher/pull/696/commits/83272f8f0b923ee2a787da89920db545cfca9b55

lib/param_parsing/versiontf.go
Intuinewin: I do like the fact that the function process a single file, as each file is independant. However I undestand your point about error handling, I can update the for loop to raise the error if the parsing fails and no version constraints is used
https://github.com/warrensbox/terraform-switcher/pull/696/changes/64087c79ddf53ec715d60573be6aee5cc716ac48

MatthewJohn: I do agree, I was hesistent to write it! But processing the files as a whole is how it used to work (this was my biggest... fear?) and the performance was mostly an addition - seeing the actual renderer instance is just a struct, re-creating it won't be too expensive)
@yermulnik , what are your thoughts on it?

lib/param_parsing/versiontf.go
Intuinewin: This breaks integrations tests which only supports the terraform product.
If you could take a look plz or I can skip this for my new test
cf https://github.com/Intuinewin/terraform-switcher/actions/runs/21817924470/job/62943689320

lib/param_parsing/versiontf.go
yermulnik: @MatthewJohn I do agree with you. We'd better keep functionality similar to what exists at the moment.
Re the parsing errors, I guess it makes sense to fail `tfswitch` if it fails to parse TF/Tofu files as it's what TF/Tofu would do I guess.

lib/param_parsing/versiontf.go
yermulnik: @Intuinewin Have you identified the cause of the failure?
By the look of it, it's failing on the test file you're adding in this PR:
```
08:36:58.853 INFO Reading version constraint from Terraform/OpenTofu module at "test-data/integration-tests/test_versiontf_opentofu"  
08:36:58.881 FATAL Interactive prompt is not meant for non-interactive terminal (please use command line flags in such case)  
```

Intuinewin: During integrations tests tfswitch is called without product so it defaults to terraform and my test fails
https://github.com/warrensbox/terraform-switcher/blob/master/.github/workflows/build.yml#L82
Not sure how you want to fix that, maybe I should just remove this test from integrations tests

lib/param_parsing/versiontf.go
Intuinewin: Let me know if that's what you had in mind: https://github.com/warrensbox/terraform-switcher/pull/696/changes/b63f7cdf2d3db47363eb4c0967472c9f0f37e3a9

lib/param_parsing/versiontf.go
yermulnik: Let me know if you need a hand on updating that GH action to switch Product based on dir name and pass the value to `tfswitch`.
Or I can push it to https://github.com/warrensbox/terraform-switcher/pull/702 and you'll rebase on it once it's approved and merged.
Thanks.

Intuinewin: > Or I can push it to https://github.com/warrensbox/terraform-switcher/pull/702 and you'll rebase on it once it's approved and merged.

That would be great. Thank you

yermulnik: Here you go: 2388ac6
Let's summon @MatthewJohn and @warrensbox to have [that](https://github.com/warrensbox/terraform-switcher/pull/696) PR approved 🙏🏻

yermulnik: Or just cherry-pick it (maybe the whole PR as it improves output).
Let me rebase it to single commit.

yermulnik: 0781c6c

Intuinewin: Rebased, thanks

yermulnik (approved):
LGTM. Appreciate the contribution 🤝 
Will leave some space for @warrensbox and @MatthewJohn to add their verdict (if any).

MatthewJohn (approved):

yermulnik:
Hey @Intuinewin,
Apologies for delay and appreciate your contribution: https://github.com/warrensbox/terraform-switcher/releases/tag/v1.14.0

PR #739, opened 2026-04-17
Replace version body parsing with JSON version

### LINKED ISSUE #738: OpenTofu versions contains incorrect cloudflare script version
At some point, opentofu (https://get.opentofu.org/tofu/) added cloudflare analytics to their downloads, the page now contains:
```
<!-- Cloudflare Pages Analytics --><script defer src='[https://static.cloudflareinsights.com/beacon.min.js](view-source:https://static.cloudflareinsights.com/beacon.min.js)' data-cf-beacon='{"token": "d197cf7327bb4345a9878620e02ac825"}'></script><!-- Cloudflare Pages Analytics --><script defer src="[https://static.cloudflareinsights.com/beacon.min.js/v8c78df7c7c0f484497ecbca7046644da1771523124516](view-source:https://static.cloudflareinsights.com/beacon.min.js/v8c78df7c7c0f484497ecbca7046644da1771523124516)" integrity="sha512-8DS7rgIrAmghBFwoOTujcf6D9rXvH8xm8JQ1Ja01h9QX8EzXldiszufYa4IFfKdLUKTTrnSFXLDkUEOTrZQ8Qg==" data-cf-beacon='{"version":"2024.11.0","token":"c74d0801821d4fd3900819b3f5c62c60","server_timing":{"name":{"cfCacheStatus":true,"cfEdge":true,"cfExtPri":true,"cfL4":true,"cfOrigin":true,"cfSpeedBrain":true},"location_startswith":null}}' crossorigin="anonymous"></script>
```

The `2024.11.0` is being caught in version enumeration

### LINKED ISSUE #741: feat: Switch TF/Tofu version parsing from HTML to JSON
To get the list of available TF/Tofu versions, currently the HTML endpoints are parsed (see `DefaultMirror` at https://github.com/warrensbox/terraform-switcher/blob/master/lib/product.go#L169-L197), which implies burden on tracking changes to HTML representation of the versions list (as it is meant for humans rather than for automation). Moreover we allow custom mirrors that might have their very own representation format (e.g. #311).

Switch from parsing HTML to querying JSON:
* Terraform: https://releases.hashicorp.com/terraform/index.json
* OpenTofu: https://get.opentofu.org/tofu/api.json

⚠️ **Caveat**: custom mirrors like Artifactory or similar might not have a JSON endpoint exposed (which we might work around by keeping HTML parser logic for such cases though 🤔)

None as proven by #738 

**Additional context**
#311 (#313), #738 (#739), probably more.

MatthewJohn:
feat: Pull version from Terraform/Opentofu JSON endpoints
    
Support backwards compatibility by falling back to body handling - this is important for any mirrors.
Add JSON entity handling per-product.
Pass product into dependent methods, such as getTfList, getTFLatest.
Add test handling for JSON bodies, with configurable mock server to ensure we are pulling the correct body, but also handling the body-fallback
    
This avoids matching bare semantic versions, which are now found for cloudflare analytics in the opentofu page (see the updated test in previous commit).
This works for both Terraform and OpenTofu, but this remains brittle.
The test data has been updated with the latest body from OpenTofu.


MatthewJohn:
At some point opentofu/terraform will release a change that that means it is not easy to parse both pages with a single regex and we'll probably need product-specific methods to extract the versions.
I suspect it _might_ (or _could_) be worth doing this sooner AND updating the parsing to be more.. specific. E.g. matching very deliberate `/terraform/<version>` or `opentofu_<version>`.
However, aside from it being a bit of refactoring (passing _something_ from the product down the methods.. most likely breaking some public function signatures), it would _also_ end up being more brittle - the more specific we are away from just detecting _any_ semantic version, we start having more and more of a reliance on the precise format of the page.  But, at the same time, we should weigh the risks of, say:
 * opentofu going from labelling as opentofu_1.1.1 to opentofu-1.1.1, which would break the proposed method because we're _more_ specific; vs
 * something else being added to the page (either terraform or opentofu), similar to this cloudflare analytics, which also breaks it because we're not specific enough

MatthewJohn:
@copilot Any idea why the goleak is failing with this configuration?

lib/utils.go
```diff
@@ -89,3 +90,22 @@ func RemoveDuplicateStrings(slice []string) []string {
...
+	}
+
+	stdinFd, err1 := fdToInt(os.Stdin.Fd())
+	stdoutFd, err2 := fdToInt(os.Stdout.Fd())
+
+	if err1 != nil || err2 != nil {
+		return false
+	}
+
+	return !term.IsTerminal(stdinFd) || !term.IsTerminal(stdoutFd)
```
yermulnik, line 110: Why the single line `!term.IsTerminal(int(os.Stdin.Fd())) || !term.IsTerminal(int(os.Stdout.Fd()))` would not be sufficient here? Why do we need all the other logic? I'd expect the Golang `term` lib to already handle odd cases... 😕

lib/logging.go
```diff
@@ -57,7 +56,7 @@ func isColorLogging() bool {
 		return false
 	} else if color.SupportColor() {
 		if os.Getenv("FORCE_COLOR") == "" {
-			return term.IsTerminal(int(os.Stdout.Fd()))
+			return isTerminal()
```
yermulnik, line 59: I guess the code didn't check both stdin and stdout on purpose as the colour output requires stdout rather than both stdin and stdout. There can even exist use case where stdin might not be connected to a terminal, while stdout is connected.
Could you please split this change into a separate PR if you do think (and tested) this change is something we need to implement. Thanks.

lib/list_versions.go
```diff
@@ -32,9 +32,9 @@ func getVersionsFromBody(body string, preRelease bool, tfVersionList *tfVersionL
 	var semver string
 	// Without the ending '"' pre-release folders would be tried and break.
 	if preRelease {
-		semver = `\/?` + regexSemVer.Full.String() + `/?"`
+		semver = `\/` + regexSemVer.Full.String() + `/"`
 	} else if !preRelease {
-		semver = `\/?` + regexSemVer.Patch.String() + `\/?"`
+		semver = `\/` + regexSemVer.Patch.String() + `\/"`
```
yermulnik: Ain't this change going to break what was reported back in https://github.com/warrensbox/terraform-switcher/issues/311 and what was fixed via https://github.com/warrensbox/terraform-switcher/pull/313? 🤔 

The `\/?` is also used as regex boundary in `getTFLatest()` and `getTFLatestImplicit()`

yermulnik: I also seem to think that we'd better off looking up something more specific to the line(s) with tf/tofu versions instead of juggling slashes (which would also still support what #311 was about). E.g.:
```go
semver = `<a href="(?:\S+\/)?` + …
```
The `(?:\S+\/)?` is non-capturing group of zero or one "_non-whitespace chars with forward slash at the end_"

yermulnik (dismissed):

yermulnik:
As a more general note, we might want to look into switching from parsing HTML to querying JSON:
* Terraform: https://releases.hashicorp.com/terraform/index.json
* OpenTofu: https://get.opentofu.org/tofu/api.json

Although this still imposes caveat around custom mirrors like Artifactory or similar... which we might work around by keeping HTML parser logic for such cases though 🤔 

I created #741 for HTML-to-JSON version querying switch.

MatthewJohn:
> As a more general note, we might want to look into switching from parsing HTML to querying JSON:
> 
>     * Terraform: https://releases.hashicorp.com/terraform/index.json
> 
>     * OpenTofu: https://get.opentofu.org/tofu/api.json
> 
> 
> Although this still imposes caveat around custom mirrors like Artifactory or similar... which we might work around by keeping HTML parser logic for such cases though 🤔
> 
> I created #741 for HTML-to-JSON version querying switch.

Nice - I think this was mentioned originally when adding OpenTofu, but I thought one of them didn't provide a JSON file. This is great

lib/list_versions.go
MatthewJohn: Ah, I hadn't appreciated local mirrors :(

MatthewJohn: I now fear that we don't know the pattern that people may have in their local mirrors.

lib/product.go
```diff
@@ -78,8 +89,23 @@ func (p TerraformProduct) GetArchivePrefix() string {
...
-	mirrorURL = strings.TrimRight(mirrorURL, "/")
-	return fmt.Sprintf("%s/%s", mirrorURL, version)
+	downloadUrl := p.DefaultDownloadMirror
+
+	// If the actual mirror is not the default, use this mirror for downloading
+	if mirrorURL != p.DefaultMirror {
+		downloadUrl = mirrorURL
+	}
+	downloadUrl = strings.TrimRight(downloadUrl, "/")
+	return fmt.Sprintf("%s/%s", downloadUrl, version)
```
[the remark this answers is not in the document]
MatthewJohn: Originally we were taking the user-supplied mirrorURL....
This is just saying, if they haven't supplied one, then we'll switch to the default download URL, since the mirror points to the JSON endpoint

lib/list_versions.go
```diff
@@ -69,71 +115,37 @@ func getTFList(mirrorURL string, preRelease bool) ([]string, error) {
...
 	}
-	return "", nil
+	return semv, nil
 }
 
 // getTFURLBody : Get list of versions from the mirror URL
 func getTFURLBody(mirrorURL string) (string, error) {
 	hasSlash := strings.HasSuffix(mirrorURL, "/")
-	if !hasSlash {
+	isJson := strings.HasSuffix(mirrorURL, ".json")
```
MatthewJohn: I'll say - I don't like this - but I'm honestly quite concerned about removing the added trailing slash, as it could break custom mirrors that are now relying on this

lib/list_versions.go
MatthewJohn: This is now reverted

lib/logging.go
MatthewJohn: I have recverted all of this :)

lib/utils.go
MatthewJohn: As I say, this was thrown as an issue by superlinter :/
I'll raise another PR to fix it indepedently and merge the fix in

lib/product.go
```diff
@@ -78,8 +89,29 @@ func (p TerraformProduct) GetArchivePrefix() string {
...
+	if p.DefaultDownloadMirror != "" {
+		downloadUrl = p.DefaultDownloadMirror
+	}
+
+	// If the actual mirror is not the default, use this mirror for downloading
+	if mirrorURL != p.DefaultMirror {
+		downloadUrl = mirrorURL
+	}
+	downloadUrl = strings.TrimRight(downloadUrl, "/")
+	return fmt.Sprintf("%s/%s", downloadUrl, version)
```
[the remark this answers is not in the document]
MatthewJohn: @copilot 
Please consider the original behavour...
We used mirror URL, which was: a root index file that we scraped and obtained versions and then re-used to download.
Users could override this mirror URL (which would be used for both).
We've now changed the default to a JSON endpoint and are retaining the old functionality (if a user passed a download mirror URL that was the index file AND the base of the downloads)

lib/param_parsing/parameters.go
```diff
@@ -159,6 +186,8 @@ func populateParams(params Params) Params {
 		// First pass to obtain environment variables to override product
 		params = GetParamsFromEnvironment(params)
```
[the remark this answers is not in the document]
MatthewJohn: @copilot 
Er, no?
The order of this function is:
 * Parse CLI options
 * Get options from TOML file
 * Get env variables
 * Setup Product Param (new function)
 * THEN:
 * GetParamsFromTfSwitch (cannot set product)
 * GetParamsFromTerraformVersion (cannot set product)
 * GetVersionFromVersionsTF (cannot set product - but, critically, requires product to be set)
 * GetVersionFromTerragrunt (cannot set product - but, critically, requires product to be set)
 * GetParamsFromEnvironment (will not override product _again_ since it was the last function to be called that sets up product-related params)

lib/param_parsing/parameters.go
```diff
@@ -71,6 +71,33 @@ func GetParameters() Params {
...
+		if params.CustomBinaryPath == "" {
+			if runtime.GOOS == "windows" {
+				params.CustomBinaryPath = filepath.Join(lib.GetHomeDirectory(), "bin", lib.ConvertExecutableExt(product.GetExecutableName()))
+			} else {
+				params.CustomBinaryPath = filepath.Join("/usr/local/bin", product.GetExecutableName())
+			}
+		}
+		params.ProductEntity = product
+	}
+}
```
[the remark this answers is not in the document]
MatthewJohn: @copilot Can you create a Github  issue for this pre-existing issue, please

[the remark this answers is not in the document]
MatthewJohn: Done #742

lib/param_parsing/parameters.go
```diff
@@ -159,6 +186,8 @@ func populateParams(params Params) Params {
 		// First pass to obtain environment variables to override product
 		params = GetParamsFromEnvironment(params)
 
+		setupProductParam(&params)
```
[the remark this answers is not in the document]
MatthewJohn: We've already raised #742 for this

lib/list_versions.go
```diff
@@ -28,6 +31,46 @@ var regexSemVer = struct {
...
+		}
+	}
+
+	// Sort versions and store in tflist
+	sort.Sort(version.Collection(parsedVersions))
+	for itx := range parsedVersions {
+		tfVersionList.tflist = append(tfVersionList.tflist, parsedVersions[len(parsedVersions)-(itx+1)].Original())
+	}
+
+	return nil
```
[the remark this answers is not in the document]
MatthewJohn: > getVersionsFromJSON can return nil even when it parsed zero versions (e.g., JSON unmarshals successfully but the expected versions field is missing/empty).

I mean, if it's valid JSON, but doesn't contain versions, then it won't be parsable via body..
If it parses JSON correctly, then we only treat it as JSON. If it's invalid, then we test against against body parsing.

lib/list_versions_test.go
```diff
@@ -40,18 +42,23 @@ const (
...
 )
 
 // TestGetTFList : Get list from hashicorp
 func TestGetTFList(t *testing.T) {
-	list, err := getTFList(hashiURL, true)
+	product := GetProductById("terraform")
+	list, err := getTFList(product, hashiURL, true)
 	if err != nil {
```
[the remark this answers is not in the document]
MatthewJohn: Fixed in b60247a

lib/list_versions.go
MatthewJohn: @yermulnik @warrensbox

lib/param_parsing/versiontf_test.go
```diff
@@ -82,7 +87,9 @@ func TestGetVersionFromVersionsTF_erroneous_file(t *testing.T) {
 	var params Params
 	params = initParams(params)
 	params.ChDirPath = "../../test-data/skip-integration-tests/test_versiontf_error"
-	params.MirrorURL = lib.GetProductById("terraform").GetDefaultMirrorUrl()
+	params.Product = "opentofu"
```
yermulnik: What's the reason to switch this test from TF to Tofu? 🤔

lib/param_parsing/versiontf_test.go
```diff
@@ -99,7 +106,9 @@ func TestGetVersionFromVersionsTF_non_existent_constraint(t *testing.T) {
 	var params Params
 	params = initParams(params)
 	params.ChDirPath = "../../test-data/skip-integration-tests/test_versiontf_non_existent"
-	params.MirrorURL = lib.GetProductById("terraform").GetDefaultMirrorUrl()
+	params.Product = "opentofu"
```
yermulnik: What's the reason to switch this test from TF to Tofu? 🤔

lib/list_versions.go
```diff
@@ -28,6 +31,46 @@ var regexSemVer = struct {
...
+	}
+	semver = "^" + semver + "$"
+	r, err := regexp.Compile(semver)
+	if err != nil {
+		logger.Fatalf("Error compiling %q regex: %v", semver, err)
+		return err
+	}
+
+	parsedVersions := make([]*version.Version, 0, len(versionList))
+	for _, versionItx := range versionList {
```
yermulnik: What `Itx` stands for in this context? Can it be made a bit more obvious, like `versionItem` or similar please?

Also instead of discarding index from `range` keyword, might it be more natural in modern Go to use `range slices.Values()` instead (`for versionItx = range slices.Values(versionList) {`)? (ref: https://pkg.go.dev/slices#Values)

lib/list_versions.go
```diff
@@ -28,6 +31,46 @@ var regexSemVer = struct {
...
+		// Sanity check that it matches our regex
+		// Could be removed in future, perhaps, if we switch to Hashicorp library entirely.
+		if r.MatchString(versionItx) {
+			parsedVersion, err := version.NewSemver(versionItx)
+			if err != nil {
+				continue
```
yermulnik, line 60: I'd suggest to add debug-level message here to ease debugging of discarded versions if any.

lib/list_versions.go
```diff
@@ -28,6 +31,46 @@ var regexSemVer = struct {
...
+			}
+			parsedVersions = append(parsedVersions, parsedVersion)
+		}
+	}
+
+	// Sort versions and store in tflist
+	sort.Sort(version.Collection(parsedVersions))
+	for itx := range parsedVersions {
```
yermulnik: The `itx` name for an iterator is definitely something not really obvious to me =(

lib/list_versions.go
```diff
@@ -52,15 +95,18 @@ func getVersionsFromBody(body string, preRelease bool, tfVersionList *tfVersionL
...
+	body, err := getTFURLBody(mirrorURL)
 	if err != nil {
 		return nil, err
 	}
 
 	var tfVerList tfVersionList
-	getVersionsFromBody(result, preRelease, &tfVerList)
+	err = getVersionsFromJSON(product, body, preRelease, &tfVerList)
+	if err != nil {
+		getVersionsFromBody(body, preRelease, &tfVerList)
```
yermulnik, line 112: I'd suggest to add a notice-level message to add clarity by adding user-facing visibility on falling back from JSON to plain HTML (from new to old behavior).
Maybe even info- or even warn-level — to make it very clear that there's a fallback happening in place because an expected data format isn't available.

lib/list_versions.go
```diff
@@ -69,57 +115,36 @@ func getTFList(mirrorURL string, preRelease bool) ([]string, error) {
...
 	if err != nil {
 		return "", err
 	}
-	// Getting versions from body; should return match /X.X.X/ where X is a number
-	semver := `\/?` + regexSemVer.Patch.String() + `\/?"`
-	r, errSemVer := regexp.Compile(semver)
-	if errSemVer != nil {
-		return "", fmt.Errorf("Error compiling %q regex: %v", semver, errSemVer)
+	if len(versions) == 0 {
+		return "", fmt.Errorf("No versions available")
```
yermulnik: I guess it's best to also provide some clarity on the source that failed providing versions. Something like this:
```suggestion
		return "", fmt.Errorf("No versions matching predefined pattern are available at %s", mirrorURL)
```

lib/list_versions.go
```diff
@@ -28,6 +31,46 @@ var regexSemVer = struct {
...
+			parsedVersion, err := version.NewSemver(versionItx)
+			if err != nil {
+				continue
+			}
+			parsedVersions = append(parsedVersions, parsedVersion)
+		}
+	}
+
+	// Sort versions and store in tflist
+	sort.Sort(version.Collection(parsedVersions))
```
yermulnik: Please consider this note from https://pkg.go.dev/sort#Sort (if it is applicable at all):
> Note: in many situations, the newer [slices.SortFunc](https://pkg.go.dev/slices#SortFunc) function is more ergonomic and runs faster.

lib/list_versions.go
```diff
@@ -69,57 +115,36 @@ func getTFList(mirrorURL string, preRelease bool) ([]string, error) {
...
-	bodyLines := strings.Split(result, "\n")
-	for i := range result {
-		if r.MatchString(bodyLines[i]) {
-			str := r.FindString(bodyLines[i])
-			trimstr := strings.Trim(str, "/\"") // remove '/' or '"' from /X.X.X/" or /X.X.X"
-			return trimstr, nil
-		}
-	}
-	return "", nil
+	return versions[0], nil
```
yermulnik, line 132:
```suggestion
    // The list is ordered in ascending order, hence return the very
    // first element which in this case is considered the "latest"
	return versions[0], nil
```

lib/list_versions_test.go
```diff
@@ -89,20 +97,55 @@ func compareLists(actual []string, expected []string) error {
 	return nil
 }
 
-func getMockListVersionServer() *httptest.Server {
+type MockListVersionServerConfig struct {
+	EnableHashicorpList bool
+	EnableHashicorpJSON bool
+	EnableOpentofuList  bool
+	EnableOpentofuJSON  bool
```
yermulnik: For clarity's sake, I'd suggest to replace `List` with `HTML`. Not critical and hence up to you though.

lib/list_versions_test.go
```diff
@@ -146,47 +191,161 @@ func TestGetVersionsFromBodyOpenTofu(t *testing.T) {
...
+		t.Errorf("Parsed version does not match expected versions: %v", cmpErr)
+	}
+
+	// Test pre-release
+	var testTfVersionListPre tfVersionList
+	err = getVersionsFromJSON(product, hashicorpJSONData, true, &testTfVersionListPre)
+	assert.NoError(t, err)
+	expectedVersion = []string{"0.12.3-beta1", "0.12.2", "0.12.1", "0.12.0", "0.12.0-rc1", "0.12.0-beta2", "0.11.13"}
+	if cmpErr := compareLists(testTfVersionListPre.tflist, expectedVersion); cmpErr != nil {
+		t.Errorf("Parsed version does not match expected versions: %v", cmpErr)
```
yermulnik: Just to ease finding the correct place that raised this error:
```suggestion
		t.Errorf("Parsed pre-release version does not match expected versions: %v", cmpErr)
```

lib/list_versions_test.go
```diff
@@ -146,47 +191,161 @@ func TestGetVersionsFromBodyOpenTofu(t *testing.T) {
...
+		{"Hashicorp JSON", GetProductById("terraform"), MockListVersionServerConfig{EnableHashicorpJSON: true}, "terraform/index.json", "0.12.2"},
+		{"Hashicorp List", GetProductById("terraform"), MockListVersionServerConfig{EnableHashicorpList: true}, "hashicorp", "0.12.2"},
+		{"Opentofu JSON", GetProductById("opentofu"), MockListVersionServerConfig{EnableOpentofuJSON: true}, "tofu/api.json", "1.7.0"},
+		{"Opentofu List", GetProductById("opentofu"), MockListVersionServerConfig{EnableOpentofuList: true}, "opentofu/", "1.7.0"},
 	}
-	expectedVersion := "0.12.2"
-	if version != expectedVersion {
-		t.Errorf("Expected latest version does not match. Expected: %s, actual: %s", expectedVersion, version)
+
+	for _, tt := range tests {
```
yermulnik: Oddly named iterators are hard to read =(
```suggestion
	for _, test := range tests {
```

lib/product.go
```diff
@@ -169,22 +223,23 @@ func (p OpenTofuProduct) GetFileExtensions() []string {
...
+			FileExtensions:        []string{"tf"},
 		},
 	},
 	OpenTofuProduct{
 		ProductDetails{
 			ID:                    "opentofu",
 			Name:                  "OpenTofu",
-			DefaultMirror:         "https://get.opentofu.org/tofu",
+			DefaultMirror:         "https://get.opentofu.org/tofu/api.json",
 			DefaultDownloadMirror: "https://github.com/opentofu/opentofu/releases/download",
```
yermulnik, line 243: Would it make sense to also switch to https://get.opentofu.org/tofu/ for downloads along with the versions retrieval change?
At the very least it would allow to bypass pretty much strict GitHub rate limits applied to anon requests 🤔

lib/product_test.go
```diff
@@ -99,6 +99,14 @@ func Test_GetArtifactUrl_Terraform(t *testing.T) {
 	}
 }
 
+func Test_GetArtifactUrl_Terraform_DefaultMirror(t *testing.T) {
+	product := GetProductById("terraform")
+	actual := product.GetArtifactUrl(product.GetDefaultMirrorUrl(), "5.3.2")
+	if expected := "https://releases.hashicorp.com/terraform/5.3.2"; actual != expected {
+		t.Errorf("Product GetArchivePrefix does not match expected ID. Expected: %q, actual: %q", expected, actual)
```
yermulnik, line 106: The `GetArchivePrefix` and the `ID` in the context of getting the URL sound a bit confusing. I might be missing some context though. Could you please either name these a bit more clearly or add a comment explaining what they mean? Thanks.

I'd also suggest to convert `5.3.2` to a var, given it's used twice.

Also, I might be not getting the gist, but the `5.3.2` seems to be intentionally incorrect (non-existing) version and hence adding corresponding comment (or renaming the Test altogether) would ease understanding the test's essence. Thanks.

lib/list_versions.go
MatthewJohn: I was warey of adding anything that would be displayed to users in a successful run if their using a mirror

lib/product.go
MatthewJohn: Do we have a ticket for it? I'd be worried about shoehorning it in under the guise of another ticket

lib/list_versions.go
MatthewJohn: Fixed in e831451

lib/list_versions.go
MatthewJohn: Fixed in 63a7a92d31292a23fd52040095e20f7df7a10edf

lib/list_versions.go
MatthewJohn: Fixed in b88b6d9

lib/list_versions.go
MatthewJohn: Fixed in b8e65c5

lib/list_versions.go
MatthewJohn: Fixed

lib/list_versions.go
MatthewJohn: Added in f97cc57

lib/list_versions.go
MatthewJohn: "decending"? :) But have fixed

lib/list_versions_test.go
MatthewJohn: Fixed

lib/list_versions_test.go
MatthewJohn: Fixed :)

lib/list_versions_test.go
MatthewJohn: Fixed

lib/param_parsing/versiontf_test.go
MatthewJohn: Fixed

lib/param_parsing/versiontf_test.go
MatthewJohn: Fixed

lib/list_versions.go
```diff
@@ -28,6 +31,49 @@ var regexSemVer = struct {
...
+	}
+
+	parsedVersions := make([]*version.Version, 0, len(versionList))
+	for versionItem := range slices.Values(versionList) {
+		// Sanity check that it matches our regex
+		// Could be removed in future, perhaps, if we switch to Hashicorp library entirely.
+		if r.MatchString(versionItem) {
+			parsedVersion, err := version.NewSemver(versionItem)
+			if err != nil {
+				logger.Debug("Failed to parse version using go-version: \"%s\", skipping", versionItem)
```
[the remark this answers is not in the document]
MatthewJohn: Fixed in 36fce96

lib/list_versions.go
yermulnik: > "decending"? :)

They say "ascending" =)
https://pkg.go.dev/sort#Sort
> Sort sorts data in ascending order as determined by the Less method.

lib/product.go
yermulnik: > Do we have a ticket for it? I'd be worried about shoehorning it in under the guise of another ticket

Not yet. Let me create one: #744

lib/list_versions_test.go
```diff
@@ -146,47 +191,161 @@ func TestGetVersionsFromBodyOpenTofu(t *testing.T) {
...
+		{"Hashicorp JSON", GetProductById("terraform"), MockListVersionServerConfig{EnableHashicorpJSON: true}, "terraform/index.json", "0.12.2"},
+		{"Hashicorp List", GetProductById("terraform"), MockListVersionServerConfig{EnableHashicorpHTML: true}, "hashicorp", "0.12.2"},
+		{"Opentofu JSON", GetProductById("opentofu"), MockListVersionServerConfig{EnableOpentofuJSON: true}, "tofu/api.json", "1.7.0"},
+		{"Opentofu List", GetProductById("opentofu"), MockListVersionServerConfig{EnableOpentofuHTML: true}, "opentofu/", "1.7.0"},
 	}
-	expectedVersion := "0.12.2"
-	if version != expectedVersion {
-		t.Errorf("Expected latest version does not match. Expected: %s, actual: %s", expectedVersion, version)
+
+	for _, test := range tests {
```
yermulnik: Would the `slices.Values()` be applicable and work here since we don't need index from `range`?
```suggestion
	for test := range slices.Values(tests) {
```

lib/list_versions_test.go
```diff
@@ -146,47 +191,161 @@ func TestGetVersionsFromBodyOpenTofu(t *testing.T) {
...
+		url          string
+		versionTests []versionTest
+	}{
+		{"Hashicorp JSON", GetProductById("terraform"), MockListVersionServerConfig{EnableHashicorpJSON: true}, "terraform/index.json", hashicorpVersions},
+		{"Hashicorp List", GetProductById("terraform"), MockListVersionServerConfig{EnableHashicorpHTML: true}, "hashicorp/", hashicorpVersions},
+		{"Opentofu JSON", GetProductById("opentofu"), MockListVersionServerConfig{EnableOpentofuJSON: true}, "tofu/api.json", opentofuVersions},
+		{"Opentofu List", GetProductById("opentofu"), MockListVersionServerConfig{EnableOpentofuHTML: true}, "opentofu/", opentofuVersions},
+	}
+
+	for _, test := range tests {
```
yermulnik: Similarly here, would the `slices.Values()` fit the use case?

ps: not nitpicking, just thinking whether we can embrace new stuff from Go as per https://go.dev/blog/range-functions#new-iterator-functions

lib/product.go
```diff
@@ -136,6 +176,7 @@ func (p OpenTofuProduct) GetArchivePrefix() string {
 
 // nolint:revive // FIXME: parameter 'mirrorURL' is not used (custom Mirror URL is not implemented for OpenTofu? 10-Mar-2025)
 // nolint:revive // FIXME: var-naming: method GetArtifactUrl should be GetArtifactURL (revive)
+// @TODO For a future release, use user-provided mirror, as we ONLY allow downloading via public OpenTofu URL
```
yermulnik, line 179: Would it make sense to create an issue for this so it doesn't get missed?

lib/product_test.go
yermulnik: > https://github.com/warrensbox/terraform-switcher/pull/739#discussion_r3107543288

Same applies to `Test_GetArtifactUrl_OpenTofu_DefaultMirror()` please.

lib/product.go
yermulnik: I created one: #745

MatthewJohn: 👍 thanks

lib/product_test.go
MatthewJohn: The repetition of value is mostly for clarity and absoluteness, the general ideal being, if I have the following:
```bash

func generateUrl(protocol, baseUrl, subUrl, os, arch, version, suffix, extension string) string {
    return protocol + baseUrl + "/" subUrl + "/" os + "_" + arch + "-" version + suffix + "." + extension
}

func test_generateUrl() {
    protocol, baseUrl, subUrl, os, arch, version, suffix, extension := "https", "example.com", "test", "linux" "amd64", "1.2.3", "test", "exe"
    actualUrl := generateUrl(protocol, baseUrl, subUrl, os, arch, version, suffix, extension)
    expectedUrl = protocol + baseUrl + "/" subUrl + "/" os + "_" + arch + "-" version + suffix + "." + extension
    assert.Equal(expectedUrl, actualUrl)
}
```
The test ends up being copy+paste logic, which not only perhaps tied to implementation of function, but it also doesn't really prove anything

Instead, having:
```
func test_generateUrl() {
    actualUrl := generateUrl("https", "example.com", "test", "linux" "amd64", "1.2.3", "test", "exe")
    assert.Equal(actualUrl, "https://example.com/test/linux-amd64_1.2.3-test.exe")
}
```
clearly shows what we're expected from the URL - yes the value is duplicated, but realistically the test shouldn't need to change and it makes crystal clear when the function doesn't return what we're expecting and helps visually clarify what's being returned.

Though, the code we're talking about is simpler with this regard, this is mostly the reason I wouldn't variablise everything between the "do" and the "verify" of the tests (of course, this doesn't mean I wouldn't create variables if I was going to re-use values in the setup/"do"ing aspect - but I would think twice before re-using them in the verification.

With regards to the version number itself, this is just a testing a static URL generation function, the version itself is truly meaningless

yermulnik: I guess I phrased my comment oddly. What I meant is that in `t.Errorf("Product GetArchivePrefix does not match expected ID. Expected: %q, actual: %q", expected, actual)` it is not clear to me where `GetArchivePrefix` and `ID` come from and what they mean.

Same for `5.3.2` string — is it random deliberately erroneous value or what is it in the essence?

These two points make these two tests unclear and a bit confusing to me 🤷🏻 

ℹ️ I'm going to approve the PR now, but I'd suggest to eliminate these bits of ambiguity before we merge this PR. Thanks.

yermulnik: Maybe just replace `t.Errorf("Product GetArchivePrefix does not match expected ID. Expected: %q, actual: %q", expected, actual)` with something like `t.Errorf("Product GetArtifactUrl match failed. Expected: %q, actual: %q", expected, actual)`

yermulnik (approved):

lib/product_test.go
yermulnik: > Same for `5.3.2` string — is it random deliberately erroneous value or what is it in the essence?

Similarly I'd suggest to replace `5.3.2` with something like `random-meaningless-value` — so that it doesn't introduce any bits of confusion like it did to me 🤷🏻

lib/list_versions.go
yermulnik: +1

lib/product.go
```diff
@@ -179,14 +233,15 @@ func (p OpenTofuProduct) GetFileExtensions() []string {
...
-			VersionPrefix:  "terraform_",
-			ExecutableName: "terraform",
-			ArchivePrefix:  "terraform_",
-			PublicKeyId:    "72D7468F",
-			PublicKeyURLs:  []string{"https://www.hashicorp.com/.well-known/pgp-key.txt", "https://keybase.io/hashicorp/pgp_keys.asc"},
+			ID:                    "terraform",
+			Name:                  "Terraform",
+			DefaultMirror:         "https://releases.hashicorp.com/terraform/index.json",
+			DefaultDownloadMirror: "https://releases.hashicorp.com/terraform",
+			VersionPrefix:         "terraform_",
```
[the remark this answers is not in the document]
MatthewJohn: Fixed in [8213816](https://github.com/warrensbox/terraform-switcher/pull/739/commits/821381694f4d8866b6ce2d0872aa405d6a8a08c2)

lib/param_parsing/parameters.go
```diff
@@ -159,6 +186,8 @@ func populateParams(params Params) Params {
...
+
```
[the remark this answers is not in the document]
MatthewJohn: This is pre-existing, please either create a ticket or provide me with the details.

@copilot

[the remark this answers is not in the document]
yermulnik: @MatthewJohn I guess what it hints at is that the `setupProductParam(&params)` should be added right after second `params = GetParamsFromEnvironment(params)` call at the end of the inner block (right after line 232 in your PR):
image

lib/list_versions.go
```diff
@@ -69,57 +119,38 @@ func getTFList(mirrorURL string, preRelease bool) ([]string, error) {
...
 		}
-		versions := strings.Split(body, "\n")
-		for i := range versions {
-			if r.MatchString(versions[i]) {
-				str := r.FindString(versions[i])
-				trimstr := strings.Trim(str, "/\"") // remove '/' or '"' from /X.X.X/" or /X.X.X"
-				return trimstr, nil
+		for _, versionItem := range tflist {
+			if r.MatchString(versionItem) {
+				return versionItem, nil
```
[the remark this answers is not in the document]
yermulnik: @MatthewJohn This Copilot's comment looks reasonable. WDYT?

lib/param_parsing/parameters.go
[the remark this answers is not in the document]
yermulnik: Talked privately and I'm resolving this thread as second call to `setupProductParam(&params)` isn't necessary as the code below the `setupProductParam(&params)` call reads things that can change the version only and that doesn't need the second call to `setupProductParam(&params)` afterwards.

lib/list_versions.go
```diff
@@ -28,6 +31,49 @@ var regexSemVer = struct {
...
+	if preRelease {
+		semver = regexSemVer.Full.String()
+	} else {
+		semver = regexSemVer.Patch.String()
+	}
+	semver = "^" + semver + "$"
+	r, err := regexp.Compile(semver)
+	if err != nil {
+		logger.Fatalf("Error compiling %q regex: %v", semver, err)
+		return err
```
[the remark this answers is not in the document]
yermulnik: @MatthewJohn That's a reasonable comment. Please take a look into this.

lib/list_versions.go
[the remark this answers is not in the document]
yermulnik: As you mentioned in Slack, the `$` anchor needs to be added to `semver` value.

lib/list_versions.go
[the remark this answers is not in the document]
MatthewJohn: The problem I see is there's two error cases 1. they're not using a JSON backend and 2. something else goes wrong.
At the moment, the PR is using the returned err to signify if it there was an issue with the decoding the JSON. Maybe switching to bool (for whether the JSON decoding worked or not) and the other for internal errors? Just so we don't constantly show JSON decoding failures to users who are using the old format

lib/list_versions.go
```diff
@@ -69,71 +122,54 @@ func getTFList(mirrorURL string, preRelease bool) ([]string, error) {
...
+	// The list is ordered in descending order, hence return the very
+	// first element which in this case is considered the "latest"
+	return versions[0], nil
 }
 
 // getTFLatestImplicit : Get the latest implicit version given the mirror URL
-func getTFLatestImplicit(mirrorURL string, preRelease bool, version string) (string, error) {
+func getTFLatestImplicit(product Product, mirrorURL string, preRelease bool, version string) (string, error) {
+	tflist, errTFList := getTFList(product, mirrorURL, preRelease) // get list of versions
+	if errTFList != nil {
```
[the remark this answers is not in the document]
MatthewJohn: Fixed in 3b851ed10acb4950d296daa52e39d79d441ff5c4

lib/list_versions.go
```diff
@@ -69,57 +119,38 @@ func getTFList(mirrorURL string, preRelease bool) ([]string, error) {
...
-		versions := strings.Split(body, "\n")
-		for i := range versions {
-			if r.MatchString(versions[i]) {
-				str := r.FindString(versions[i])
-				trimstr := strings.Trim(str, "/\"") // remove '/' or '"' from /X.X.X/" or /X.X.X"
-				return trimstr, nil
+		for _, versionItem := range tflist {
+			if r.MatchString(versionItem) {
+				return versionItem, nil
 			}
```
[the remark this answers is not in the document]
MatthewJohn: Fixed

lib/list_versions.go
```diff
@@ -38,29 +83,37 @@ func getVersionsFromBody(body string, preRelease bool, tfVersionList *tfVersionL
...
 	var tfVerList tfVersionList
-	getVersionsFromBody(result, preRelease, &tfVerList)
+	err = getVersionsFromJSON(product, body, preRelease, &tfVerList)
+	if err != nil {
+		logger.Infof("Failed to parse mirror response as JSON; falling back to extracting versions from raw body: %v", err)
+		err = getVersionsFromBody(body, preRelease, &tfVerList)
+		if err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
MatthewJohn: > If the response is valid JSON but doesn’t match the expected schema

The previous version from expecting a HTML body - do you really think it would parse as valid JSON?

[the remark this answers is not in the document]
MatthewJohn: If they have provided JSON and it's not in the correct context, if we fallback to body parsing, it might still work, but presumably something really wrong with their setup.
So falling back seems like the wrong approach.

lib/list_versions.go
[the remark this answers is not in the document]
MatthewJohn: Fixed :)

PR #747, opened 2026-04-21
fix: Verify checksum signature against every PGP key in the public key file

### LINKED ISSUE #746: Bug: PGP signature verification fails for Terraform 1.14.9 — `openpgp: key expired`
## Description
 
`tfswitch` fails to install Terraform `1.14.9` due to an expired PGP key (`72D7468F`) embedded in the tool. The verification step downloads the SHA256SUMS signature file, attempts to verify it, and fatally errors out, making it impossible to use any version of Terraform signed with this key.
 
This is a **blocking issue for CI/CD pipelines** as well as local development.
 
## Reproduction
 
```
tfswitch 1.14.9
```
 
or via `.terraform-version` / `required_version` constraint resolution (e.g. `~> 1.1` matching `1.14.9`).
 
## Error output
 
```
INFO  Verifying PGP signature of checksum file: "...terraform_1.14.9_SHA256SUMS"
ERROR Could not verify PGP signature: Signature Verification Error: Invalid signature caused by openpgp: key expired
INFO  Deleting "...terraform_1.14.9_SHA256SUMS"
INFO  Deleting "...terraform_1.14.9_SHA256SUMS.72D7468F.sig"
FATAL Error downloading: Signature of checksum file could not be verified
```
 
## Affected environments
 
| Environment | Details |
|---|---|
| macOS (arm64) | `darwin_arm64`, tfswitch `1.14.9` |
| Linux (amd64) | `linux_amd64`, GitHub Actions runner (`ubuntu-latest`) |
 
## Logs
 
**macOS (local)**
 
```
22:07:55.451 INFO  Verifying PGP signature of checksum file: "/Users/RS6399/.terraform.versions/terraform_1.14.9_SHA256SUMS"
22:07:55.453 ERROR Could not verify PGP signature: Signature Verification Error: Invalid signature caused by openpgp: key expired
22:07:55.454 FATAL Error downloading: Signature of checksum file could not be verified
```
 
**GitHub Actions (linux/amd64)**
 
```
19:51:59.602 INFO  Verifying PGP signature of checksum file: "/home/runner/.terraform.versions/terraform_1.14.9_SHA256SUMS"
19:51:59.602 ERROR Could not verify PGP signature: Signature Verification Error: Invalid signature caused by openpgp: key expired
19:51:59.603 FATAL Error downloading: Signature of checksum file could not be verified
```
 
## Root cause hypothesis
 
The PGP public key `72D7468F` bundled in the `tfswitch` binary has expired. HashiCorp periodically rotates their signing keys. The embedded key needs to be updated to the current valid HashiCorp signing key.
 
The signature file requested is `terraform_1.14.9_SHA256SUMS.72D7468F.sig`, confirming the key ID in use.
 
## Expected behavior
 
`tfswitch` should successfully verify the checksum file and install the requested Terraform version, or at minimum fall back gracefully (e.g. checksum-only verification without PGP) with a clear warning rather than a fatal error.
 
## Suggested fix
 
- Update the bundled HashiCorp PGP public key to the current non-expired key
- Consider fetching the public key dynamically from HashiCorp's keyserver or `https://www.hashicorp.com/security` to avoid this class of issue in the future
- Alternatively, expose a `--skip-pgp-verify` flag (or `TFSWITCH_SKIP_PGP` env var) as a documented escape hatch for environments where PGP verification is handled externally
## Workaround
 
None currently available without patching the binary or switching to an alternative install method (e.g. `hashicorp/setup-terraform` GitHub Action).

pszypowicz:
## Summary

Fixes #746. HashiCorp's `pgp-key.txt` at `https://www.hashicorp.com/.well-known/pgp-key.txt` currently ships **two** armored PGP public key blocks: the original block (whose self-signature expired on 2026-04-18) followed by a refreshed block with an extended self-signature. Releases such as Terraform 1.14.9 are signed using the refreshed block.

`lib/checksum.go` parses the downloaded key file with `crypto.NewKeyFromArmored`, and gopenpgp v3 is explicit that this function reads exactly one entity and errors on multi-entity input. Only the first block is ever considered, so every install that needs the refreshed key fails fatally with:

```
ERROR Could not verify PGP signature: Signature Verification Error: Invalid signature caused by openpgp: key expired
FATAL Error downloading: Signature of checksum file could not be verified
```

HashiCorp confirmed the root cause in hashicorp/terraform#38404 and the same pattern is breaking hc-install, atlantis, and older versions of tenv. See also hashicorp/terraform#38418.

## The fix

Parse every armored block from the key file and register each parsed key with gopenpgp's verify handle builder. Successive `VerificationKey(k)` calls append to the builder's internal keyring, and `VerifyDetached` picks the key matching the signature's KeyID automatically.

```go
keys, err := parsePublicKeys(string(keyFileContent))
// ...
verifyBuilder := crypto.PGP().Verify()
for _, key := range keys {
    verifyBuilder = verifyBuilder.VerificationKey(key)
}
signingKey, err := verifyBuilder.New()
```

`strings.Split` on the BEGIN marker is the pragmatic parse strategy: `armor.Decode(io.Reader)` from `ProtonMail/go-crypto` documents that the reader is unusable after one call so it cannot be looped, and gopenpgp v3 does not expose a plural `NewKeysFromArmored`. Blocks that fail to parse are logged at Debug and skipped; an error is returned only when zero usable keys remain.

## Test matrix

`lib/checksum_test.go` (new unit tests, ephemeral keys generated via the pattern already used in `lib/download_test.go`):

- `Test_parsePublicKeys_singleBlock` - one armored block returns one key.
- `Test_parsePublicKeys_multipleBlocks` - two armored blocks return both keys in file order.
- `Test_parsePublicKeys_emptyInput` - empty / whitespace-only input returns an error.
- `Test_parsePublicKeys_mixedValidAndGarbage` - valid block plus a malformed block returns only the valid key; malformed block is Debug-logged and skipped.
- `Test_checkSignatureOfChecksums_singleKey` - happy-path regression for the single-block case (backward compatibility).
- `Test_checkSignatureOfChecksums_signerIsFirst` - two keys, the one at position 0 signed the checksum: verifies.
- `Test_checkSignatureOfChecksums_signerIsSecond` - two keys, the one at position 1 signed the checksum. **Direct regression test for this issue.**
- `Test_checkSignatureOfChecksums_noMatchingKey` - signature made by a key absent from the file: rejected, error mentions `tried 2 keys`.
- `Test_checkSignatureOfChecksums_malformedKeyFile` - key file contains garbage or only a marker with no body: rejected without panic.

`lib/download_test.go`:

- `DownloadProductTestConfig` gains a `SecondaryPublicKey` field; when set, `setupTestDownloadServer` serves the secondary key concatenated in front of `PublicKey` (two newlines between them) so the signing key ends up second - the shape of HashiCorp's current file.
- `TestDownloadProductFromURL_multiple_public_keys` exercises the full download + verification path against a two-block key file.

## Verification

Locally reproducing the `build.yml` pipeline (`go vet`, `go fmt`, `go build`, `go test -v ./...`, plus the `test-data/integration-tests/*` loop) comes back clean. `golangci-lint` with `.github/linters/.golangci.yml` reports no new issues. `govulncheck` reports no vulnerabilities.

End-to-end smoke test against the real HashiCorp endpoint (with a fresh key cache):

```
$ rm -f ~/.local/share/.terraform.versions/terraform_72D7468F.asc
$ ./build/tfswitch 1.14.9
INFO  Verifying PGP signature of checksum file: ".../terraform_1.14.9_SHA256SUMS"
INFO  Checksum file PGP signature verification successful
INFO  Switched Terraform to version "1.14.9"
```

## Scope

The issue also suggested adding a `--skip-pgp-verify` escape hatch. I've kept that out of this PR because the multi-key fix removes the motivating failure and a new flag would expand the security surface. Similarly, adding TTL-based refresh to the cached key file (so stale caches self-heal on future rotations) felt like a separate concern worth its own PR. Happy to follow up on either if a maintainer would like them.

No user-facing behavior changes beyond "installs that used to fail now succeed", so no README or `www/docs/usage/` updates. `CHANGELOG.md` is generated from PR labels per `CHANGELOG.howto.md`; the auto-applied `enhancement` label (and/or a manually added `bug`) routes this entry into the next release's **Fixed** section.

## Test plan

- [x] `go vet ./...`
- [x] `go fmt ./...` (no changes)
- [x] `go build -v ./...`
- [x] `go test -v ./...` (full suite passes, including 9 new tests)
- [x] Integration test loop in `test-data/integration-tests/*` (all 7 pass)
- [x] `golangci-lint` with the repo config (no new issues)
- [x] `govulncheck` (clean)
- [x] Manual install of Terraform 1.14.9 succeeds with a fresh key cache
- [x] Manual install of Terraform 1.14.8 still succeeds (backward compatibility)

### DIFF
--- a/lib/checksum.go
+++ b/lib/checksum.go
@@ -80,7 +80,7 @@ func parsePublicKeys(armored string) ([]*crypto.Key, error) {
 	// preserves the mandatory blank line between the marker/headers and
 	// the base64 payload that RFC 4880 armor requires.
 	parts := strings.Split(armored, pgpPublicKeyBegin)
-	keys := make([]*crypto.Key, 0, len(parts))
+	keys := make([]*crypto.Key, 0, len(parts)-1)
 	for i, part := range parts {
 		if i == 0 {
 			continue
--- a/lib/checksum.go
+++ b/lib/checksum.go
@@ -122,6 +122,10 @@ func checkSignatureOfChecksums(keyFile *os.File, hashFile *os.File, signatureFil
 	}
 
 	verifyBuilder := crypto.PGP().Verify()
+	// Parse every armored block from the key file and register each parsed key
+	// with verify handle builder. Successive VerificationKey() calls append
+	// to the builder's internal keyring, and the key matching the signature's
+	// KeyID is picked automatically.
 	for key := range slices.Values(keys) {
 		verifyBuilder = verifyBuilder.VerificationKey(key)
 	}
--- a/lib/checksum.go
+++ b/lib/checksum.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"crypto/sha256"
 	"encoding/hex"
+	"errors"
 	"io"
 	"os"
 	"path/filepath"
@@ -11,6 +12,11 @@ import (
 
 	"github.com/ProtonMail/gopenpgp/v3/crypto"
 )
+
+// pgpPublicKeyBegin is the ASCII armor marker that delimits every public
+// key block in a concatenated armored file (such as HashiCorp's
+// pgp-key.txt, which publishes more than one key around rotations).
+const pgpPublicKeyBegin = "-----BEGIN PGP PUBLIC KEY BLOCK-----"
 
 // getChecksumFromFile Extract the checksum from the signature file
 func getChecksumFromHashFile(signatureFilePath string, terraformFileName string) (string, error) {
@@ -58,6 +64,40 @@ func checkChecksumMatches(hashFile string, targetFile *os.File) bool {
 	return true
 }
 
+// parsePublicKeys extracts every armored PGP public key from the given
+// blob. HashiCorp's pgp-key.txt contains more than one key around key
+// rotations, and the key that signed a release may not be the first
+// block. crypto.NewKeyFromArmored refuses multi-entity input, so the
+// caller must feed it one armored block at a time.
+//
+// Blocks that fail to parse are logged at Debug and skipped; the
+// function returns an error only when no usable keys remain.
+func parsePublicKeys(armored string) ([]*crypto.Key, error) {
+	// strings.Split leaves the text preceding the first BEGIN marker in
+	// parts[0] (always non-key content); each subsequent entry is the
+	// body of one block. Re-prepending the marker rather than trimming
+	// preserves the mandatory blank line between the marker/headers and
+	// the base64 payload that RFC 4880 armor requires.
+	parts := strings.Split(armored, pgpPublicKeyBegin)
+	keys := make([]*crypto.Key, 0, len(parts))
+	for i, part := range parts {
+		if i == 0 {
+			continue
+		}
+		block := pgpPublicKeyBegin + part
+		key, err := crypto.NewKeyFromArmored(block)
+		if err != nil {
+			logger.Debugf("Skipping unparseable PGP public key block: %v", err)
+			continue
+		}
+		keys = append(keys, key)
+	}
+	if len(keys) == 0 {
+		return nil, errors.New("no parseable PGP public keys found in key file")
+	}
+	return keys, nil
+}
+
 // checkSignatureOfChecksums This will verify the signature of the file containing the hash sums
 func checkSignatureOfChecksums(keyFile *os.File, hashFile *os.File, signatureFile *os.File) bool {
 	var fileHandlersToClose []*os.File
@@ -74,13 +114,17 @@ func checkSignatureOfChecksums(keyFile *os.File, hashFile *os.File, signatureFil
 		return false
 	}
 
-	keyFromArmored, err := crypto.NewKeyFromArmored(string(keyFileContent))
+	keys, err := parsePublicKeys(string(keyFileContent))
 	if err != nil {
-		logger.Errorf("Could not read PGP armored key: %v", err)
+		logger.Errorf("Could not parse PGP keys from %q: %v", keyFile.Name(), err)
 		return false
 	}
 
-	signingKey, err := crypto.PGP().Verify().VerificationKey(keyFromArmored).New()
+	verifyBuilder := crypto.PGP().Verify()
+	for _, key := range keys {
+		verifyBuilder = verifyBuilder.VerificationKey(key)
+	}
+	signingKey, err := verifyBuilder.New()
 	if err != nil {
 		logger.Errorf("Could not read PGP signing key: %v", err)
 		return false
@@ -105,7 +149,7 @@ func checkSignatureOfChecksums(keyFile *os.File, hashFile *os.File, signatureFil
 	}
 
 	if err := verifyRes.SignatureError(); err != nil {
-		logger.Errorf("Could not verify PGP signature: %v", err)
+		logger.Errorf("Could not verify PGP signature (tried %d keys): %v", len(keys), err)
 		return false
 	}
 
--- a/lib/checksum.go
+++ b/lib/checksum.go
@@ -154,7 +154,7 @@ func checkSignatureOfChecksums(keyFile *os.File, hashFile *os.File, signatureFil
 	}
 
 	if err := verifyRes.SignatureError(); err != nil {
-		logger.Errorf("Could not verify PGP signature (tried %d keys): %v", len(keys), err)
+		logger.Errorf("Could not verify PGP signature using %d loaded keys from %q: %v", len(keys), keyFile.Name(), err)
 		return false
 	}
 
--- a/lib/checksum.go
+++ b/lib/checksum.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/ProtonMail/gopenpgp/v3/crypto"
@@ -87,13 +88,13 @@ func parsePublicKeys(armored string) ([]*crypto.Key, error) {
 		block := pgpPublicKeyBegin + part
 		key, err := crypto.NewKeyFromArmored(block)
 		if err != nil {
-			logger.Debugf("Skipping unparseable PGP public key block: %v", err)
+			logger.Debugf("Skipping unparsable PGP public key block №%d: %v", i, err)
 			continue
 		}
 		keys = append(keys, key)
 	}
 	if len(keys) == 0 {
-		return nil, errors.New("no parseable PGP public keys found in key file")
+		return nil, errors.New("no parsable PGP public keys found in key file")
 	}
 	return keys, nil
 }
@@ -121,7 +122,7 @@ func checkSignatureOfChecksums(keyFile *os.File, hashFile *os.File, signatureFil
 	}
 
 	verifyBuilder := crypto.PGP().Verify()
-	for _, key := range keys {
+	for key := range slices.Values(keys) {
 		verifyBuilder = verifyBuilder.VerificationKey(key)
 	}
 	signingKey, err := verifyBuilder.New()
--- a/lib/checksum_test.go
+++ b/lib/checksum_test.go
@@ -3,6 +3,7 @@ package lib
 import (
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"sync"
 	"testing"
@@ -57,7 +58,7 @@ func sharedTestKeys(t *testing.T) (*crypto.Key, *crypto.Key, *crypto.Key) {
 	t.Helper()
 	sharedTestKeysOnce.Do(func() {
 		pgp := crypto.PGPWithProfile(profile.RFC4880())
-		for _, target := range []**crypto.Key{&sharedTestKeyA, &sharedTestKeyB, &sharedTestKeyC} {
+		for target := range slices.Values([]**crypto.Key{&sharedTestKeyA, &sharedTestKeyB, &sharedTestKeyC}) {
 			gen := pgp.KeyGeneration().AddUserId("tfswitch-test", "tfswitch-test@example.invalid").New()
 			k, err := gen.GenerateKeyWithSecurity(constants.StandardSecurity)
 			if err != nil {
@@ -97,7 +98,10 @@ func signDetached(t *testing.T, key *crypto.Key, payload []byte) []byte {
 
 // writeReadableTempFile writes content to a fresh file under t.TempDir()
 // and returns it opened read-only. checkSignatureOfChecksums closes every
-// file it receives, so each test hands out its own handle.
+// file it receives, so each test hands out its own handle. A t.Cleanup is
+// registered to close the handle even when a test aborts before
+// checkSignatureOfChecksums runs; os.File.Close on an already-closed file
+// is harmless (returns an error we ignore), so the belt-and-braces is safe.
 func writeReadableTempFile(t *testing.T, name string, content []byte) *os.File {
 	t.Helper()
 	path := filepath.Join(t.TempDir(), name)
@@ -108,11 +112,12 @@ func writeReadableTempFile(t *testing.T, name string, content []byte) *os.File {
 	if err != nil {
 		t.Fatalf("Open: %v", err)
 	}
+	t.Cleanup(func() { _ = f.Close() })
 	return f
 }
 
 func Test_parsePublicKeys_singleBlock(t *testing.T) {
-	InitLogger("TRACE")
+	InitLogger("DEBUG")
 	keyA, _, _ := sharedTestKeys(t)
 	keys, err := parsePublicKeys(armoredPublicKey(t, keyA))
 	if err != nil {
@@ -127,7 +132,7 @@ func Test_parsePublicKeys_singleBlock(t *testing.T) {
 }
 
 func Test_parsePublicKeys_multipleBlocks(t *testing.T) {
-	InitLogger("TRACE")
+	InitLogger("DEBUG")
 	keyA, keyB, _ := sharedTestKeys(t)
 	concatenated := armoredPublicKey(t, keyA) + "\n" + armoredPublicKey(t, keyB)
 
@@ -147,8 +152,8 @@ func Test_parsePublicKeys_multipleBlocks(t *testing.T) {
 }
 
 func Test_parsePublicKeys_emptyInput(t *testing.T) {
-	InitLogger("TRACE")
-	for _, input := range []string{"", "   \n\n\t  "} {
+	InitLogger("DEBUG")
+	for input := range slices.Values([]string{"", "   \n\n\t  "}) {
 		_, err := parsePublicKeys(input)
 		if err == nil {
 			t.Errorf("parsePublicKeys(%q) expected error, got nil", input)
@@ -157,7 +162,7 @@ func Test_parsePublicKeys_emptyInput(t *testing.T) {
 }
 
 func Test_parsePublicKeys_mixedValidAndGarbage(t *testing.T) {
-	InitLogger("TRACE")
+	InitLogger("DEBUG")
 	keyA, _, _ := sharedTestKeys(t)
 	garbage := pgpPublicKeyBegin + "\nnot-a-real-armored-body\n-----END PGP PUBLIC KEY BLOCK-----\n"
 	concatenated := armoredPublicKey(t, keyA) + "\n" + garbage
@@ -175,7 +180,7 @@ func Test_parsePublicKeys_mixedValidAndGarbage(t *testing.T) {
 }
 
 func Test_checkSignatureOfChecksums_singleKey(t *testing.T) {
-	InitLogger("TRACE")
+	InitLogger("DEBUG")
 	keyA, _, _ := sharedTestKeys(t)
 	payload := []byte("abc123  terraform_1.7.5_linux_amd64.zip\n")
 	signature := signDetached(t, keyA, payload)
@@ -190,7 +195,7 @@ func Test_checkSignatureOfChecksums_singleKey(t *testing.T) {
 }
 
 func Test_checkSignatureOfChecksums_signerIsFirst(t *testing.T) {
-	InitLogger("TRACE")
+	InitLogger("DEBUG")
 	keyA, keyB, _ := sharedTestKeys(t)
 	payload := []byte("abc123  terraform_1.7.5_linux_amd64.zip\n")
 	signature := signDetached(t, keyA, payload)
@@ -210,7 +215,7 @@ func Test_checkSignatureOfChecksums_signerIsFirst(t *testing.T) {
 // key first and the active signing key second, and every release signed
 // with the successor key must still verify.
 func Test_checkSignatureOfChecksums_signerIsSecond(t *testing.T) {
-	InitLogger("TRACE")
+	InitLogger("DEBUG")
 	keyA, keyB, _ := sharedTestKeys(t)
 	payload := []byte("abc123  terraform_1.14.9_linux_amd64.zip\n")
 	signature := signDetached(t, keyB, payload)
@@ -221,12 +226,12 @@ func Test_checkSignatureOfChecksums_signerIsSecond(t *testing.T) {
 	sigFile := writeReadableTempFile(t, "SHA256SUMS.sig", signature)
 
 	if !checkSignatureOfChecksums(keyFile, hashFile, sigFile) {
-		t.Fatal("checkSignatureOfChecksums returned false with signer at position 1 (the #746 scenario)")
+		t.Fatal("checkSignatureOfChecksums returned false with signer at position 1 (GitHub issue #746 scenario)")
 	}
 }
 
 func Test_checkSignatureOfChecksums_noMatchingKey(t *testing.T) {
-	InitLogger("TRACE")
+	InitLogger("DEBUG")
 	keyA, keyB, keyC := sharedTestKeys(t)
 	payload := []byte("abc123  terraform_1.7.5_linux_amd64.zip\n")
 	signature := signDetached(t, keyC, payload)
@@ -242,7 +247,7 @@ func Test_checkSignatureOfChecksums_noMatchingKey(t *testing.T) {
 }
 
 func Test_checkSignatureOfChecksums_malformedKeyFile(t *testing.T) {
-	InitLogger("TRACE")
+	InitLogger("DEBUG")
 	keyA, _, _ := sharedTestKeys(t)
 	payload := []byte("abc123  terraform_1.7.5_linux_amd64.zip\n")
 	signature := signDetached(t, keyA, payload)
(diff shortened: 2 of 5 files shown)


lib/checksum_test.go
```diff
@@ -28,3 +35,224 @@ func Test_checkChecksumMatches(t *testing.T) {
...
+		pgp := crypto.PGPWithProfile(profile.RFC4880())
+		for _, target := range []**crypto.Key{&sharedTestKeyA, &sharedTestKeyB, &sharedTestKeyC} {
+			gen := pgp.KeyGeneration().AddUserId("tfswitch-test", "tfswitch-test@example.invalid").New()
+			k, err := gen.GenerateKeyWithSecurity(constants.StandardSecurity)
+			if err != nil {
+				t.Fatalf("PGP key generation failed: %v", err)
+			}
+			*target = k
+		}
+	})
```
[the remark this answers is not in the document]
pszypowicz: Good catch, addressed in 1533691. The generator error is now captured into a package-level `sharedTestKeysErr`; the outer helper checks it on every call and fails the calling test deterministically, instead of tripping `t.Fatalf` from inside the `sync.Once` closure (which would have marked the Once done and left subsequent tests staring at nil keys).

lib/checksum.go
```diff
@@ -58,6 +64,40 @@ func checkChecksumMatches(hashFile string, targetFile *os.File) bool {
...
+	parts := strings.Split(armored, pgpPublicKeyBegin)
+	keys := make([]*crypto.Key, 0, len(parts))
+	for i, part := range parts {
+		if i == 0 {
+			continue
+		}
+		block := pgpPublicKeyBegin + part
+		key, err := crypto.NewKeyFromArmored(block)
+		if err != nil {
+			logger.Debugf("Skipping unparseable PGP public key block: %v", err)
```
yermulnik: To ease debugging it might probably worth adding info on which key failed parsing. E.g.:
```suggestion
			logger.Debugf("Skipping unparseable PGP public key block №%d: %v", err, i)
```

lib/checksum.go
```diff
@@ -74,13 +114,17 @@ func checkSignatureOfChecksums(keyFile *os.File, hashFile *os.File, signatureFil
...
+	keys, err := parsePublicKeys(string(keyFileContent))
 	if err != nil {
-		logger.Errorf("Could not read PGP armored key: %v", err)
+		logger.Errorf("Could not parse PGP keys from %q: %v", keyFile.Name(), err)
 		return false
 	}
 
-	signingKey, err := crypto.PGP().Verify().VerificationKey(keyFromArmored).New()
+	verifyBuilder := crypto.PGP().Verify()
+	for _, key := range keys {
```
yermulnik: Instead of discarding index, please use `slices.Values()`:
```suggestion
	for key := range slices.Values(keys) {
```

ref: https://go.dev/doc/go1.23#iterators

lib/checksum.go
yermulnik: Also vocab says it's `parsable` rather than `parseable` that is more correct. Could you please replace `parseable` with `parsable` across the PR? Thanks.

lib/checksum.go
```diff
@@ -74,13 +114,17 @@ func checkSignatureOfChecksums(keyFile *os.File, hashFile *os.File, signatureFil
...
+		verifyBuilder = verifyBuilder.VerificationKey(key)
+	}
+	signingKey, err := verifyBuilder.New()
```
yermulnik: I'm a bit confused and seem to be missing something essential: the loop takes all parsable signing keys, overwrites `verifyBuilder` until the last key reached, and then does `verifyBuilder.New()` with the last key.
Could you please help me figure this out? 🤔

lib/checksum_test.go
```diff
@@ -28,3 +35,235 @@ func Test_checkChecksumMatches(t *testing.T) {
...
+	sharedTestKeyB     *crypto.Key
+	sharedTestKeyC     *crypto.Key
+	sharedTestKeysErr  error
+)
+
+func sharedTestKeys(t *testing.T) (*crypto.Key, *crypto.Key, *crypto.Key) {
+	t.Helper()
+	sharedTestKeysOnce.Do(func() {
+		pgp := crypto.PGPWithProfile(profile.RFC4880())
+		for _, target := range []**crypto.Key{&sharedTestKeyA, &sharedTestKeyB, &sharedTestKeyC} {
```
yermulnik:
```suggestion
		for target := range slices.Values([]**crypto.Key{&sharedTestKeyA, &sharedTestKeyB, &sharedTestKeyC}) {
```

lib/checksum_test.go
```diff
@@ -28,3 +35,235 @@ func Test_checkChecksumMatches(t *testing.T) {
...
+	}
+	f, err := os.Open(path)
+	if err != nil {
+		t.Fatalf("Open: %v", err)
+	}
+	return f
+}
+
+func Test_parsePublicKeys_singleBlock(t *testing.T) {
+	InitLogger("TRACE")
```
yermulnik: Is `TRACE` indeed necessary and useful here? Might `DEBUG` suffice in the use case as we're usually using `DEBUG` for tests?

lib/checksum.go
```diff
@@ -58,6 +64,40 @@ func checkChecksumMatches(hashFile string, targetFile *os.File) bool {
...
+	// body of one block. Re-prepending the marker rather than trimming
+	// preserves the mandatory blank line between the marker/headers and
+	// the base64 payload that RFC 4880 armor requires.
+	parts := strings.Split(armored, pgpPublicKeyBegin)
+	keys := make([]*crypto.Key, 0, len(parts))
+	for i, part := range parts {
+		if i == 0 {
+			continue
+		}
+		block := pgpPublicKeyBegin + part
```
yermulnik, line 88: I'm not entirely sure, though wouldn't `\n` be necessary between the marker and the `part`? Or is it left there after split operation? 🤔

lib/checksum_test.go
```diff
@@ -28,3 +35,235 @@ func Test_checkChecksumMatches(t *testing.T) {
...
+		t.Errorf("first key fingerprint mismatch: want %s, got %s", keyA.GetFingerprint(), keys[0].GetFingerprint())
+	}
+	if keys[1].GetFingerprint() != keyB.GetFingerprint() {
+		t.Errorf("second key fingerprint mismatch: want %s, got %s", keyB.GetFingerprint(), keys[1].GetFingerprint())
+	}
+}
+
+func Test_parsePublicKeys_emptyInput(t *testing.T) {
+	InitLogger("TRACE")
+	for _, input := range []string{"", "   \n\n\t  "} {
```
yermulnik:
```suggestion
	for input := range slices.Values([]string{"", "   \n\n\t  "}) {
```

lib/checksum_test.go
```diff
@@ -28,3 +35,235 @@ func Test_checkChecksumMatches(t *testing.T) {
...
+
+func Test_checkSignatureOfChecksums_singleKey(t *testing.T) {
+	InitLogger("TRACE")
+	keyA, _, _ := sharedTestKeys(t)
+	payload := []byte("abc123  terraform_1.7.5_linux_amd64.zip\n")
+	signature := signDetached(t, keyA, payload)
+
+	keyFile := writeReadableTempFile(t, "pubkey.asc", []byte(armoredPublicKey(t, keyA)))
+	hashFile := writeReadableTempFile(t, "SHA256SUMS", payload)
+	sigFile := writeReadableTempFile(t, "SHA256SUMS.sig", signature)
```
yermulnik, line 188: It is probably best to clean up temp files once the function completes.
Please find an example `t.Cleanup` func in `TestDownloadFromURL_FileNameMatch()` in `lib/download_test.go`

lib/checksum_test.go
```diff
@@ -28,3 +35,235 @@ func Test_checkChecksumMatches(t *testing.T) {
...
+	payload := []byte("abc123  terraform_1.14.9_linux_amd64.zip\n")
+	signature := signDetached(t, keyB, payload)
+	keyring := armoredPublicKey(t, keyA) + "\n" + armoredPublicKey(t, keyB)
+
+	keyFile := writeReadableTempFile(t, "pubkey.asc", []byte(keyring))
+	hashFile := writeReadableTempFile(t, "SHA256SUMS", payload)
+	sigFile := writeReadableTempFile(t, "SHA256SUMS.sig", signature)
+
+	if !checkSignatureOfChecksums(keyFile, hashFile, sigFile) {
+		t.Fatal("checkSignatureOfChecksums returned false with signer at position 1 (the #746 scenario)")
```
yermulnik:
```suggestion
		t.Fatal("checkSignatureOfChecksums returned false with signer at position 1 (GitHub issue #746 scenario)")
```

lib/download_test.go
```diff
@@ -97,6 +97,11 @@ type DownloadProductTestConfig struct {
 	ZipFileChecksum     string
 	ChecksumFileContent string
 	PublicKey           string
+	// SecondaryPublicKey, when non-empty, is prepended to PublicKey in
+	// the armored response so that the signing key sits second in the
+	// concatenated file. This mirrors HashiCorp's pgp-key.txt layout
+	// during key rotations and is the shape that regressed in #746.
```
yermulnik:
```suggestion
	// during key rotations and is the shape that regressed in GitHub issue #746.
```

yermulnik (commented):
Thank you for the contribution ❤️ 
Please consider my comments and suggestions.

yermulnik:
@pszypowicz For the `GO_MODULES` linter failure (https://github.com/warrensbox/terraform-switcher/actions/runs/24719467676) please pick the `gosec` suppressions from https://github.com/warrensbox/terraform-switcher/pull/739/changes#diff-90697d6eae2e2416dcb80680a07d02ed87ac625142983a3f4ce392f236f4be54R364 and https://github.com/warrensbox/terraform-switcher/pull/739/changes#diff-7ed1133505ef53325566e6105e151d84014e8fe42002dff95f056a77c402767dR60 (or otherwise I can push to your branch — let me know). Thanks.

lib/checksum.go
yermulnik: Ah, I guess it's already outlined in the PR description:
> Parse every armored block from the key file and register each parsed key with gopenpgp's verify handle builder. Successive `VerificationKey(k)` calls append to the builder's internal keyring, and `VerifyDetached` picks the key matching the signature's KeyID automatically.

lib/checksum.go
pszypowicz: Good question - `strings.Split(s, marker)` removes only the marker, so the `\n` that originally followed it stays as the first byte of `part`. `pgpPublicKeyBegin + part` already has the blank line RFC 4880 requires between the marker/headers and the base64 payload - no extra `\n` needed. `Test_parsePublicKeys_multipleBlocks` exercises the two-block case end to end, so a missing separator would fail the suite. Happy to reword the comment above the line if it would read better.

lib/checksum.go
pszypowicz: That was my first instinct too, but `VerificationKey` actually appends rather than replaces. In gopenpgp v3.4.0 (`crypto/verify_handle_builder.go` lines 25-35) the first call creates the internal `VerifyKeyRing` via `NewKeyRing`, and every subsequent call routes through `AddKey` on that same keyring. After the loop the builder holds every key, and `verifyBuilder.New()` produces a handle that verifies against any of them - gopenpgp picks the one matching the signature's KeyID.

`Test_checkSignatureOfChecksums_signerIsSecond` is the direct proof: drop the loop and that test fails because the builder would only carry the expired first key.

If you'd prefer the intent to be obvious at a glance, I can swap the loop for a single `verifyBuilder.VerificationKeys(keyRing)` call (building the keyring from all parsed keys first). Let me know - happy either way.

pszypowicz:
Picked both suppressions in 89d490a, thanks - `golangci-lint run --config .github/linters/.golangci.yml ./...` is clean locally now. The `//nolint:gosec` comments match the wording from #739 verbatim.

lib/download_test.go
```diff
@@ -97,6 +97,11 @@ type DownloadProductTestConfig struct {
 	ZipFileChecksum     string
 	ChecksumFileContent string
 	PublicKey           string
+	// SecondaryPublicKey, when non-empty, is prepended to PublicKey in
+	// the armored response so that the signing key sits second in the
+	// concatenated file. This mirrors HashiCorp's pgp-key.txt layout
+	// during key rotations and is the shape that regressed in GitHub issue #746.
+	SecondaryPublicKey string
```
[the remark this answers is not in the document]
pszypowicz: Fair call - renamed to `LeadingPublicKey` in 049cf0f, which matches the "first block in the served file" behavior. Updated the doc comment alongside.

lib/checksum.go
yermulnik: Nah, no need to replace it with something else.
Just, for the sake of clarity, please add this as comment (given we both were caught by this bit), e.g.:
```go
	// Parse every armored block from the key file and register each parsed key
	// with verify handle builder. Successive VerificationKey() calls append
	// to the builder's internal keyring, and the key matching the signature's
	// KeyID is picked automatically.
	for _, key := range keys {
		verifyBuilder = verifyBuilder.VerificationKey(key)
	}
	signingKey, err := verifyBuilder.New()
```

lib/checksum.go
yermulnik: Ah, great, thanks for confirming this.

lib/checksum.go
pszypowicz: Applied verbatim in 3527f97, kept the `slices.Values` iteration form since you had asked for that earlier. Thanks for the wording.

yermulnik: > kept the `slices.Values` iteration form since you had asked for that earlier.

Ah, apologies for the confusion, copy&pasta issue =))

lib/checksum.go
```diff
@@ -105,7 +154,7 @@ func checkSignatureOfChecksums(keyFile *os.File, hashFile *os.File, signatureFil
 	}
 
 	if err := verifyRes.SignatureError(); err != nil {
-		logger.Errorf("Could not verify PGP signature: %v", err)
+		logger.Errorf("Could not verify PGP signature (tried %d keys): %v", len(keys), err)
```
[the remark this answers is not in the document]
pszypowicz: Good catch - applied verbatim in c2632a6. "Tried" did imply fallback-style retries that gopenpgp never does, and surfacing the key file path in the error makes the diagnostic actually actionable.

lib/download_test.go
```diff
@@ -384,3 +394,59 @@ func TestDownloadProductFromURL_invalid_public_key(t *testing.T) {
...
+func TestDownloadProductFromURL_multiple_public_keys(t *testing.T) {
+	logger = InitLogger("DEBUG")
+
+	// Generate a throwaway companion key. It never signs anything and is
+	// unrelated to the signing key that setupTestDownloadServer builds;
+	// its job is to be the first armored block so the code under test
+	// has to skip past it and try the second one.
+	pgp4880 := crypto.PGPWithProfile(profile.RFC4880())
+	companionGen := pgp4880.KeyGeneration().AddUserId("TestProductCompanion", "companion@localhost.com").New()
+	companionKey, err := companionGen.GenerateKeyWithSecurity(constants.HighSecurity)
```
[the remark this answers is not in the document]
pszypowicz: Applied verbatim in c2632a6 - the companion key only has to render as an armored block, so RSA 2048 is identical coverage at a fraction of the keygen time.

lib/checksum_test.go
pszypowicz: Addressed in 89d490a - `writeReadableTempFile` now registers `t.Cleanup(func() { _ = f.Close() })` so the open handle is released even when a test aborts before `checkSignatureOfChecksums`' deferred close runs. The files themselves live under `t.TempDir()`, which the testing framework removes recursively at test teardown, so that leg is covered by the standard library rather than an explicit `os.Remove`.

lib/checksum_test.go
```diff
@@ -28,3 +36,239 @@ func Test_checkChecksumMatches(t *testing.T) {
...
+// keeps the full suite in the single-digit-seconds range.
+//
+// Failures inside the sync.Once closure are captured into sharedTestKeysErr
+// rather than calling t.Fatalf directly, because Once.Do marks itself done
+// regardless of how its closure exits: aborting the first caller with
+// t.Fatalf would leave every subsequent caller looking at nil keys and
+// panicking instead of reporting the real cause.
+var (
+	sharedTestKeysOnce sync.Once
+	sharedTestKeyA     *crypto.Key
```
MatthewJohn: I'm not sure why storing these indepdently, especially with neefing a list of pointer to pointers of keys. Would a slice or an array be more suitable here?

lib/checksum_test.go
yermulnik: Ah, good point, I missed this feature of `t.TempDir()`. Thanks for confirming.

lib/checksum.go
MatthewJohn: The `pgpPublicKeyBegin` is the entire string that's being split on, so adding it to the start of it wouldn't need \n

lib/checksum.go
```diff
@@ -58,6 +65,40 @@ func checkChecksumMatches(hashFile string, targetFile *os.File) bool {
...
+// Blocks that fail to parse are logged at Debug and skipped; the
+// function returns an error only when no usable keys remain.
+func parsePublicKeys(armored string) ([]*crypto.Key, error) {
+	// strings.Split leaves the text preceding the first BEGIN marker in
+	// parts[0] (always non-key content); each subsequent entry is the
+	// body of one block. Re-prepending the marker rather than trimming
+	// preserves the mandatory blank line between the marker/headers and
+	// the base64 payload that RFC 4880 armor requires.
+	parts := strings.Split(armored, pgpPublicKeyBegin)
+	keys := make([]*crypto.Key, 0, len(parts))
```
MatthewJohn:
```suggestion
	keys := make([]*crypto.Key, 0, len(parts)-1)
```
Since we're skipping the first element

lib/checksum_test.go
MatthewJohn: Is there any reason not to have an array of Key pointers? Would this make more sense than defining three separate variables?

MatthewJohn:
To me, this seems to make sense - I played with something similar locally, but didn't manage to get it polished before running out of time

yermulnik (approved):
LGTM

@pszypowicz Please consider comments/suggestions added by @MatthewJohn and we should be good to merge then I guess 🤞🏻 

@MatthewJohn Please approve this PR when you're good with updates that'll be made according to your comments. Thanks.

lib/checksum_test.go
pszypowicz: Yes, that reads much cleaner. Working shape locally:
```go
var (
    sharedTestKeysOnce sync.Once
    sharedTestKeyCache [3]*crypto.Key
    sharedTestKeysErr  error
)

func sharedTestKeys(t *testing.T) (*crypto.Key, *crypto.Key, *crypto.Key) {
    t.Helper()
    sharedTestKeysOnce.Do(func() {
        pgp := crypto.PGPWithProfile(profile.RFC4880())
        for i := range sharedTestKeyCache {
            gen := pgp.KeyGeneration().AddUserId("tfswitch-test", "tfswitch-test@example.invalid").New()
            k, err := gen.GenerateKeyWithSecurity(constants.StandardSecurity)
            if err != nil {
                sharedTestKeysErr = err
                return
            }
            sharedTestKeyCache[i] = k
        }
    })
    if sharedTestKeysErr != nil {
        t.Fatalf("PGP key generation failed: %v", sharedTestKeysErr)
    }
    return sharedTestKeyCache[0], sharedTestKeyCache[1], sharedTestKeyCache[2]
}
```

I kept the three-value return so call sites stay `keyA, keyB, _ := sharedTestKeys(t)` - happy to change the return to an array too if you'd rather callers use `keys[0]/keys[1]/keys[2]`, just wanted to check before touching every test body.

lib/checksum.go
pszypowicz: Good catch - applied verbatim in 8d78453.

lib/checksum_test.go
pszypowicz: Landed in 8d78453 with the three-value return preserved. Happy to flip the return to an array in a follow-up commit if you prefer callers to use `keys[0]/keys[1]/...`, just let me know.

MatthewJohn: Thank you - this looks good, thank you :)

MatthewJohn (approved):

lib/logging.go
```diff
@@ -57,7 +57,7 @@ func isColorLogging() bool {
 		return false
 	} else if color.SupportColor() {
 		if os.Getenv("FORCE_COLOR") == "" {
-			return term.IsTerminal(int(os.Stdout.Fd()))
+			return term.IsTerminal(int(os.Stdout.Fd())) //nolint:gosec // G115: file descriptor from os.Stdout is always a valid small integer
```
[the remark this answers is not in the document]
pszypowicz: Keeping this verbatim. Both `//nolint:gosec` lines were picked from PR #739 at @yermulnik's direction to unblock the `GO_MODULES` linter on this branch, and that PR is already on master with the same wording. Rewording here would desync from the `install.go` counterpart and from what's on master. If the Windows HANDLE concern is real it applies to master too and is worth its own issue rather than scope creep on the PGP fix.

lib/install.go
```diff
@@ -361,7 +361,7 @@ func InstallProductOption(product Product, listAll, dryRun, showRequiredFlag boo
 		}
 
 		/* prompt user to select version of product */
-		if !term.IsTerminal(int(os.Stdin.Fd())) || !term.IsTerminal(int(os.Stdout.Fd())) {
+		if !term.IsTerminal(int(os.Stdin.Fd())) || !term.IsTerminal(int(os.Stdout.Fd())) { //nolint:gosec // G115: file descriptors from os.Stdin/Stdout are always valid small integers
 			//nolint:revive // error-strings: error strings should not be capitalized or end with punctuation or a newline (revive)
```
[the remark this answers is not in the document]
pszypowicz: Same reasoning as on the `logging.go` suppression - this was picked verbatim from PR #739 at @yermulnik's direction to unblock `GO_MODULES`, and is already on master in that form. Restructuring the callers to bound-check `Fd()` goes beyond the scope of this PR; happy to open a separate issue if the Windows HANDLE behavior is worth fixing everywhere it shows up.

yermulnik:
Once #748 is merged, we'll cut a new release.
@pszypowicz Thanks for the contribution ❤️

pszypowicz:
Thank you for driving review!

yermulnik:
https://github.com/warrensbox/terraform-switcher/releases/tag/v1.17.0

PR #750, opened 2026-04-21
fix: Verify checksum signature per key block instead of via a merged keyring

### LINKED ISSUE #746: Bug: PGP signature verification fails for Terraform 1.14.9 — `openpgp: key expired`

pszypowicz:
## Summary

#747 parsed multiple armored PGP blocks from `pgp-key.txt` and registered them on a single gopenpgp verify handle, then trusted the library's KeyID lookup to pick the right one. That works when blocks carry distinct KeyIDs, which is what the #747 tests cover.

HashiCorp's `pgp-key.txt` post the 2026-04-18 rotation publishes **two blocks with identical primary fingerprint and identical signing subkey fingerprint** - only the expiration differs (legacy: expired 2026-04-18; renewed: expires 2030-03-01). A merged keyring collapses those to a single entry, so gopenpgp can latch onto the expired copy and report `openpgp: key expired` instead of falling through to the renewed block. That's exactly what users have been reporting on #746 against v1.17.0.

This change swaps the merged keyring for per-block iteration: a fresh verify handle per block, first success wins. Every Terraform release signed with either the legacy or renewed subkey now verifies end to end.

## What changed

- `lib/checksum.go` - per-block iteration in `checkSignatureOfChecksums`. The three gopenpgp failure categories (handle build / `VerifyDetached` / `SignatureError`) each log at Debug with the block index; the final Error line reports the loaded-keys count, the key file path, and the last block's error.
- `lib/checksum_test.go` - new fixture-backed regression test `Test_checkSignatureOfChecksums_hashicorpDualKey` with two subtests straddling the rotation: `pre-rotation-1.12.2` verifies against the legacy block, `post-rotation-1.14.9` verifies only against the renewed block (and is the direct reproducer for #746).
- `test-data/hashicorp-pgp-key.txt`, `terraform_1.12.2_SHA256SUMS(.sig)`, `terraform_1.14.9_SHA256SUMS(.sig)` - real public HashiCorp artifacts serving as the regression fixture. Refresh them together once the renewed subkey's expiry is within reach.

## Verification

Locally built binary installs cleanly on both sides of the rotation:

| Version | Signed with | Result |
|---|---|---|
| 1.11.4, 1.12.2, 1.13.0 | legacy subkey | `Checksum file PGP signature verification successful` |
| 1.14.9 | renewed subkey | `Checksum file PGP signature verification successful` (block 1 fires `key expired`, block 2 verifies) |

CI-equivalent checks locally:
- `go vet ./...` - clean
- `gofmt -l` on touched files - empty
- `CGO_ENABLED=0 go build -v ./...` - clean
- `CGO_ENABLED=0 go test ./...` - all packages green
- All 7 integration-test dirs in `test-data/integration-tests/` - passed
- `golangci-lint` with project config - `0 issues`

## Relationship to #749

I noticed #749 landed a very similar per-block iteration approach right before I opened this. There's overlap on the fix itself. What this PR adds beyond #749:
- Fixture-backed regression that reproduces the exact HashiCorp dual-key/same-fingerprint shape (#749's diff changes only `lib/checksum.go`, no test).
- Pre-rotation coverage (1.12.2) to guard against a future "fix" that favors the renewed block and silently regresses legacy releases.

Happy to close this in favor of #749 if you'd prefer - or rebase this fixture-backed test on top of #749 once it merges. Whichever is lowest-friction for you.

## Test plan

- [x] Unit: `go test ./lib/ -run Test_checkSignatureOfChecksums_hashicorpDualKey`
- [x] Unit (full): `CGO_ENABLED=0 go test ./...`
- [x] Integration: all 7 dirs under `test-data/integration-tests/`
- [x] End-to-end: locally built binary installs Terraform 1.11.4, 1.12.2, 1.13.0, 1.14.9

Fixes #746
Refs #747, #749


yermulnik:
Hey @pszypowicz 
Thanks for quick turnaround. I'd prefer to stick with #749

pszypowicz:
Absolutely understandable 🫡 I already had this prepared, was just a few minutes shy because this time I wanted to be super extra safe - see the additional test including actual HC PGP key and regression test. But figured out at least maybe it will be useful as a reference, even if it would be closed. Thanks one more time!

PR #755, opened 2026-04-22
Simplify verification logic

MatthewJohn:
Replace all close handlers with defers.
Update checkSignatureOfChecksums to take content of files, so it doesn't handle filehandle Re-use content of files across calls to checkSignatureOfChecksums, with io Reader for legacy fallback


yermulnik (approved):
LGTM
Appreciate the assistance ❤️

PR #756, opened 2026-04-25
feat(versiontf): Higher priority for .terraform-version file

shinespb:
PR introduces --explicit-tfversion (-e) to support deterministic CI installs without ignoring module compatibility rules.

In CI, we often want to pin a specific Terraform/OpenTofu version via .terraform-version for reproducibility. At the same time, we still need to respect required_version constraints declared in version.tf / version.tofu. With this flag enabled, tfswitch treats .terraform-version as an explicit additional constraint during resolution.

If the pinned version satisfies module constraints, that exact version is selected.
If it conflicts, tfswitch fails with a clear constraint-mismatch error.
Default behavior is unchanged when --explicit-tfversion is not used.

This addresses real CI fragility scenarios like [#746](https://github.com/warrensbox/terraform-switcher/issues/746), where relying on non-pinned/latest resolution caused unexpected install failures (GPG key issue), by letting pipelines pin safely while still enforcing declared compatibility.


yermulnik:
@shinespb Thank you for the contribution.
As I'm not really clear about the particular use case, what this PR resolves and why the `required_version` can't be used to pin TF version, could you please elaborate and provide more details?

> If the pinned version satisfies module constraints, that exact version is selected.
> If it conflicts, tfswitch fails with a clear constraint-mismatch error.

If I'm getting the idea right, what you need is: instead of creating `.terraform-version` with e.g. `1.14.5` string in it, you'd create the `.terraform-version.tf` (or whatever name you prefer — the only requirement for it is to have `tf` extension) with the below content:
```hcl
terraform {
  required_version = "1.14.5"
}
```
And then both `tfswtich` and `terraform` (or `tofu`) would parse the compound constraint and fail if constraints contradict. E.g.:
```shell
> cat terraform.tf
terraform {
  required_version = "~> 1.14"
}

> tfswitch --dry-run
15:12:34.880 INFO Reading configuration from "/home/georgii/.tfswitch.toml"
15:12:34.880 INFO Reading version constraint from Terraform/OpenTofu module at "."
15:12:35.554 INFO Reading required version from constraint: "~> 1.14"
15:12:35.555 INFO Matched version: "1.14.9"
15:12:35.555 INFO [DRY-RUN] No changes will be made
15:12:35.643 INFO [DRY-RUN] Would have attempted to install Terraform version "1.14.9" for "linux_amd64"

# Added ".terraform-version.tf"
> cat .terraform-version.tf
terraform {
  required_version = "1.14.5"
}

> tfswitch --dry-run
15:13:19.204 INFO Reading configuration from "/home/georgii/.tfswitch.toml"
15:13:19.205 INFO Reading version constraint from Terraform/OpenTofu module at "."
15:13:19.526 INFO Reading required version from constraint: "1.14.5, ~> 1.14"
15:13:19.527 INFO Matched version: "1.14.5"
15:13:19.527 INFO [DRY-RUN] No changes will be made
15:13:19.623 INFO [DRY-RUN] Would have attempted to install Terraform version "1.14.5" for "linux_amd64"

# Changed constraint to impossible in ".terraform-version.tf"
> cat .terraform-version.tf
terraform {
  required_version = "1.1.1"
}

> tfswitch --dry-run
15:13:53.096 INFO Reading configuration from "/home/georgii/.tfswitch.toml"
15:13:53.096 INFO Reading version constraint from Terraform/OpenTofu module at "."
15:13:53.378 INFO Reading required version from constraint: "1.1.1, ~> 1.14"
15:13:53.379 ERROR No version found matching "1.1.1, ~> 1.14"
15:13:53.379 FATAL Failed to obtain settings from Terraform module: Did not find version matching constraint: 1.1.1, ~> 1.14
```

yermulnik:
JFYI: support for `.terraform-version` file was introduced long ago at the very early stage of `tfswitch` as a basic compatibility layer between `tfswitch` and the other tool that at that time was the only(?) commonly popular tool to manage TF versions. Granted this feature is external to Terraform/OpenTofu, I'd encourage you to prefer Terraform's (and respectively `tfswitch`'s) native feature (the `required_version` constraints) instead of relying upon external files (which would fail providing desired result or guardrail if someone used Terraform bypassing the necessity to run `tfswitch` as a pre-step).

MatthewJohn:
If I'm understanding correctly, you want to specify a version of Terraform, but still respect the constaints set in code (and fail if not).
If so, I wouldn't set another place to set a version - as far as I know, we ordering of argument presendence in place where CLI args should override constraints obtained via code. So if 1.1.1 is specified via CLI, it should override the constraints (which is why default-version was added as a fallback).
At the moment, the constaints are used to derrive a version... what I would suggest instead is: versioning can stay the same (CLI should take presedence, which if I understand is what you want) and then add a flag to enforce respecting of constraints.
This would then mean the version is still prefered from CLI args (or env var), but is then blocked by not matching constraints and would throw an error

yermulnik:
I lean to think that OP wanted to keep the sort of temp constraint tracked via IaaC. Let's wait for OP to post an update — or else we'll be closing this PR. @shinespb Could you pelase?

shinespb:
> If I'm understanding correctly, you want to specify a version of Terraform, but still respect the constaints set in code (and fail if not).

Yes, it's correct. And the idea was not to define version(I know it's possible) via CLI(for that we would need to change pipeline), but there's already `.terraform-version` which is read anyway(and I guess it simpler to have version there and change it per environment / project). In my case we have multiple envs in the same repository and for some environments we can use different TF versions. 

> If I'm getting the idea right, what you need is: instead of creating .terraform-version with e.g. 1.14.5 string in it, you'd create the .terraform-version.tf (or whatever name you prefer — the only requirement for it is to have tf extension)

I guess it also do the trick. Just wanted to have simpler way to pin version.

yermulnik:
> I guess it also do the trick. Just wanted to have simpler way to pin version.

Well, my suggestion is simpler and more straightforward I guess: you just drop a single file in the directory with TF code and do not need to adjust `tfswitch` command in your CI. Once you don't need the "quirk" anymore, you just remove the file and do not need to adjust `tfswitch` command back to normal. Besides that this approach would safely pin TF version for when TF is run outside of your CI (for whichever reason it might happen).

I'm closing this PR. Thank you for the contribution 🤝

Issue #745, opened 2026-04-20
feat(opentofu): Allow user-provided mirror for artifact download


yermulnik:
Relates to #739 and #744 

https://github.com/warrensbox/terraform-switcher/pull/739#discussion_r3110241509
https://github.com/warrensbox/terraform-switcher/pull/739/changes/BASE..36fce9637e57c7310e9b9ddfc077c7991034652e#diff-376c3614d498774d798a18d5de9c1ce45d6db10614a209cb5927f4f7dd171bd1R179

> // TODO For a future release, use user-provided mirror, as we ONLY allow downloading via public OpenTofu URL

yermulnik:
FYI: #777

PR #772, opened 2026-06-03
feat(mirror): support `file://` schema for the mirror URL

derekbassett:
## Summary

Allow `--mirror` to point at a local file via a `file://` URL, so the list of available versions can be resolved from disk instead of over HTTP.

This enables version resolution in **air-gapped environments** where `releases.hashicorp.com` (or any HTTP mirror) is unreachable. The version list can be generated ahead of time and dropped on disk, then `tfswitch` resolves `required_version` constraints against it offline.

```bash
tfswitch --mirror file:///opt/tfswitch/index.json
```

## What changed

- `getTFURLBody` branches on the `file://` scheme and reads the file directly via a new `getTFFileBody` helper.
- The file is expected to contain the same content an HTTP mirror would serve — either the JSON index (e.g. the body of `https://releases.hashicorp.com/terraform/index.json`) or the HTML directory listing. It flows through the existing JSON-then-HTML parsing unchanged.
- Docs: added a "Local file mirror" subsection under the existing custom-mirror documentation.

## Scope

This affects **only how the version list is resolved**. Binary downloads are unchanged and still occur over HTTP from the mirror's archive tree. (Offline binary download is a larger, separate change — it relates to the download-mirror separation discussed in #745 — and is intentionally out of scope here.)

## Tests

- `TestGetTFURLBody_File` — reads the version list from a `file://` URL.
- `TestGetTFURLBody_FileMissing` — a missing file returns an error rather than terminating the process.
- `TestGetTFList_File` — end-to-end version resolution through `getTFList` from a local file.

All `lib` and `lib/param_parsing` tests pass; `go vet` and `go build` are clean.

## Note for reviewers

The `file://` path returns an error on a missing/unreadable file, rather than following the HTTP path's existing `logger.Fatalf` convention. This is intentional: `getTFList` already propagates returned errors, it keeps the function testable, and it surfaces a clear failure instead of an empty version list. Happy to switch to `Fatalf` if you'd prefer consistency with the surrounding code.

### DIFF
--- a/lib/list_versions.go
+++ b/lib/list_versions.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"reflect"
 	"regexp"
 	"slices"
@@ -168,6 +169,14 @@ func getTFLatestImplicit(product Product, mirrorURL string, preRelease bool, req
 
 // getTFURLBody : Get list of versions from the mirror URL
 func getTFURLBody(mirrorURL string) (string, error) {
+	// Local file mirror: read the version list straight off disk instead of
+	// fetching it over HTTP. Useful for air-gapped environments where no HTTP
+	// mirror is reachable. The file is expected to contain the same JSON (or
+	// HTML directory listing) that an HTTP mirror would serve.
+	if strings.HasPrefix(mirrorURL, "file://") {
+		return getTFFileBody(mirrorURL)
+	}
+
 	hasSlash := strings.HasSuffix(mirrorURL, "/")
 	isJSON := strings.HasSuffix(mirrorURL, ".json")
 	if !hasSlash && !isJSON {
@@ -192,6 +201,18 @@ func getTFURLBody(mirrorURL string) (string, error) {
 	bodyString := string(body)
 
 	return bodyString, nil
+}
+
+// getTFFileBody : Read the version list from a local `file://` mirror URL.
+// Unlike the HTTP path, a missing or unreadable file returns an error rather
+// than terminating the process, so callers can surface a clear failure.
+func getTFFileBody(mirrorURL string) (string, error) {
+	filePath := strings.TrimPrefix(mirrorURL, "file://")
+	body, errBody := os.ReadFile(filePath) // nolint:gosec // `filePath' is expected to be variable
+	if errBody != nil {
+		return "", fmt.Errorf("Error reading version list from file %q: %v", filePath, errBody)
+	}
+	return string(body), nil
 }
 
 // versionExist : Check if requested version exists
--- a/lib/list_versions_test.go
+++ b/lib/list_versions_test.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"os"
+	"path/filepath"
 	"reflect"
 	"slices"
 	"strings"
@@ -372,6 +374,57 @@ func TestGetTFURLBody(t *testing.T) {
 	}
 	if body != hashicorpBody {
 		t.Errorf("Body not returned correctly. Expected: %s, actual: %s", hashicorpBody, body)
+	}
+}
+
+// TestGetTFURLBody_File : Test getTFURLBody reads from a local file:// mirror
+func TestGetTFURLBody_File(t *testing.T) {
+	logger = InitLogger("DEBUG")
+
+	tempDir := t.TempDir()
+	versionFile := filepath.Join(tempDir, "index.json")
+	if err := os.WriteFile(versionFile, []byte(hashicorpJSONData), 0o600); err != nil {
+		t.Fatalf("Could not write version list fixture: %v", err)
+	}
+
+	body, err := getTFURLBody("file://" + versionFile)
+	if err != nil {
+		t.Errorf("Unexpected error reading file:// mirror: %v", err)
+	}
+	if body != hashicorpJSONData {
+		t.Errorf("Body not returned correctly. Expected: %s, actual: %s", hashicorpJSONData, body)
+	}
+}
+
+// TestGetTFURLBody_FileMissing : Test getTFURLBody returns an error (rather
+// than terminating the process) when a file:// mirror points at a missing file
+func TestGetTFURLBody_FileMissing(t *testing.T) {
+	logger = InitLogger("DEBUG")
+
+	missing := filepath.Join(t.TempDir(), "does-not-exist.json")
+	_, err := getTFURLBody("file://" + missing)
+	if err == nil {
+		t.Error("Expected an error for a missing file:// mirror, got nil")
+	}
+}
+
+// TestGetTFList_File : Test getTFList resolves versions end-to-end from a
+// local file:// mirror
+func TestGetTFList_File(t *testing.T) {
+	logger = InitLogger("DEBUG")
+
+	tempDir := t.TempDir()
+	versionFile := filepath.Join(tempDir, "index.json")
+	if err := os.WriteFile(versionFile, []byte(hashicorpJSONData), 0o600); err != nil {
+		t.Fatalf("Could not write version list fixture: %v", err)
+	}
+
+	list, err := getTFList(GetProductById("terraform"), "file://"+versionFile, true)
+	if err != nil {
+		t.Errorf("Unexpected error listing versions from file:// mirror: %v", err)
+	}
+	if !slices.Contains(list, "0.12.2") {
+		t.Errorf("Expected version list from file:// mirror to contain 0.12.2, got: %v", list)
 	}
 }
 
--- a/www/docs/usage/commandline.md
+++ b/www/docs/usage/commandline.md
@@ -89,6 +89,24 @@ the mirror root, matching that same layout.
 tfswitch --mirror https://example.jfrog.io/artifactory/hashicorp
 ```
 
+### Local file mirror
+
+The version list can also be read from a local file using a `file://` URL. This
+is useful in air-gapped environments where no HTTP mirror is reachable: generate
+the version list ahead of time and point `tfswitch` at the file on disk.
+
+```bash
+tfswitch --mirror file:///opt/tfswitch/index.json
+```
+
+The file must contain the same content an HTTP mirror would serve — either the
+JSON index (e.g. the body of `<https://releases.hashicorp.com/terraform/index.json>`)
+or the HTML directory listing. A missing or unreadable file causes `tfswitch` to
+exit with an error rather than silently producing an empty version list.
+
+This affects only how the list of available versions is resolved; the binaries
+themselves are still downloaded over HTTP from the mirror's archive tree.
+
 ## Install to non-default location
 
 By default `tfswitch` will download the Terraform binary to the user home


lib/list_versions.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
```
yermulnik, line 8: Let's add some robustness to URL parsing (this and other commit suggestions down the PR)

```suggestion
	"net/url"
	"os"
	"path/filepath"
	"runtime"
```

The `"runtime"` would need to go after `"regexp"` though. Please put it accordingly as I can make it a commit suggestion.

I'm also adding commit suggestion w/o validating them. Please take care of this if I made any mistakes unintentionally.

lib/list_versions.go
```diff
@@ -168,6 +169,14 @@ func getTFLatestImplicit(product Product, mirrorURL string, preRelease bool, req
 
 // getTFURLBody : Get list of versions from the mirror URL
 func getTFURLBody(mirrorURL string) (string, error) {
+	// Local file mirror: read the version list straight off disk instead of
+	// fetching it over HTTP. Useful for air-gapped environments where no HTTP
+	// mirror is reachable. The file is expected to contain the same JSON (or
+	// HTML directory listing) that an HTTP mirror would serve.
+	if strings.HasPrefix(mirrorURL, "file://") {
+		return getTFFileBody(mirrorURL)
+	}
```
yermulnik, line 178:
```suggestion
	parsedURL, err := url.Parse(mirrorURL)
	if err != nil {
		return "", fmt.Errorf("Error parsing mirror URL %q: %v", mirrorURL, err)
	}

	// Local file mirror: read the version list straight off disk instead of
	// fetching it over HTTP. Useful for air-gapped environments where no HTTP
	// mirror is reachable. The file is expected to contain the same JSON (or
	// HTML directory listing) that an HTTP mirror would serve.
	if parsedURL.Scheme == "file" {
		return getTFFileBody(parsedURL.Path)
	}
```

lib/list_versions.go
```diff
@@ -194,6 +203,18 @@ func getTFURLBody(mirrorURL string) (string, error) {
...
+// Unlike the HTTP path, a missing or unreadable file returns an error rather
+// than terminating the process, so callers can surface a clear failure.
+func getTFFileBody(mirrorURL string) (string, error) {
+	filePath := strings.TrimPrefix(mirrorURL, "file://")
+	body, errBody := os.ReadFile(filePath) // nolint:gosec // `filePath' is expected to be variable
+	if errBody != nil {
+		return "", fmt.Errorf("Error reading version list from file %q: %v", filePath, errBody)
+	}
+	return string(body), nil
+}
```
yermulnik, line 216:
```suggestion
func getTFFileBody(parsedPath string) (string, error) {
	if runtime.GOOS == "windows" {
		// If path starts with a slash followed by a drive letter
		// (e.g., "/C:/path"), strip the leading slash
		if len(parsedPath) > 2 && parsedPath[0] == '/' && parsedPath[2] == ':' {
			parsedPath = parsedPath[1:]
		}
		// Convert forward slashes to backslashes for Windows filesystem
		parsedPath = filepath.FromSlash(parsedPath)
	}

	body, err := os.ReadFile(parsedPath) // nolint:gosec // `parsedPath' is expected to be variable
	if err != nil {
		return "", fmt.Errorf("Error reading version list from file %q: %v", parsedPath, err)
	}
	return string(body), nil
}
```

lib/list_versions_test.go
```diff
@@ -375,6 +377,57 @@ func TestGetTFURLBody(t *testing.T) {
...
+
+	tempDir := t.TempDir()
+	versionFile := filepath.Join(tempDir, "index.json")
+	if err := os.WriteFile(versionFile, []byte(hashicorpJSONData), 0o600); err != nil {
+		t.Fatalf("Could not write version list fixture: %v", err)
+	}
+
+	body, err := getTFURLBody("file://" + versionFile)
+	if err != nil {
+		t.Errorf("Unexpected error reading file:// mirror: %v", err)
```
yermulnik, line 392:
```suggestion
		t.Errorf("Unexpected error reading mirror file %q: %v", versionFile, err)
```

lib/list_versions_test.go
```diff
@@ -375,6 +377,57 @@ func TestGetTFURLBody(t *testing.T) {
...
+
+// TestGetTFURLBody_FileMissing : Test getTFURLBody returns an error (rather
+// than terminating the process) when a file:// mirror points at a missing file
+func TestGetTFURLBody_FileMissing(t *testing.T) {
+	logger = InitLogger("DEBUG")
+
+	missing := filepath.Join(t.TempDir(), "does-not-exist.json")
+	_, err := getTFURLBody("file://" + missing)
+	if err == nil {
+		t.Error("Expected an error for a missing file:// mirror, got nil")
```
yermulnik, line 407:
```suggestion
		t.Error("Expected an error for a missing mirror file %q, got nil", missing)
```

lib/list_versions_test.go
```diff
@@ -375,6 +377,57 @@ func TestGetTFURLBody(t *testing.T) {
...
+
+	tempDir := t.TempDir()
+	versionFile := filepath.Join(tempDir, "index.json")
+	if err := os.WriteFile(versionFile, []byte(hashicorpJSONData), 0o600); err != nil {
+		t.Fatalf("Could not write version list fixture: %v", err)
+	}
+
+	list, err := getTFList(GetProductById("terraform"), "file://"+versionFile, true)
+	if err != nil {
+		t.Errorf("Unexpected error listing versions from file:// mirror: %v", err)
```
yermulnik, line 424:
```suggestion
		t.Errorf("Unexpected error listing versions from mirror file %q: %v", versionFile, err)
```

lib/list_versions_test.go
```diff
@@ -375,6 +377,57 @@ func TestGetTFURLBody(t *testing.T) {
...
+	}
+	if !slices.Contains(list, "0.12.2") {
+		t.Errorf("Expected version list from file:// mirror to contain 0.12.2, got: %v", list)
```
yermulnik, line 427:
```suggestion
		t.Errorf("Expected version list from mirror file %q to contain 0.12.2, got: %v", versionFile, list)
```

www/docs/usage/commandline.md
```diff
@@ -89,6 +89,24 @@ the mirror root, matching that same layout.
...
+tfswitch --mirror file:///opt/tfswitch/index.json
+```
+
+The file must contain the same content an HTTP mirror would serve — either the
+JSON index (e.g. the body of `<https://releases.hashicorp.com/terraform/index.json>`)
+or the HTML directory listing. A missing or unreadable file causes `tfswitch` to
+exit with an error rather than silently producing an empty version list.
+
+This affects only how the list of available versions is resolved; the binaries
+themselves are still downloaded over HTTP from the mirror's archive tree.
```
yermulnik, line 108:
```suggestion
themselves are still downloaded over HTTP from the mirror's archive tree (if
they don't already exist in install path (see `--install` command-line flag)).
```

yermulnik (commented):
Thank you for the contribution ❤️ 
Please see my code suggestions — hope they make sense =)

lib/list_versions.go
MatthewJohn: Does `go fmt` not fix this and, if so, is there a PR check that validates against go fmt?

lib/list_versions.go
```diff
@@ -194,6 +203,18 @@ func getTFURLBody(mirrorURL string) (string, error) {
 	return bodyString, nil
 }
 
+// getTFFileBody : Read the version list from a local `file://` mirror URL.
+// Unlike the HTTP path, a missing or unreadable file returns an error rather
+// than terminating the process, so callers can surface a clear failure.
+func getTFFileBody(mirrorURL string) (string, error) {
```
MatthewJohn, line 209: I think the name of the function could maybe be a little more explicit, at least due to naming, i.e. get (applies to retrieving it locally or remotely), tf (obviously tf), file (a remote file could easily be named this way) and body is also maybe confusing since it's normally described as around the context of http.
Perhaps:
```suggestion
func getLocalFileContent(mirrorURL string) (string, error) {
```
?

lib/list_versions.go
```diff
@@ -194,6 +203,18 @@ func getTFURLBody(mirrorURL string) (string, error) {
...
+	filePath := strings.TrimPrefix(mirrorURL, "file://")
```
MatthewJohn, line 210: I think with the use of `url.Parse` (as suggested in https://github.com/warrensbox/terraform-switcher/pull/772/changes#r3357779197), this function could then just be passed the URL's path instead and then no need the stripping of the scheme

lib/list_versions.go
MatthewJohn: As per above, I'd hope (like to think, perhaps) that URL parse would handle this conversion - I could _well_ be wrong, though?

lib/list_versions_test.go
```diff
@@ -375,6 +377,57 @@ func TestGetTFURLBody(t *testing.T) {
...
+func TestGetTFURLBody_File(t *testing.T) {
+	logger = InitLogger("DEBUG")
+
+	tempDir := t.TempDir()
+	versionFile := filepath.Join(tempDir, "index.json")
+	if err := os.WriteFile(versionFile, []byte(hashicorpJSONData), 0o600); err != nil {
+		t.Fatalf("Could not write version list fixture: %v", err)
+	}
+
+	body, err := getTFURLBody("file://" + versionFile)
```
MatthewJohn, line 390: I'd be incredibly keen on having some explicit tests around absolute and relative paths - I've seen other projects with real.. interesting behavoir around this, e.g. `file://test` vs `file:///test` vs `file://./test`

lib/list_versions_test.go
```diff
@@ -375,6 +377,57 @@ func TestGetTFURLBody(t *testing.T) {
...
+	}
+}
+
+// TestGetTFURLBody_FileMissing : Test getTFURLBody returns an error (rather
+// than terminating the process) when a file:// mirror points at a missing file
+func TestGetTFURLBody_FileMissing(t *testing.T) {
+	logger = InitLogger("DEBUG")
+
+	missing := filepath.Join(t.TempDir(), "does-not-exist.json")
+	_, err := getTFURLBody("file://" + missing)
```
MatthewJohn, line 405: We should also validate what the first returned value also is (i.e. that it's an empty string)

lib/list_versions.go
yermulnik: While I agree with Matt's suggestion, I guess @derekbassett targeted to maintain consistency with the ancestor-function naming `getTFURLBody()`.
And in the context of this specific file and use case, I'd lean to say the `getTFFileBody()` fits better (and we can re-factor naming and stuff when we get to re-work things for v2).

lib/list_versions.go
yermulnik: 100%. It's what I suggested too.

lib/list_versions.go
yermulnik: Yeah, I guess our PR checks should catch this. I just wanted author to not decide that I'm suggesting incorrect imports sorting on purpose =)

yermulnik:
@derekbassett Just checking in: will you be up to consider comments and suggestions? Thank you.

derekbassett:
Yes, I'm sorry. I will get to this soon. Agree with everything.

yermulnik:
> Yes, I'm sorry. I will get to this soon. Agree with everything.

@derekbassett Are you still up to take care of this PR please?

yermulnik:
I'm closing this PR as stale. @derekbassett Please re-open if you still up to complete it. Thank you.

PR #777, opened 2026-06-06
feat(download): Allow user-provided mirror for artifact download

### LINKED ISSUE #745: feat(opentofu): Allow user-provided mirror for artifact download

yermulnik:
Resolves #745


lib/param_parsing/parameters.go
```diff
@@ -86,6 +87,11 @@ func setupProductParam(params *Params) {
 			logger.Debugf("Default mirror URL: %q", params.MirrorURL)
 		}
 
+		if params.MirrorDownloadURL == "" {
+			params.MirrorDownloadURL = product.GetDefaultDownloadMirrorUrl()
+			logger.Debugf("Default download URL: %q", params.MirrorDownloadURL)
+		}
```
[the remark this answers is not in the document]
yermulnik: This is intentional as OpenTofu's URLs differ anyways.

lib/install.go
```diff
@@ -118,7 +119,15 @@ func install(product Product, dryRun, showRequiredFlag bool, tfversion, binPath,
...
-	lockFile := filepath.Join(os.TempDir(), ".tfswitch."+product.GetId()+".lock")
+	// Assume no race condition between different users running tfswitch on
+	// the same machine as they're meant to use user-specific locations and
+	// only root is expected to write to system-wide location
+	// (user.Current() would resolve to root under sudo)
+	currentUser, err := user.Current()
+	if err != nil {
+		return fmt.Errorf("Failed to look up current user: %v", err)
+	}
+	lockFile := filepath.Join(os.TempDir(), ".tfswitch."+currentUser.Username+".lock")
```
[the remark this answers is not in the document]
yermulnik: Switched to `currentUser.Uid`

lib/install.go
```diff
@@ -230,34 +239,34 @@ func installableBinLocation(product Product, userBinPath string) string {
...
+func InstallLatestVersion(dryRun, showRequiredFlag bool, customBinaryPath, installPath, mirrorURL, mirrorDownloadURL, arch string) {
 	product := getLegacyProduct()
 	//nolint:errcheck // Function is deprecated
-	InstallLatestProductVersion(product, dryRun, showRequiredFlag, customBinaryPath, installPath, mirrorURL, arch)
+	InstallLatestProductVersion(product, dryRun, showRequiredFlag, customBinaryPath, installPath, mirrorURL, mirrorDownloadURL, arch)
 }
 
 // InstallLatestProductVersion install latest stable tf version
-func InstallLatestProductVersion(product Product, dryRun, showRequiredFlag bool, customBinaryPath, installPath, mirrorURL, arch string) error {
+func InstallLatestProductVersion(product Product, dryRun, showRequiredFlag bool, customBinaryPath, installPath, mirrorURL, mirrorDownloadURL, arch string) error {
```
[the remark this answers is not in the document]
yermulnik: We've got no dependents so it's safe as of now.

lib/product.go
```diff
@@ -41,6 +41,7 @@ type Product interface {
 	GetId() string
 	GetName() string
 	GetDefaultMirrorUrl() string
+	GetDefaultDownloadMirrorUrl() string
 	GetVersionPrefix() string
```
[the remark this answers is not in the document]
yermulnik: We've got no dependents so it's safe as of now.

lib/install.go
```diff
@@ -118,7 +119,15 @@ func install(product Product, dryRun, showRequiredFlag bool, tfversion, binPath,
...
+	// Assume no race condition between different users running tfswitch on
+	// the same machine as they're meant to use user-specific locations and
+	// only root is expected to write to system-wide location
+	// (user.Current() would resolve to root under sudo)
+	currentUser, err := user.Current()
+	if err != nil {
+		return fmt.Errorf("Failed to look up current user: %v", err)
+	}
+	lockFile := filepath.Join(os.TempDir(), ".tfswitch."+currentUser.Uid+".lock")
 	// 90 attempts * 2 seconds = 3 minutes to acquire lock, otherwise bail out
```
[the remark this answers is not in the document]
yermulnik: Installing product implies changing symlink shared by both products — so the code is meant to prevent race condition.

lib/utils.go
```diff
@@ -82,3 +85,18 @@ func RemoveDuplicateStrings(slice []string) []string {
...
+	parsedURL, err := url.ParseRequestURI(urlString)
+	if err != nil {
+		return fmt.Errorf("URL %v", err)
+	}
+
+	allowedSchemes := []string{"file", "http", "https"}
+
+	if !slices.Contains(allowedSchemes, parsedURL.Scheme) || parsedURL.Host == "" {
+		return fmt.Errorf("URL must have a valid host and a scheme must be one of: %s: %q", strings.Join(allowedSchemes, ", "), urlString)
+	}
```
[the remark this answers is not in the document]
yermulnik: `ftp` was meant. fixed.

lib/download.go
```diff
@@ -20,6 +20,10 @@ func DownloadFromURL(installLocation, mirrorURL, tfversion, versionPrefix, goos,
 }
 
 func DownloadProductFromURL(product Product, installLocation, mirrorURL, tfversion, versionPrefix, goos, goarch string) (string, error) {
+	if mirrorURL == "" {
+		return "", errors.New("download URL is invalid")
+	}
```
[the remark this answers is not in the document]
yermulnik: This is intentional as the empty string is returned when error parsing URL occurs.

lib/product.go
```diff
@@ -77,6 +78,10 @@ func (p TerraformProduct) GetDefaultMirrorUrl() string {
 	return p.DefaultMirror
 }
 
+func (p TerraformProduct) GetDefaultDownloadMirrorURL() string {
```
MatthewJohn, line 81: We should probably move to a base struct at some point, which implements some of the common interfaces

lib/product.go
```diff
@@ -100,10 +105,23 @@ func (p TerraformProduct) GetArtifactUrl(mirrorURL string, version string) strin
 	}
 
 	// If the actual mirror is not the default, use this mirror for downloading
-	if mirrorURL != p.DefaultMirror {
+	if mirrorURL != "" && mirrorURL != p.DefaultDownloadMirror {
```
MatthewJohn, line 108:
```suggestion
	if mirrorURL != "" && mirrorURL != p.GetDefaultDownloadMirrorURL() {
```
We should use this to unify how we "get" the default download URL and probably try to avoid touching the struct attribute as much as possible

lib/install.go
```diff
@@ -318,7 +331,7 @@ type VersionSelector struct {
 // InstallProductOption displays & installs tf version
 /* listAll = true - all versions including beta and rc will be displayed */
 /* listAll = false - only official stable release are displayed */
-func InstallProductOption(product Product, listAll, dryRun, showRequiredFlag bool, customBinaryPath, installPath, mirrorURL, arch string) error {
+func InstallProductOption(product Product, listAll, dryRun, showRequiredFlag bool, customBinaryPath, installPath, mirrorURL, mirrorDownloadURL, arch string) error {
```
MatthewJohn, line 334: Mentioned in chat, but we should be careful (or at least actively decide) if we want to break public signatures in non-major versions.
Going forward, might be worth considering param structs for these public methods, so we can add attributes to the structs (and then cleanly handle default values) as non-breaking changes.

lib/install.go
```diff
@@ -116,9 +117,21 @@ func install(product Product, dryRun, showRequiredFlag bool, tfversion, binPath,
...
+	// * Put lockfile in temp directory to get it cleaned up on reboot
+	// * Assume no race condition between different users running tfswitch on
+	//   the same machine as they're meant to use user-specific locations and
+	//   only root is expected to write to system-wide location
+	//   (JFYI: user.Current() would resolve to "root" under "sudo")
+	// * user.Current() can fail in statically linked builds (CGO_ENABLED=0)
+	//   or minimal/container environments without user database entries, so
+	//   default to "unknown" if we can't get a UID for the current user
+	uid := "unknown"
+	currentUser, err := user.Current()
```
MatthewJohn, line 130: This feels like an unrelated change that's snuck in - was it causing an issue or trying to fix a particular check or something? If so, is it worth capturing as a comment? (or maybe there's commit message that I should read 👁️ )

lib/utils.go
```diff
@@ -82,3 +85,20 @@ func RemoveDuplicateStrings(slice []string) []string {
...
+
+func IsValidRemoteURL(urlString string) error {
+	parsedURL, err := url.ParseRequestURI(urlString)
+	if err != nil {
+		return fmt.Errorf("URL %v", err)
+	}
+
+	// downloads are performed via net/http (http.Get), which will fail at
+	// runtime for ftp://... URLs, so allow only http and https schemes here
+	allowedSchemes := []string{"http", "https"}
```
MatthewJohn, line 97: I feel like it's worth considering how this will impact https://github.com/warrensbox/terraform-switcher/pull/772 ?

lib/utils.go
```diff
@@ -82,3 +85,20 @@ func RemoveDuplicateStrings(slice []string) []string {
...
+
+	if !slices.Contains(allowedSchemes, parsedURL.Scheme) || parsedURL.Host == "" {
```
MatthewJohn, line 99: Is it worth performing two different conditions to be able to provide a specific error, i.e. empty host or invalid scheme?

lib/install.go
yermulnik: Totally agree. We should avoid such breaking changes. Luckily we've got no dependent projects based on GH's data so I went with changing public func 🤷🏻 

> Going forward, might be worth considering param structs for these public methods, so we can add attributes to the structs (and then cleanly handle default values) as non-breaking changes.

I wish we all could get together and for a refactoring for v2 major version 🤞🏻

lib/product.go
yermulnik: Agree. This time (like many other times) I've gone with implementing code as much consistently with what already exists as I could.

lib/install.go
yermulnik: While I was testing new the feature from this PR, I've noticed that previously implemented locking feature treated `terraform` and `opentofu` differently allowing race condition for simultaneous install of both. So I reworked locking file naming to rely upon user id instead of Product name. This change seemed to me not much critical and I decided to not branch it as dedicated PR.
Do you like me to fork it into its own separate PR?

lib/utils.go
yermulnik: At the moment this function is used to validate URLs for artifact downloads only (and hence its name has `remote` word in it — so it's meant for remote URL (while it doesn't allow `ftp` on purpose as `ftp` protocol isn't handled by our download function).
The #772 is for API URL so at the moment these two do not overlap and no impact that I can imagine.

lib/utils.go
yermulnik: I thought on this and it didn't make much sense to me. The error thrown provides original URL and hence consumer can see whether it's schema or host.
Let me know if you'd like this to be two different `if`s and I'll be happy to split in two conditionals.
Thanks.

MatthewJohn: It's probably good for now, I guess.
Can always change if it does cause confusion

lib/install.go
```diff
@@ -116,9 +117,21 @@ func install(product Product, dryRun, showRequiredFlag bool, tfversion, binPath,
...
+	// * Create exclusive lock to prevent multiple concurrent installations
+	// * Put lockfile in temp directory to get it cleaned up on reboot
+	// * Assume no race condition between different users running tfswitch on
+	//   the same machine as they're meant to use user-specific locations and
+	//   only root is expected to write to system-wide location
+	//   (JFYI: user.Current() would resolve to "root" under "sudo")
+	// * user.Current() can fail in statically linked builds (CGO_ENABLED=0)
+	//   or minimal/container environments without user database entries, so
+	//   default to "unknown" if we can't get a UID for the current user
+	uid := "unknown"
```
warrensbox, line 129: Is this standard practice when uid is not available?

warrensbox (approved):

lib/install.go
yermulnik: I'd say it's more of a safety measure as I can't imagine `tfswitch` being used in a very minimal or obscure environments which might lack user database which provide user IDs.
But to stay on the safe side, I thought we're better off setting a safe fallback for a potential conditions so that we don't end up with empty value.

